PR #1, opened 2023-10-10
Add CI to run non-indexer E2E tests

banool:
### Description
This adds CI that runs against a local testnet from the latest CLI.

### Test Plan
CI.


.github/actions/run-e2e-tests/action.yaml
```diff
@@ -0,0 +1,56 @@
...
+    # not included in the commit.
+    - run: pnpm install --frozen-lockfile
+      shell: bash
+
+    # Install the CLI.
+    - run: pnpm install -g @aptos-labs/aptos-cli
+      shell: bash
+
+    # Run a local testnet in the background.
+    - run: aptos node run-local-testnet --force-restart --assume-yes --with-faucet &
```
0xmaayan, line 30: interesting... I thought we would need to `npx aptos....`

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "aptos",
   "description": "Aptos SDK",
-  "packageManager": "pnpm@8.3.1",
+  "packageManager": "pnpm@8.9.0",
```
0xmaayan, line 4: why?

.github/actions/run-e2e-tests/action.yaml
banool: In this case since runners are transient I just install the CLI globally.

package.json
banool: Lockfiles were giving me a hard time, I found some bug that is fixed in like 8.7+. Any reason not to?

0xmaayan: nope, just curios

gregnazario (approved):

PR #3, opened 2023-10-10
Fix broken tests

gregnazario:
### Description
Disables a test, and fixes the type tag error message test

### Test Plan
See CI

### DIFF
--- a/src/core/account_address.ts
+++ b/src/core/account_address.ts
@@ -246,8 +246,8 @@ export class AccountAddress extends Serializable {
       } else if (args.input.length !== 3) {
         // 0x + one hex char is the only valid SHORT form for special addresses.
         throw new ParsingError(
-          `The given hex string ${args.input} is a special address not in LONG form, 
-          it must be 0x0 to 0xf without padding zeroes.`,
+            // eslint-disable-next-line max-len
+          `The given hex string ${args.input} is a special address not in LONG form, it must be 0x0 to 0xf without padding zeroes.`,
           AddressInvalidReason.INVALID_PADDING_ZEROES,
         );
       }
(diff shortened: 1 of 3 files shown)


0xjinn (approved):

xbtmatt (approved):

src/core/account_address.ts
```diff
@@ -246,8 +246,8 @@ export class AccountAddress extends Serializable {
       } else if (args.input.length !== 3) {
         // 0x + one hex char is the only valid SHORT form for special addresses.
         throw new ParsingError(
-          `The given hex string ${args.input} is a special address not in LONG form, 
-          it must be 0x0 to 0xf without padding zeroes.`,
+            // eslint-disable-next-line max-len
+          `The given hex string ${args.input} is a special address not in LONG form, it must be 0x0 to 0xf without padding zeroes.`,
```
0xmaayan, line 250: we shouldn't need to disable this line. local formatter should know and add a break line  like https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/core/account_address.ts#L249

gregnazario: Unfortunately the way that this written before, it did not.  The new line won't handle the same way see:
The code gave this error, where there were extra spaces from the padding.  Otherwise, you'd have to put a `+` and a newline since it can't differentiate between spaces at the beginning and the end of the line.

```
The given hex string 0x000 is a special address not in LONG form, it must be 0x0 to 0xf without padding zeroes.
    + The given hex string 0x000 is a special address not in LONG form, 
    +           it must be 0x0 to 0xf without padding zeroes.

```

0xmaayan: not completely understand the problem, but lets deal with it now

0xmaayan (approved):

PR #9, opened 2023-10-11
[examples] Simple Javascript example for coin transfer

gregnazario:
### Description
The javascript example works and is able to do a coin transfer

### Test Plan
```
pnpm test 

> js-test@1.0.0 test /opt/git/typescript-sdk/examples/javascript/simple_transfer
> node transfer.js

This example will create two accounts (Alice and Bob), fund them, and transfer between them.
=== Addresses ===

Alice's address is: 0xfb37999c6f6932c63902eea6aaa12f3e3ebd0d2b545fc1a9e87c2349d0b86a72
Bob's address is: 0x65dd44a57d1f3a0feab1d7e7b57e16488e0b1fc51a614f7affe1e5acf5b14ead

=== Funding accounts ===

Alice's fund transaction:  38ce8e546efeffa5e81df3af725b864753a84ecf3fb4863fd49dd5d81837edf8
Bob's fund transaction:  43c8bc1a3dc90f53ba4f1fe9cbf625b563e793d05af047b4db4ef9c294a2f3eb

=== Balances ===

Alice's balance is: 100000000
Bob's balance is: 100

=== Transfer transaction ===

Committed transaction: 0xac7302d52d6229a0ffa3d0bd73dfa38099da803ff855822f420b26ff687fb66a

=== Balances after transfer ===

Alice's balance is: 100000000
Bob's balance is: 100

```


heliuchuan (approved):
lgtm

0xmaayan:
this is a huge PR, and has same fixes another  PR has. I am a bit worried merging it now in regards to conflicts etc

gregnazario:
> this is a huge PR, and has same fixes another PR has. I am a bit worried merging it now in regards to conflicts etc

Sure, we can break it up into 3 PRs, but unfortunately one of them is still going to be big.

gregnazario:
> this is a huge PR, and has same fixes another PR has. I am a bit worried merging it now in regards to conflicts etc

I've shrunk this down to just the javascript example now

examples/javascript/simple_transfer/transfer.js
```diff
@@ -0,0 +1,99 @@
...
+ * @param name
+ * @param address
+ * @returns {Promise<*>}
+ *
+ * TODO: Change to AccountAddress for address
+ */
+const balance = async (aptos, name, address) => {
+  let balance = await aptos.getAccountResource({ accountAddress: address, resourceType: APTOS_COIN });
+
+  let amount = Number(balance.data.coin.value);
```
0xjinn, line 24: I was doing this on my PR but got error saying `coin` does not exist in `data`

probably have to cast it? `Number((balance.data as { coin: { value: string } }).coin.value);`

0xjinn: Ahh didn't realize this is `js` file. Gotcha

examples/javascript/simple_transfer/transfer.js
```diff
@@ -0,0 +1,99 @@
...
+const example = async () => {
+  console.log("This example will create two accounts (Alice and Bob), fund them, and transfer between them.");
+
+  // Setup the client
+  const config = new AptosConfig({ network: Network.DEVNET });
+  const aptos = new Aptos(config);
+
+  // Create two accounts
+  let alice = Account.generate({ scheme: 0 });
+  let bob = Account.generate({ scheme: 0 });
```
0xjinn, line 39: nit: Should we use SigningScheme enum here, instead of directly using 0?

0xjinn (approved):

examples/javascript/simple_transfer/transfer.js
gregnazario: The enum only exists in Typescript, so unfortunately, it didn't work!  Had to use the number directly rather than the enum in Javascript.

It's been an interesting learning of Javascript vs Typescript when using them together

gregnazario: We may want both the constants and the enum then (since I was the one who pushed to replace the constants with an enum)

0xjinn: I like the enum as well, too bad for js i guess lol

0xmaayan: interesting.... we might want to have consts for js and enums for ts, we can probably build those "types" in separate files so if you use js import x.js

PR #11, opened 2023-10-11
Add internal transaction submission file

0xmaayan:
### Description
Just realized that previous PR was not completely full as we need to have those pure functions internally that handles the logic.  
This PR conforms to the sdk design in a way that everything in `api/` folder is a user facing interface functions and those in `internal/` folder are responsible on the actual logic.
more info https://github.com/aptos-labs/aptos-ts-sdk/pull/11/files#diff-a5a44d0986e0254ae62cd623aeaac609a288c9b2eabe5c741fd5b4a4f350a3ddR2

### Test Plan
`pnpm jest`

### DIFF
--- a/src/internal/transaction_submission.ts
+++ b/src/internal/transaction_submission.ts
@@ -1,0 +1,163 @@
+/**
+ * This file contains the underlying implementations for exposed API surface in
+ * the {@link api/transaction_submission}. By moving the methods out into a separate file,
+ * other namespaces and processes can access these methods without depending on the entire
+ * transaction_submission namespace and without having a dependency cycle error.
+ */
+
+import { AptosConfig } from "../api/aptos_config";
+import { postAptosFullNode } from "../client";
+import { Account } from "../core/account";
+import { AccountAuthenticator } from "../transactions/authenticator/account";
+import {
+  buildTransaction,
+  generateTransactionPayload,
+  generateSignedTransactionForSimulation,
+  generateSignedTransaction,
+  sign,
+} from "../transactions/transaction_builder/transaction_builder";
+import { GenerateTransactionInput, AnyRawTransaction, SimulateTransactionData } from "../transactions/types";
+import { UserTransactionResponse, PendingTransactionResponse, MimeType } from "../types";
+
+/**
+ * Generates any transaction by passing in the required arguments
+ *
+ * @param args.sender The transaction sender's account address as a HexInput
+ * @param args.data EntryFunctionData | ScriptData | MultiSigData
+ * @param feePayerAddress optional. For a fee payer (aka sponsored) transaction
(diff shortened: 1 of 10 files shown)


heliuchuan (commented):
We should remove the dupe func docs in internal and have /api be source of truth?

src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -180,12 +185,10 @@ export async function generateRawTransaction(args: {
  * When we call our `generateTransaction` function with the relevant type properties,
  * Typescript can infer the return type based on the appropriate function overload.
  */
-export async function generateTransaction(
-  args: GenerateSingleSignerRawTransactionArgs,
-): Promise<SingleSignerTransaction>;
-export async function generateTransaction(args: GenerateFeePayerRawTransactionArgs): Promise<FeePayerTransaction>;
-export async function generateTransaction(args: GenerateMultiAgentRawTransactionArgs): Promise<MultiAgentTransaction>;
-export async function generateTransaction(args: GenerateRawTransactionArgs): Promise<AnyRawTransaction>;
+export async function buildTransaction(args: GenerateSingleSignerRawTransactionArgs): Promise<SingleSignerTransaction>;
```
xbtmatt, line 188: Much better name!

src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -180,12 +185,10 @@ export async function generateRawTransaction(args: {
  * When we call our `generateTransaction` function with the relevant type properties,
```
xbtmatt, line 185:
```suggestion
 * When we call our `buildTransaction` function with the relevant type properties,
```

xbtmatt (approved):
Agree with @heliuchuan on consolidating the docs to one section although we can do it later and I think you have a better idea of the intended structure here

lgtm, `buildTransaction` is definitely a much better name!

src/internal/transaction_submission.ts
```diff
@@ -0,0 +1,163 @@
...
+ *  feePayerAddress?: AccountAddress
+ * }
+ * ```
+ */
+export async function generateTransaction(
+  args: { aptosConfig: AptosConfig } & GenerateTransactionInput,
+): Promise<AnyRawTransaction> {
+  const { aptosConfig, sender, data, options, secondarySignerAddresses, feePayerAddress } = args;
+  const payload = await generateTransactionPayload(data);
+  const rawTransaction = await buildTransaction({
```
gregnazario, line 67: This is different than buildTransaction, or should it also be named buildTransaction?

gregnazario (commented):
Can we fix the E2E transaction submission test I had to disable?

gregnazario (approved):

0xmaayan:
> Can we fix the E2E transaction submission test I had to disable?

the whole e2e transaction submission tests will get changed

PR #13, opened 2023-10-11
Enable commented out test

heliuchuan:
### Description
Use waitForTransaction to fix the mempool issue with the test.

### Test Plan
pnpm test


0xmaayan:
Thanks! we can now run the test with the newly added query! But I believe we have this fix locally as people work on transaction submission test suite.

PR #14, opened 2023-10-11
Use test constants keys/function address

heliuchuan:
### Description
Turns the commonly used test private keys and entry function addresses into consts

### Test Plan
pnpm test

### DIFF
--- a/tests/e2e/api/transaction_submission.test.ts
+++ b/tests/e2e/api/transaction_submission.test.ts
@@ -18,6 +18,7 @@ import {
   TransactionPayloadScript,
 } from "../../../src/transactions/instances/transactionPayload";
 import { SigningScheme } from "../../../src/types";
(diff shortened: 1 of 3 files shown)


tests/e2e/api/transaction_submission.test.ts
```diff
@@ -26,7 +27,7 @@ describe("transaction submission", () => {
       const aptos = new Aptos(config);
       const alice = Account.fromPrivateKey({
         privateKey: new Ed25519PrivateKey({
-          hexInput: "0x5aba8dab1c523be32bd4dafe2cc612f7f8050ce42a3322b60216ef67dc97768c",
+          hexInput: ALICE_PRIVATE_KEY,
```
0xmaayan, line 30: these tests are going to change. We now test it against devnet and eventually will change it to run against localnet. We had these fix accounts because we couldn't use the fund query so we fund it using the cli and "created" it on devnet. 
We will probably not use any fix private keys to create an account but create and fund it on run time

0xmaayan (changes requested):
these tests are going to change

PR #15, opened 2023-10-11
Add staking queries

heliuchuan:
### Description
Adds the getDelegatedStakingActivities and getNumberOfDelegators queries.  Also adds a query to get number of delegators for all pools.

### Test Plan
pnpm test


src/internal/staking.ts
```diff
@@ -0,0 +1,73 @@
...
+  const address = Hex.fromHexInput({ hexInput: poolAddress }).toString();
+  const query = {
+    query: GetNumberOfDelegators,
+    variables: { poolAddress: address },
+  };
+  const data: GetNumberOfDelegatorsQuery = await queryIndexer<GetNumberOfDelegatorsQuery>({ aptosConfig, query });
+  if (data.num_active_delegator_per_pool.length === 0) {
+    throw Error("Delegator pool not found");
+  }
+  return data.num_active_delegator_per_pool[0].num_active_delegator;
```
0xmaayan, line 40: [this](https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/types/indexer.ts#L43) how we handle/format the response. It is to deal with this weird scenario and also to support filtering. You can follow [getAccountOwnedTokens](https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/api/account.ts#L218)

src/api/staking.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  /**
+   * Queries delegated staking activities
+   *
+   * @param delegatorAddress Delegator address
+   * @param poolAddress Pool address
+   * @returns GetDelegatedStakingActivitiesResponse response type
+   */
+  async getDelegatedStakingActivities(args: {
```
0xmaayan, line 51: let's add orderBy support, like https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/api/account.ts#L223

src/api/staking.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.config = config;
+  }
+
+  /**
+   * Queries current number of delegators in a pool.  Throws an error if the pool is not found.
+   *
+   * @param poolAddress Pool address
+   * @returns The number of delegators for the given pool
+   */
+  async getNumberOfDelegators(args: { poolAddress: HexInput }): Promise<number> {
```
0xmaayan, line 28: looks like in other count like queries we return `{count:number}`, returning just the number might make more sense.

src/internal/queries/getNumberOfDelegatorsQuery.graphql
```diff
@@ -0,0 +1,8 @@
+query getNumberOfDelegators($poolAddress: String) {
```
0xmaayan, line 1: nit
this and getNumberOfDelegatorsForAllPools can be one query that accept optional where condition, and then the function can pass in the where, something like https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/providers/indexer.ts#L306

tests/e2e/api/staking.test.ts
```diff
@@ -0,0 +1,43 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { AptosConfig, Aptos } from "../../../src";
+import { Network } from "../../../src/utils/apiEndpoints";
+
+// Note: These test against main net and could fail.
```
gregnazario, line 7: Let's not have unit tests that could fail, they're used for CI

What I suggest is you can probably get a query for all pools, and then verify that you are getting values that match the expected data.

The data in the response is pretty irrelevant since it comes directly from the indexer

src/types/generated/types.ts
```diff
@@ -2461,6 +2455,8 @@ export type CurrentCollectionOwnershipV2ViewVarianceFields = {
 
 /** columns and relationships of "current_collections_v2" */
 export type CurrentCollectionsV2 = {
+  /** An object relationship */
+  cdn_asset_uris?: Maybe<NftMetadataCrawlerParsedAssetUris>;
```
gregnazario, line 2459: Did we remove these on purpose earlier?

heliuchuan: No clue, this is just generated via pnpm run indexer-codegen and then fmt

src/api/staking.ts
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Queries delegated staking activities
+   *
+   * @param delegatorAddress Delegator address
+   * @param poolAddress Pool address
+   * @returns GetDelegatedStakingActivitiesResponse response type
+   */
+  async getDelegatedStakingActivities(args: {
+    delegatorAddress: HexInput;
+    poolAddress: HexInput;
```
0xmaayan, line 62: should we also add orderBy here?

src/internal/account.ts
```diff
@@ -194,8 +191,14 @@ export async function getAccountTokensCount(args: {
     query: graphqlQuery,
     originMethod: "getAccountTokensCount",
   });
+  if (
+    !data.current_token_ownerships_v2_aggregate.aggregate ||
+    data.current_token_ownerships_v2_aggregate.aggregate.count === undefined
+  ) {
+    throw Error("Failed to get the count of account tokens");
```
0xmaayan, line 198: I think we can simply return `data.current_token_ownerships_v2_aggregate.aggregate.` and if there is no token, it would just be 0

src/internal/account.ts
```diff
@@ -352,7 +355,14 @@ export async function getAccountTransactionsCount(args: {
...
 
-  return data.account_transactions_aggregate.aggregate;
+  if (
+    !data.account_transactions_aggregate.aggregate ||
+    data.account_transactions_aggregate.aggregate.count === undefined
+  ) {
+    throw Error("Failed to get the count of account transactions");
+  }
+
+  return data.account_transactions_aggregate.aggregate.count;
```
0xmaayan, line 365:
```suggestion
  return data.account_transactions_aggregate.aggregate?.count;
```

src/api/staking.ts
heliuchuan: i think we can like p2, if there's a need for a use.  it wasn't in v1 and with object as param we can extend after i think

src/internal/account.ts
heliuchuan: don't quite understand.  to return `data.current_token_ownerships_v2_aggregate.aggregate` that would return type 
{count: number} | undefined.  We want to return type number here no?

src/internal/account.ts
heliuchuan: see other comment

src/internal/account.ts
0xmaayan: oh I am saying return `data.current_token_ownerships_v2_aggregate.aggregate?.number` and then I think it would return `0` or some number...?

heliuchuan: ah gotcha, I read the typing wrong.  you're right no need to check for count undefined.

src/internal/staking.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+import { AptosConfig } from "../api/aptos_config";
+import { Hex } from "../core";
+import { GetDelegatedStakingActivitiesResponse, GetNumberOfDelegatorsResponse, HexInput, OrderBy } from "../types";
+import { GetDelegatedStakingActivitiesQuery, GetNumberOfDelegatorsQuery } from "../types/generated/operations";
+import { GetDelegatedStakingActivities, GetNumberOfDelegators } from "../types/generated/queries";
+import { queryIndexer } from "./general";
+
+export async function getNumberOfDelegators(args: {
```
0xmaayan, line 18: I would rename here to `getNumberOfDelegatorsByPoolAddress` and the second one `getNumberOfDelegators`

src/internal/account.ts
```diff
@@ -194,8 +191,10 @@ export async function getAccountTokensCount(args: {
     query: graphqlQuery,
     originMethod: "getAccountTokensCount",
   });
-
-  return data.current_token_ownerships_v2_aggregate.aggregate;
+  if (!data.current_token_ownerships_v2_aggregate.aggregate) {
+    throw Error("Failed to get the count of account tokens");
```
0xmaayan, line 195: I guess I am just confused why we suddenly throw here? it doesn't seem like a pattern we use in other queries... also is it really a failure? because it is not a server error...

0xmaayan (approved):

src/internal/account.ts
heliuchuan: I don't think it will actually ever practically throw, but the generated type data.current_token_ownerships_v2_aggregate.aggregate is optional so you need a check.  We can revisit later

PR #16, opened 2023-10-11
[tests] Combine and shorten imports where applicable

gregnazario:
### Description
We export many of our types and functions up to the top level, which is where all of them should be in the end for the SDK. This is a good way to view the user experience when using the SDK.

### Test Plan
```
pnpm fmt && pnpm lint && pnpm test
```


heliuchuan (approved):
lgtm

PR #18, opened 2023-10-11
[api] Memoize gas estimation and module APIs

gregnazario:
### Description
Specifically only memoize the common use case, and do not memoize ledger versions specifically.

### Test Plan
This one is a hard one to test without inspecting the requests directly.  Once we have more E2E examples, I will watch and inspect the requests.

### DIFF
--- a/src/api/transaction.ts
+++ b/src/api/transaction.ts
@@ -11,6 +11,7 @@ import {
 } from "../internal/transaction";
 import { AnyNumber, GasEstimation, HexInput, PaginationArgs, TransactionResponse } from "../types";
 import { AptosConfig } from "./aptos_config";
+import { memoizeAsync } from "../utils/memoize";
 
 export class Transaction {
   readonly config: AptosConfig;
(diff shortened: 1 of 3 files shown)


heliuchuan (approved):

src/api/account.ts
```diff
@@ -109,7 +110,18 @@ export class Account {
     moduleName: string;
     options?: LedgerVersion;
   }): Promise<MoveModuleBytecode> {
-    const module = await getModule({ aptosConfig: this.config, ...args });
+    // We don't memoize the account module by ledger version, as it's not a common use case, this would be handled
+    // by the developer directly
+    if (args.options?.ledgerVersion !== undefined) {
```
0xmaayan, line 115: I think we should memoize the internal functions. other sdk functions use the internal queries

gregnazario: Done

src/api/transaction.ts
```diff
@@ -129,9 +130,14 @@ export class Transaction {
    * ```
    */
   async getGasPriceEstimation(): Promise<GasEstimation> {
-    const gasEstimation = await getGasPriceEstimation({
-      aptosConfig: this.config,
-    });
+    const gasEstimation = await memoizeAsync(
```
0xmaayan, line 133: remove it?

PR #20, opened 2023-10-11
Add built-in `transferCoinTransaction()` function

0xmaayan:
### Description
Adds a built in `transferCoinTransaction` to `Aptos`.
```
const aptos = new Aptos()
const transaction = await aptos.transferCoinTransaction()
```

This function generates a transfer coin transaction that can be simulated and/or sign and submit. That is to have support for wallets and developers.

In addition, added a `signAndSubmitTransaction` API function for a single signer transaction to ease the user submission flow

### Test Plan
```
coin
    ✓ it generates a transfer coin transaction with AptosCoin coin type (802 ms)
    ✓ it generates a transfer coin transaction with a custom coin type (481 ms)
    ✓ it transfers APT coin aomunt from sender to recipient (1609 ms)
```

### DIFF
--- a/src/api/account.ts
+++ b/src/api/account.ts
@@ -299,10 +299,10 @@ export class Account {
   }
 
   /**
-   * Queries the count of an account's coins
+   * Queries the count of an account's coins aggregated
    *
    * @param accountAddress The account address we want to get the total count for
-   * @returns An object { count : number }
+   * @returns An object { count : number } where `number` is the aggregated count of all account's coin
    */
   async getAccountCoinsCount(args: { accountAddress: HexInput }): Promise<GetAccountCoinsCountResponse> {
     const count = getAccountCoinsCount({ aptosConfig: this.config, ...args });
--- a/src/api/coin.ts
+++ b/src/api/coin.ts
@@ -1,0 +1,40 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Account } from "../core";
+import { transaferCoinTransaction } from "../internal/coin";
+import { SingleSignerTransaction, GenerateTransactionOptions } from "../transactions/types";
+import { AnyNumber, HexInput, MoveResourceType } from "../types";
+import { AptosConfig } from "./aptos_config";
+
+/**
+ * A class to handle all `Coin` operations
+ */
+export class Coin {
+  readonly config: AptosConfig;
+
+  constructor(config: AptosConfig) {
+    this.config = config;
+  }
+
+  /**
+   * Generate a transfer coin transaction that can be simulated and/or signed and submitted
+   *
+   * @param args.sender The sender account
+   * @param args.recipient The recipient address
+   * @param args.amount The amount to transfer
+   * @param args.coinType optional. The coin struct type to transfer. Default to 0x1::aptos_coin::AptosCoin
+   *
+   * @returns SingleSignerTransaction
+   */
+  async transaferCoinTransaction(args: {
+    sender: Account;
+    recipient: HexInput;
+    amount: AnyNumber;
+    coinType?: MoveResourceType;
+    options?: GenerateTransactionOptions;
+  }): Promise<SingleSignerTransaction> {
+    const response = await transaferCoinTransaction({ aptosConfig: this.config, ...args });
+    return response;
+  }
+}
--- a/src/api/coin.ts
+++ b/src/api/coin.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { Account } from "../core";
-import { transaferCoinTransaction } from "../internal/coin";
+import { transferCoinTransaction } from "../internal/coin";
 import { SingleSignerTransaction, GenerateTransactionOptions } from "../transactions/types";
 import { AnyNumber, HexInput, MoveResourceType } from "../types";
 import { AptosConfig } from "./aptos_config";
@@ -27,14 +27,14 @@ export class Coin {
    *
    * @returns SingleSignerTransaction
    */
-  async transaferCoinTransaction(args: {
+  async transferCoinTransaction(args: {
     sender: Account;
     recipient: HexInput;
     amount: AnyNumber;
     coinType?: MoveResourceType;
     options?: GenerateTransactionOptions;
   }): Promise<SingleSignerTransaction> {
-    const response = await transaferCoinTransaction({ aptosConfig: this.config, ...args });
+    const response = await transferCoinTransaction({ aptosConfig: this.config, ...args });
     return response;
   }
 }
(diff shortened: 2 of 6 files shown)


tests/e2e/api/coin.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+describe("coin", () => {
+  test("it generates a transfer coin transaction with AptosCoin coin type", async () => {
+    const config = new AptosConfig({ network: Network.DEVNET });
+    const aptos = new Aptos(config);
+    const sender = Account.generate({ scheme: SigningScheme.Ed25519 });
+    const recipient = Account.generate({ scheme: SigningScheme.Ed25519 });
+    await aptos.fundAccount({ accountAddress: sender.accountAddress.toString(), amount: 100000000 });
+
+    const transaction = await aptos.transaferCoinTransaction({
```
gregnazario, line 15:
```suggestion
    const transaction = await aptos.transferCoinTransaction({
```

tests/e2e/api/coin.test.ts
```diff
@@ -0,0 +1,76 @@
...
+      sender,
+      recipient: recipient.accountAddress.toString(),
+      amount: 10,
+    });
+
+    const txnDeserializer = new Deserializer(transaction.rawTransaction);
+    const rawTransaction = RawTransaction.deserialize(txnDeserializer);
```
gregnazario, line 22: Hmm, this is an interesting way of handling it, where you need to deserialize by passing the deserializer in.  Might be interesting if we also just provide a way to deserialize the bytes directly (if you only have that).

tests/e2e/api/coin.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  });
+
+  test("it generates a transfer coin transaction with a custom coin type", async () => {
+    const config = new AptosConfig({ network: Network.DEVNET });
+    const aptos = new Aptos(config);
+    const sender = Account.generate({ scheme: SigningScheme.Ed25519 });
+    const recipient = Account.generate({ scheme: SigningScheme.Ed25519 });
+    await aptos.fundAccount({ accountAddress: sender.accountAddress.toString(), amount: 100000000 });
+
+    const transaction = await aptos.transaferCoinTransaction({
```
gregnazario, line 36:
```suggestion
    const transaction = await aptos.transferCoinTransaction({
```

tests/e2e/api/coin.test.ts
```diff
@@ -0,0 +1,76 @@
...
+      sender,
+      recipient: recipient.accountAddress.toString(),
+      amount: 10,
+    });
+    const response = await aptos.signAndSubmitTransaction({ signer: sender, transaction });
+
+    await waitForTransaction({ aptosConfig: config, txnHash: response.hash });
+
+    const recipientCoins = await aptos.getAccountCoinsData({ accountAddress: recipient.accountAddress.toString() });
+    const senderCoinsAfter = await aptos.getAccountCoinsData({ accountAddress: sender.accountAddress.toString() });
```
gregnazario, line 70: We may want to consider unioning `HexInput` with `AccountAddress` or only taking in `AccountAddress`, since this seems to be a common pattern that is being repeated around.

The wallet adapter can be less stringent probably.

tests/e2e/api/coin.test.ts
```diff
@@ -0,0 +1,76 @@
...
+import { TypeTagStruct } from "../../../src/transactions/typeTag/typeTag";
+import { SigningScheme } from "../../../src/types";
+
+describe("coin", () => {
+  test("it generates a transfer coin transaction with AptosCoin coin type", async () => {
+    const config = new AptosConfig({ network: Network.DEVNET });
+    const aptos = new Aptos(config);
+    const sender = Account.generate({ scheme: SigningScheme.Ed25519 });
+    const recipient = Account.generate({ scheme: SigningScheme.Ed25519 });
+    await aptos.fundAccount({ accountAddress: sender.accountAddress.toString(), amount: 100000000 });
```
gregnazario, line 13: I suggest we make these numbers constants for the test

src/internal/coin.ts
```diff
@@ -0,0 +1,32 @@
+import { AptosConfig } from "../api/aptos_config";
+import { U64 } from "../bcs/serializable/move-primitives";
+import { Account, AccountAddress } from "../core";
+import { GenerateTransactionOptions, SingleSignerTransaction } from "../transactions/types";
+import { StructTag, TypeTagStruct } from "../transactions/typeTag/typeTag";
+import { HexInput, AnyNumber, MoveResourceType } from "../types";
+import { APTOS_COIN } from "../utils/const";
+import { generateTransaction } from "./transaction_submission";
+
+export async function transaferCoinTransaction(args: {
```
gregnazario, line 10: So, we can't call it just `transfer` because everything is unioned together on a single interface?

src/api/account.ts
```diff
@@ -299,10 +299,10 @@ export class Account {
...
   /**
-   * Queries the count of an account's coins
+   * Queries the count of an account's coins aggregated
    *
    * @param accountAddress The account address we want to get the total count for
-   * @returns An object { count : number }
+   * @returns An object { count : number } where `number` is the aggregated count of all account's coin
    */
   async getAccountCoinsCount(args: { accountAddress: HexInput }): Promise<GetAccountCoinsCountResponse> {
     const count = getAccountCoinsCount({ aptosConfig: this.config, ...args });
```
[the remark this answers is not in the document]
0xmaayan: this is an existing query in sdk v1. I believe it returns the former. can validate it in a second PR as it is not really relevant to this PR

src/api/coin.ts
```diff
@@ -0,0 +1,40 @@
...
+   * Generate a transfer coin transaction that can be simulated and/or signed and submitted
+   *
+   * @param args.sender The sender account
+   * @param args.recipient The recipient address
+   * @param args.amount The amount to transfer
+   * @param args.coinType optional. The coin struct type to transfer. Default to 0x1::aptos_coin::AptosCoin
+   *
+   * @returns SingleSignerTransaction
+   */
+  async transaferCoinTransaction(args: {
```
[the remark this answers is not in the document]
0xmaayan: och I really need to het a typo plugin or something

tests/e2e/api/coin.test.ts
0xmaayan: I believe @xbtmatt has some thoughts about it

tests/e2e/api/coin.test.ts
0xmaayan: I think most users will get this data from some servers, that will return the address as a string (or in some cases as Uint8Array) - that was the thought behind accepting HexInput as an input and not force them initialize `AccountAddress` when they want to query for an address

tests/e2e/api/coin.test.ts
0xmaayan: yes, we need a test framework. will work on it after we have full test cover to our code (not a P0 for alpha version)

src/internal/coin.ts
0xmaayan: in the future, we will have support for FA, NFT, Object, etc. since they all under the same namespace we can't call each `transfer`. we can always but each related transfer function under the each namespace, but that's just confusing for the user imo.
also, I think transferCoin makes sense, as this function transfers coin

gregnazario (approved):

PR #21, opened 2023-10-11
[SDKv2] Add lookupOriginalAccountAddress query

0xjinn:
### Description

1. Add lookupOriginalAccountAddress query
2. Add e2e test

Note: I will add additional e2e to this method when rotate key is ready.

### Test Plan

pnpm fmt & pnpm test

### DIFF
--- a/src/api/account.ts
+++ b/src/api/account.ts
@@ -38,7 +38,9 @@ import {
   getResource,
   getResources,
   getTransactions,
+  lookupOriginalAccountAddress,
 } from "../internal/account";
+import { AccountAddress } from "../core";
 
 /**
  * A class to query all `Account` related queries on Aptos.
@@ -174,6 +176,20 @@ export class Account {
   }): Promise<MoveResource> {
     const resource = await getResource({ aptosConfig: this.config, ...args });
     return resource;
+  }
+
+  /**
+   * Lookup the original address by the current derived address or authentication key
+   *
+   * @param args.addressOrAuthKey The derived address or authentication key
+   * @returns Promise<AccountAddress> The original address
+   */
+  async lookupOriginalAccountAddress(args: {
+    authenticationKey: HexInput;
+    options?: LedgerVersion;
+  }): Promise<AccountAddress> {
+    const address = await lookupOriginalAccountAddress({ aptosConfig: this.config, ...args });
+    return address;
   }
 
   /**
--- a/src/internal/account.ts
+++ b/src/internal/account.ts
@@ -9,9 +9,9 @@
  */
 
 import { AptosConfig } from "../api/aptos_config";
-import { getAptosFullNode, paginateWithCursor } from "../client";
+import { AptosApiError, getAptosFullNode, paginateWithCursor } from "../client";
 import { AccountAddress, Hex } from "../core";
-import { queryIndexer } from "./general";
+import { getTableItem, queryIndexer } from "./general";
 import {
   AccountData,
   GetAccountCoinsCountResponse,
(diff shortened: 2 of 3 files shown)


0xjinn:
* main
  * **PR #21** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-ts-sdk/21" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-ts-sdk/21?utm_source=stack-comment).

0xmaayan:
> Current dependencies on/for this PR:
> 
> * main
>   
>   * **PR [[SDKv2] Add lookupOriginalAccountAddress query #21](https://github.com/aptos-labs/aptos-ts-sdk/pull/21)** [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-ts-sdk/21)  👈
> 
> This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-ts-sdk/21?utm_source=stack-comment).

so they generate message even if there is no dependency PR? or im reading it wrong?

src/internal/account.ts
```diff
@@ -169,6 +169,47 @@ export async function getResource(args: {
   return data;
 }
 
+export async function lookupOriginalAccountAddress(args: {
+  aptosConfig: AptosConfig;
+  addressOrAuthKey: HexInput;
+  options?: LedgerVersion;
+}): Promise<AccountAddress> {
```
0xmaayan, line 176: can't we just return the string instead of converting it into AccountAddress?

0xjinn:
> > Current dependencies on/for this PR:
> > 
> > * main
> >   
> >   * **PR [[SDKv2] Add lookupOriginalAccountAddress query #21](https://github.com/aptos-labs/aptos-ts-sdk/pull/21)** [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-ts-sdk/21)  👈
> > 
> > This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-ts-sdk/21?utm_source=stack-comment).
> 
> so they generate message even if there is no dependency PR? or im reading it wrong?

@0xmaayan Yep Graphite does that, regardless of # of stacked PRs. I believe they drop everything into a `stack`.

tests/e2e/api/account.test.ts
```diff
@@ -79,6 +80,20 @@ describe("account api", () => {
       expect(data).toHaveProperty("type");
       expect(data.type).toBe("0x1::account::Account");
     });
+
+    test("lookupOriginalAccountAddress - Look up account address before key rotation", async () => {
+      const config = new AptosConfig({ network: Network.LOCAL });
+      const aptos = new Aptos(config);
+      const account = Account.generate({ scheme: SigningScheme.Ed25519 });
+
+      // Fund and create account onchain
```
0xmaayan, line 89: nit
```suggestion
     
```

src/internal/account.ts
0xjinn: Hmm, personally I prefer AccountAddress since it follows OOD better. Similar to the Account.Sign method, which returns the Signature instead of hex data. 

With the AccountAddress, user can chose to convert to String (with or without prefix) or U8intArray easier.

0xmaayan: hmm I think it is a bit different, the query response is just the address, then we might want to return whatever the server returns....

0xjinn: Or I can have the this query to return a raw string, but having `lookupOriginalAccountAddress` method from the api/account to return `AccountAddress`?

src/internal/account.ts
```diff
@@ -169,6 +169,47 @@ export async function getResource(args: {
   return data;
 }
 
+export async function lookupOriginalAccountAddress(args: {
+  aptosConfig: AptosConfig;
+  addressOrAuthKey: HexInput;
```
gregnazario, line 174: We should just rename this to `authenticationKey`, you technically always lookup an address by the auth key

src/api/account.ts
```diff
@@ -176,6 +178,20 @@ export class Account {
...
   }
 
+  /**
+   * Lookup the original address by the current derived address or authentication key
+   *
+   * @param args.addressOrAuthKey The derived address or authentication key
+   * @returns Promise<AccountAddress> The original address
+   */
+  async lookupOriginalAccountAddress(args: {
+    addressOrAuthKey: HexInput;
```
gregnazario, line 188: similarly

src/internal/account.ts
xbtmatt: I generally agree with returning `AccountAddress`, especially since we're now stringently parsing everything and it should all be kosher in terms of its string representation with `.toString()`

0xjinn:
1. Updated param naming
2. Fixed merging conflict

gregnazario (approved):

PR #24, opened 2023-10-11
[exports] Add more exports to allow using the SDK

gregnazario:
### Description
As part of the external examples, adding the exports as the unit tests can't catch that these exports aren't visible outside

### Test Plan
I have two example PRs that are upcoming and dependent on this

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,3 +6,6 @@ export * from "./api";
 export * from "./client";
 export * from "./utils/apiEndpoints";
 export * from "./bcs";
+export * from "./types";
+export * from "./transactions/types";
+export * from "./transactions/typeTag/typeTag";
--- a/src/bcs/index.ts
+++ b/src/bcs/index.ts
@@ -3,3 +3,4 @@
 
 export * from "./deserializer";
 export * from "./serializer";
+export * from "./serializable/move-primitives";


src/index.ts
```diff
@@ -6,3 +6,6 @@ export * from "./api";
 export * from "./client";
 export * from "./utils/apiEndpoints";
 export * from "./bcs";
+export * from "./types";
```
0xmaayan, line 9: we need to export only `types/index` and `type/indexer`. exporting everything will get us with a huge package size, and that is the whole point in how we structure indexer type

gregnazario: Done

src/index.ts
```diff
@@ -6,3 +6,7 @@ export * from "./api";
 export * from "./client";
 export * from "./utils/apiEndpoints";
 export * from "./bcs";
+export * from "./types/index";
+export * from "./types/indexer";
```
0xmaayan, line 10: oh shit lol

src/index.ts
```diff
@@ -6,3 +6,6 @@ export * from "./api";
 export * from "./client";
 export * from "./utils/apiEndpoints";
 export * from "./bcs";
+export * from "./types/index";
```
0xmaayan, line 9: yep makes sense

0xmaayan (approved):

PR #25, opened 2023-10-12
[ci] Update codeowners and ignore files

gregnazario:
### Description
This should add the users individually for now, and additionally prevent extra stuff from making it into builds.

### Test Plan
CI, and I ran all commands individually


0xmaayan (approved):
noice!

PR #28, opened 2023-10-12
[Examples] Add Typescript and Typescript ESM examples

gregnazario:
### Description

Simple examples to show that Typescript and Typescript ESM can be used with the SDK with no issue.

### Test Plan
 
Each works by running `pnpm test` in each.  Next step would be to add these examples into the CI system once there is not as much churn.


xbtmatt:
Hm, I'm getting dependency errors:

`transfer.ts:15:8 - error TS2307: Cannot find module 'aptos' or its corresponding type declarations.`

my directory is there, but just not working for me, I'm running it from the `simple_transfer` directory in both cases, am I just doing it wrong ? can never tell when it's a package manager error

examples/typescript-esm/simple_transfer/transfer.ts
```diff
@@ -0,0 +1,107 @@
...
+  Aptos,
+  AptosConfig,
+  Network,
+  SigningScheme,
+  StructTag,
+  TypeTagStruct,
+  U64,
+} from "aptos";
+
+const APTOS_COIN = "0x1::coin::CoinStore<0x1::aptos_coin::AptosCoin>";
```
0xmaayan, line 17: nit
we have this const in `aptos`

examples/typescript-esm/simple_transfer/transfer.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Prints the balance of an account
+ * @param aptos
+ * @param name
+ * @param address
+ * @returns {Promise<*>}
+ *
+ */
+const balance = async (aptos: Aptos, name: string, address: AccountAddress) => {
+  let resource = await aptos.getAccountResource({ accountAddress: address.toUint8Array(), resourceType: APTOS_COIN });
+  let amount = Number((resource.data as { coin: { value: string } }).coin.value);
```
0xmaayan, line 32: och

examples/typescript-esm/simple_transfer/transfer.ts
```diff
@@ -0,0 +1,107 @@
...
+
+const example = async () => {
+  console.log("This example will create two accounts (Alice and Bob), fund them, and transfer between them.");
+
+  // Setup the client
+  const config = new AptosConfig({ network: Network.DEVNET });
+  const aptos = new Aptos(config);
+
+  // Create two accounts
+  let alice = Account.generate({ scheme: SigningScheme.Ed25519 });
```
0xmaayan, line 46: wonder if we should default to a ed25519 scheme

examples/typescript-esm/simple_transfer/transfer.ts
```diff
@@ -0,0 +1,107 @@
...
+  // Show the balances
+  console.log("\n=== Balances ===\n");
+  let aliceBalance = await balance(aptos, "Alice", alice.accountAddress);
+  let bobBalance = await balance(aptos, "Bob", bob.accountAddress);
+
+  if (aliceBalance !== ALICE_INITIAL_BALANCE) throw new Error("Alice's balance is incorrect");
+  if (bobBalance !== BOB_INITIAL_BALANCE) throw new Error("Bob's balance is incorrect");
+
+  // Transfer between users
+  const txn = await aptos.transactionSubmission.generateTransaction({
```
0xmaayan, line 77: can simply do
`aptos.generateTransaction`
also, why not use `aptos.transferCoinTransaction`?

gregnazario: Yeah, I wrote this before that.  We can change it but this does point out some annoyances around parsing output types.  So we'll probably want a similar example that doesn't use the coin transfer

gregnazario:
> Hm, I'm getting dependency errors:
> 
> `transfer.ts:15:8 - error TS2307: Cannot find module 'aptos' or its corresponding type declarations.`
> 
> my directory is there, but just not working for me, I'm running it from the `simple_transfer` directory in both cases, am I just doing it wrong ? can never tell when it's a package manager error

It uses a link dependency (I'm no expert of typescript build systems).

So, because it uses a link dependency you have to run `pnpm build` in the main repo, then run that one.

I suspect it can be cleaned up need to do a little research into pnpm

examples/typescript-esm/simple_transfer/transfer.ts
gregnazario: We probably should, since we would push for it over some others for security at this time

examples/typescript-esm/simple_transfer/transfer.ts
gregnazario: It's unfortunately not exported.  Will do on another round of exports in a different PR.

examples/typescript-esm/simple_transfer/transfer.ts
gregnazario: This is definitely a user experience we can fix.

PR #34, opened 2023-10-12
[Tests] Transaction submission test suite. The initial attempt at testing all authenticators, payloads and argument types 

xbtmatt:
## Test suite

TODO:
- [x] Add 1 - n signers for the functions
- [x] Verify the order of signers in the contract with assertions

### Argument types
- [x] basic argument types
- [x] vector argument types
- [x] option argument types

### Authenticators & payloads

**Ed25519** (1 sender, 1 signer)
- [x] public entry function
- [x] private entry function
- [ ] script

**Secp256k1** (1 sender, 1 signer)
- [ ] public entry function
- [ ] private entry function
- [ ] script

**MultiEd25519** (k-of-n signers)
- TODO

**Multi Agent** (1 sender, n - 1 secondary signers, n signers)
- [x] public entry function Ed25519
- [ ] private entry function Ed25519
- [ ] script Ed25519
- [ ] public entry function Secp256k1
- [ ] private entry function Secp256k1
- [ ] script Secp256k1

**Fee Payer** (1 sender, 1 fee payer, 0 - n signers)
- [x] public entry function Ed25519
- [ ] private entry function Ed25519
- [ ] script Ed25519
- [ ] public entry function Secp256k1
- [ ] private entry function Secp256k1
- [ ] script Secp256k1

### Miscellaneous:
- [ ] view function of all the argument types listed above
- [ ] type tags, including various nested generics


src/bcs/serializable/move-structs.ts
```diff
@@ -439,7 +439,7 @@ export class MoveObject extends Serializable implements TransactionArgument {
     if (value instanceof AccountAddress) {
       this.value = value;
     } else {
-      this.value = AccountAddress.fromHexInput({ input: value });
+      this.value = AccountAddress.fromHexInputRelaxed({ input: value });
```
gregnazario, line 442: Why is this changing?

xbtmatt: I would like to use relaxed if we can, I was running into a bunch of errors because of it- basically what we were talking about with daniel

I was just putting it here as a fix so I didn't have to fix every instance of an object in the test suite (I wrote it down as feedback/things to think about in the doc), but yea, the issue was I was getting errors from addresses I was getting from the API on chain

xbtmatt:
TODO: I think we could add something like these (just with single signer, ed25519) as a `bcs-examples.ts` file to show people how to use the new stuff, since it is very comprehensive

tests/e2e/api/transaction_arguments.test.ts
```diff
@@ -0,0 +1,390 @@
...
+import { TypeTag } from "../../../src/transactions/typeTag/typeTag";
+import { EntryFunctionArgumentTypes } from "../../../src/transactions/types";
+import {
+  SigningScheme,
+  TransactionFeePayerSignature,
+  TransactionMultiAgentSignature,
+  UserTransactionResponse,
+} from "../../../src/types";
+
+jest.setTimeout(30000);
```
gregnazario, line 24: Is there a test that's actually taking 30 seconds?

tests/e2e/api/transaction_arguments.test.ts
```diff
@@ -0,0 +1,390 @@
...
+  let rawTxnMultiAgentHelper: (...args: any[]) => Promise<UserTransactionResponse>;
+
+  beforeAll(async () => {
+    const config = new AptosConfig({ network: Network.LOCAL });
+    aptos = new Aptos(config);
+    account1 = Account.generate();
+    account2 = Account.generate();
+    account3 = Account.generate();
+    account4 = Account.generate();
+    account5 = Account.generate();
```
gregnazario, line 57: Probably would have been easier just to have an array of accounts.

tests/e2e/api/transaction_arguments.test.ts
```diff
@@ -0,0 +1,390 @@
...
+    accountFeePayer = Account.generate();
+    const accounts = [account1, account2, account3, account4, account5, accountFeePayer];
+    for (const account of accounts) {
+      await aptos.fundAccount({
+        accountAddress: account.accountAddress.data,
+        amount: 100_000_000,
```
gregnazario, line 63: We have a constant for this now

tests/e2e/api/transaction_arguments.test.ts
```diff
@@ -0,0 +1,390 @@
...
+            sender: sender.accountAddress.toString(),
+            data: {
+              function: `0x${account1.accountAddress.toStringWithoutPrefix()}::tx_args_module::${functionName}`,
+              type_arguments: typeArgs,
+              arguments: args,
+            },
+            secondarySignerAddresses: secondarySignerAddresses.map((address) => address.data),
+          });
+        } else {
+          return await aptos.generateTransaction({
```
gregnazario, line 170: :/ super fragile, but we'll have to come back

tests/e2e/api/transaction_arguments.test.ts
```diff
@@ -0,0 +1,390 @@
...
+              arguments: args,
+            },
+            feePayerAddress: feePayerAddress!.data,
+          });
+          return response;
+        } else if (feePayerAddress === undefined) {
+          return await aptos.generateTransaction({
+            sender: sender.accountAddress.toString(),
+            data: {
+              function: `0x${account1.accountAddress.toStringWithoutPrefix()}::tx_args_module::${functionName}`,
```
gregnazario, line 163: Why not just do `toString()` with the prefix instead of adding the 0x?

tests/e2e/api/transaction_arguments.test.ts
```diff
@@ -0,0 +1,390 @@
...
+            },
+            secondarySignerAddresses: secondarySignerAddresses.map((address) => address.data),
+            feePayerAddress: feePayerAddress.data,
+          });
+        }
+      })();
+
+      const accounts = [sender, account2, account3, account4, account5];
+      // sign wtih an account its address is in secondarySignerAddresses OR if it's the sender
+      // TODO: Fix the ugly
```
gregnazario, line 185: :/

tests/e2e/api/transaction_arguments.test.ts
```diff
@@ -0,0 +1,390 @@
...
+            transaction: transaction,
+          });
+        });
+
+      const feePayerAuthenticator = feePayerAddress
+        ? aptos.signTransaction({
+            signer: accountFeePayer,
+            transaction: transaction,
+          })
+        : undefined;
```
gregnazario, line 204: Ternarys are usually frowned upon for readability purposes

gregnazario (approved):
This is a lot of code.  Let's please break this up into functions in the future for each step rather than have a really really big `beforeAll`

tests/e2e/api/transaction_arguments.test.ts
xbtmatt: No, I think I just changed it to 30 to make sure it wouldn't fail, I think I can do like 10 or 15

tests/e2e/api/transaction_arguments.test.ts
xbtmatt: Yeah, the only reason I did it like this is because the types weren't resolving correctly because of the `?: undefined`, otherwise it should have correctly resolved to a general type I could use

tests/e2e/api/transaction_arguments.test.ts
xbtmatt: was wondering why I would ever do this and remembered the issue is the type is that `0x${...}::${...}` type, so it gets mad otherwise

0xmaayan (approved):
it is a big test. maybe we should put all transactions tests under a transactions folder. can also have a helper file for transaction specific tests

src/bcs/serializable/move-structs.ts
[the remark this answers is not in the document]
xbtmatt: Because I was getting objects from the API, it returned it with no leading 0, meaning I needed to sanitize my input. I then realized I don't see in any case why someone *wouldn't* want the "relaxed" version when using it as input.

You generally would only use a `MoveObject` to serialize an entry function argument, so the end serialization result is the same regardless of whether or not you input a `0x0...` or `0x...`, so `fromHexInputRelaxed` makes more sense to me than having people run into errors 1/16 times because they didn't expect the address they passed in to to fail the parsing test

tests/e2e/api/transaction_arguments.test.ts
```diff
@@ -0,0 +1,390 @@
...
+    });
+    const signedTxn = await aptos.signTransaction({
+      signer: account1,
+      transaction: rawTxn,
+    });
+    const txnHash = await aptos.submitTransaction({
+      transaction: rawTxn,
+      senderAuthenticator: signedTxn,
+    });
+    const _ = await aptos.waitForTransaction({
```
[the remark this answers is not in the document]
xbtmatt: no I'm not sure why I did that tbh, removed it

PR #35, opened 2023-10-12
Add getChainTopUserTransactions.graphql

0xjinn:
### Description

Add the query getChainTopUserTransactions and e2e test

### Test Plan

pnpm test
pnpm fmt

### DIFF
--- a/src/types/generated/operations.ts
+++ b/src/types/generated/operations.ts
@@ -331,6 +331,12 @@ export type GetAccountTransactionsCountQuery = {
   account_transactions_aggregate: { aggregate?: { count: number } | null };
 };
 
+export type GetChainTopUserTransactionsQueryVariables = Types.Exact<{
+  limit?: Types.InputMaybe<Types.Scalars["Int"]>;
+}>;
+
+export type GetChainTopUserTransactionsQuery = { user_transactions: Array<{ version: any }> };
+
 export type GetCollectionDataQueryVariables = Types.Exact<{
   where_condition: Types.CurrentCollectionsV2BoolExp;
 }>;
--- a/tests/e2e/api/general.test.ts
+++ b/tests/e2e/api/general.test.ts
@@ -103,4 +103,12 @@ describe("general api", () => {
 
     expect(chainId.ledger_infos[0].chain_id).toBe(2);
   });
+
+  test("it should fetch chain top user transactions", async () => {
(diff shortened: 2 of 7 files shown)


0xjinn:
* main
  * **PR #35** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-ts-sdk/35" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-ts-sdk/35?utm_source=stack-comment).

gregnazario (commented):
I'm wondering how important this is to be in the SDK?

I'm assuming this is only useful for explorers and statistic sites

0xjinn:
> I'm wondering how important this is to be in the SDK?
> 
> I'm assuming this is only useful for explorers and statistic sites

@gregnazario Good question. Yeah I don't think this is that much of a P0 query. Shall we revisit the priority on these queries? - https://www.notion.so/aptoslabs/fetch-queries-to-port-over-into-sdk-v2-8a38cf2c47e542e2ae299c900894277b

src/types/generated/operations.ts
```diff
@@ -1,336 +1,93 @@
-import * as Types from "./types";
+import * as Types from './types';
 
-export type CurrentTokenOwnershipFieldsFragment = {
-  token_standard: string;
-  token_properties_mutated_v1?: any | null;
-  token_data_id: string;
-  table_type_v1?: string | null;
```
gregnazario, line 7: Any reason this file just disappears?

tests/e2e/api/general.test.ts
```diff
@@ -103,4 +103,12 @@ describe("general api", () => {
 
     expect(chainId.ledger_infos[0].chain_id).toBe(2);
   });
+
+  test("it should fetch chain top user transactions", async () => {
+    const config = new AptosConfig({ network: Network.TESTNET });
```
gregnazario, line 108: This should also be local

gregnazario (approved):
Let's please get these PRs in, and we may think about which queries we need in the future on here.

src/types/generated/operations.ts
0xjinn: hmm wait, which file? or you mean the `CurrentTokenOwnershipFieldsFragment` type? I think it got re-formatted

PR #39, opened 2023-10-13
Add collection queries

heliuchuan:
### Description
Adds collection related queries to the SDK

### Test Plan
pnpm test - tests against a random collection I selected from mainnet

### DIFF
--- a/src/internal/collection.ts
+++ b/src/internal/collection.ts
@@ -54,7 +54,7 @@ export async function getCollectionData(args: {
   const data = await queryIndexer<GetCollectionDataQuery>({
     aptosConfig,
     query: graphqlQuery,
-    originMethod: "getAccountCollectionsWithOwnedTokens",
+    originMethod: "getCollectionData",
   });
 
   if (data.current_collections_v2.length === 0) {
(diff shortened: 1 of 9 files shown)


heliuchuan:
why does pnpm fmt not work? ....

src/internal/collection.ts
```diff
@@ -0,0 +1,89 @@
...
+    originMethod: "getCollectionData",
+  });
+
+  if (data.current_collections_v2.length === 0) {
+    throw Error("Collection not found");
+  }
+
+  if (data.current_collections_v2.length > 1) {
+    throw Error(
+      "More than one collection wirh the same name found.  Use args.options.tokenStandard to specify v1 or v2",
```
0xmaayan, line 66:
```suggestion
      "More than one collection with the same name found.  Use args.options.tokenStandard to specify v1 or v2",
```

tests/e2e/api/collection.test.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Aptos, AptosConfig, Network } from "../../../src";
+
+describe("Collection", () => {
+  test("it should get collection data", async () => {
+    const config = new AptosConfig({ network: Network.MAINNET });
```
0xmaayan, line 8: that's a great test, I think we can skip test it for now until we have a local support for creating collection and then test against localnet

0xmaayan (approved):

PR #40, opened 2023-10-13
[docs] Add docs generation to the npm scripts

gregnazario:
### Description
Now we can generate documentation, to prevent churn at this moment, I'm still hiding docs via .gitignore so they aren't added.  We will be able to use GitHub pages directly to publish the TS Docs

### Test Plan
`pnpm doc`

```
ls docs/aptos-2.0.0 
assets       classes      enums        functions    index.html   interfaces   modules.html types        variables
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
     "test": "pnpm jest",
     "unit-test": "pnpm jest tests/unit",
     "e2e-test": "pnpm jest tests/e2e",
-    "indexer-codegen": "graphql-codegen --config ./src/types/codegen.yaml"
+    "indexer-codegen": "graphql-codegen --config ./src/types/codegen.yaml",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -47,6 +48,7 @@
     "@types/node": "18.6.2",
     "@typescript-eslint/eslint-plugin": "5.36.2",
     "@typescript-eslint/parser": "5.36.2",
+    "cross-var": "^1.1.0",
```
0xmaayan, line 51: what's cross-var?

0xmaayan (approved):

package.json
gregnazario: Just a minor dependency that will handle letting the docs build on Windows as well as Linux

PR #41, opened 2023-10-13
Run E2E tests on high-perf-docker

banool:
### Description
The GH provided runners have a bunch of stuff running on them that the local testnet collides with. This doesn't happen with our own runners.

### Test Plan
I reran the workflow 5 times and it worked every time.

### DIFF
--- a/.github/actions/run-e2e-tests/action.yaml
+++ b/.github/actions/run-e2e-tests/action.yaml
@@ -26,22 +26,6 @@ runs:
     - run: pnpm install -g @aptos-labs/aptos-cli
       shell: bash
 
-    # Kill anything running on ports we need. This is what GH recommends, see more:
-    # https://github.com/actions/runner-images/issues/2820#issuecomment-790367908
-    # Eventually we'll make it possible for the CLI to do this itself.
-    - run: |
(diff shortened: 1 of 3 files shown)


gregnazario (approved):

.github/actions/run-e2e-tests/action.yaml
```diff
@@ -26,22 +26,6 @@ runs:
     - run: pnpm install -g @aptos-labs/aptos-cli
       shell: bash
 
-    # Kill anything running on ports we need. This is what GH recommends, see more:
-    # https://github.com/actions/runner-images/issues/2820#issuecomment-790367908
-    # Eventually we'll make it possible for the CLI to do this itself.
-    - run: |
-        declare -a PORTS=("8080" "8081" "8070" "8090" "43234" "43235" "43236" "43237" "43238" "43239" "43240" "43241" "43242" "43243" "43244" "43245" "43246" "43247" "43248" "43249" "50051" "50052")
```
0xjinn, line 33: I see, so some of these port numbers collides with the port # that are being used by gh runners?

0xjinn (approved):

.github/actions/run-e2e-tests/action.yaml
banool: Yeah I think some services just randomly pick ports that we try to use.

PR #42, opened 2023-10-13
[CI] Avoid exhausting file descriptors, disable secp256 tests

banool:
### Description
This PR does two things:
1. @0xmaayan observed that sometimes running the full test suite would result in the machine running out of fds. We help avoid this by limiting concurrency in jest. I timed it, on my machine this doesn't seem to affect how long the tests take overall.
2. The user txn processor currently breaks when handling txns with secp256. This is the case across all networks and in main, so it's just totally broken upstream. This PR skips those tests for now to make CI green.

### Test Plan
See CI run.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -17,4 +17,6 @@ module.exports = {
       statements: 50, // 95,
     },
   },
+  // To help avoid exhausting all the available fds.
+  maxWorkers: 4,
 };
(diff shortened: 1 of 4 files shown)


gregnazario (changes requested):
Not letting you land it :)

0xmaayan:
> 2. The user txn processor currently breaks when handling txns with secp256. This is the case across all networks and in main, so it's just totally broken upstream. This PR skips those tests for now to make CI green.


thats weird.... I was able to run secp256k1 tests separately

banool:
Idk Larry says it's meant to be broken on both? I'll look into it, maybe he landed a fix.

gregnazario (approved):

0xmaayan (approved):

PR #45, opened 2023-10-13
Remove generated from prettierignore

heliuchuan:
### Description
This has fmt run on the generated graphql code.  This PR also chains fmt to run in the codegen script.

### Test Plan
pnpm run indexer-codegen


gregnazario (approved):
May need to run `pnpm fmt` with this PR

PR #46, opened 2023-10-13
[lint] Fixes a bad code pattern

gregnazario:
### Description
Fixes a bad code pattern that was previously used to avoid a lint.

```
async outer(): Promise<> {
  const ret = await inner();
  return ret;
}
```

is now the below because the inner one doesn't have to be awaited (just passed back out)
```
async outer(): Promise<> {
  return inner();
}
```

### Test Plan

CI

### DIFF
--- a/src/api/account.ts
+++ b/src/api/account.ts
@@ -4,21 +4,21 @@
 import { AptosConfig } from "./aptos_config";
 import {
   AccountData,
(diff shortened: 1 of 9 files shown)


xbtmatt (approved):
Nice, makes everything a little more readable

PR #48, opened 2023-10-14
Add indexer version sync

0xjinn:
### Description

Update waitForTransaction to include and wait for indexer sync

1. Add `getProcessorStatus` to get the latest success version from indexer
2. Add method `getIndexerLastSuccessVersion` to fetch the last success version from indexer
3. Update waitForTransaction to include this check (Developer can opt-out if preferred)

### Test Plan

pnpm fmt
pnpm test

### DIFF
--- a/src/api/general.ts
+++ b/src/api/general.ts
@@ -9,6 +9,7 @@ import {
   getTableItem,
   queryIndexer,
   view,
+  getIndexerLastSuccessVersion,
 } from "../internal/general";
 import {
   Block,
@@ -170,5 +171,9 @@ export class General {
       aptosConfig: this.config,
       ...args,
     });
+  }
+
+  async getIndexerLastSuccessVersion(): Promise<number> {
+    return getIndexerLastSuccessVersion({ aptosConfig: this.config });
   }
 }
--- a/src/internal/general.ts
+++ b/src/internal/general.ts
@@ -13,6 +13,7 @@ import { getAptosFullNode, postAptosFullNode, postAptosIndexer } from "../client
 import {
   Block,
   GetChainTopUserTransactionsResponse,
+  GetProcessorStatusResponse,
   GraphqlQuery,
   LedgerInfo,
   LedgerVersion,
@@ -20,8 +21,8 @@ import {
   TableItemRequest,
   ViewRequest,
 } from "../types";
-import { GetChainTopUserTransactionsQuery } from "../types/generated/operations";
-import { GetChainTopUserTransactions } from "../types/generated/queries";
+import { GetChainTopUserTransactionsQuery, GetProcessorStatusQuery } from "../types/generated/operations";
+import { GetChainTopUserTransactions, GetProcessorStatus } from "../types/generated/queries";
 
 export async function getLedgerInfo(args: { aptosConfig: AptosConfig }): Promise<LedgerInfo> {
   const { aptosConfig } = args;
@@ -129,4 +130,25 @@ export async function queryIndexer<T>(args: {
     overrides: { WITH_CREDENTIALS: false },
   });
   return data;
+}
+
+export async function getProcessorStatuses(args: { aptosConfig: AptosConfig }): Promise<GetProcessorStatusResponse> {
+  const { aptosConfig } = args;
+
+  const graphqlQuery = {
+    query: GetProcessorStatus,
+  };
+
+  const data = await queryIndexer<GetProcessorStatusQuery>({
+    aptosConfig,
+    query: graphqlQuery,
+    originMethod: "getProcessorStatuses",
+  });
+
+  return data.processor_status;
+}
+
+export async function getIndexerLastSuccessVersion(args: { aptosConfig: AptosConfig }): Promise<number> {
+  const response = await getProcessorStatuses({ aptosConfig: args.aptosConfig });
+  return response[0].last_success_version;
 }
--- a/src/internal/transaction.ts
+++ b/src/internal/transaction.ts
@@ -21,6 +21,7 @@ import {
 import { DEFAULT_TXN_TIMEOUT_SEC } from "../utils/const";
 import { sleep } from "../utils/helpers";
 import { memoizeAsync } from "../utils/memoize";
+import { getIndexerLastSuccessVersion } from "./general";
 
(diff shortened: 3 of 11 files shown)


0xjinn:
* main
  * **PR #48** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-ts-sdk/48" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-ts-sdk/48?utm_source=stack-comment).

src/internal/transaction.ts
```diff
@@ -167,9 +169,48 @@ export async function waitForTransaction(args: {
...
+    // check for timeout
+    if (new Date().getTime() - startTime > timeoutMiliseconds) {
+      throw new Error("waitForLastSuccessIndexerVersionSync timeout");
+    }
+
+    // eslint-disable-next-line no-await-in-loop
+    indexerVersion = await getIndexerLastSuccessVersion({ aptosConfig });
+
+    // eslint-disable-next-line no-await-in-loop
+    await sleep(500);
```
gregnazario, line 210: We will likely want to do an exponential backoff in the future (we should later work on a helper function for it).  Otherwise, you can do no better than 500ms or 1 second here if it doesn't occur immediately.

let's change to 200ms for now.

gregnazario (approved):

src/internal/transaction.ts
```diff
@@ -167,9 +169,48 @@ export async function waitForTransaction(args: {
...
     );
   }
+
+  // Make sure indexer is synced with the latest ledger version
+  if (indexerVersionCheck) {
+    try {
+      await waitForLastSuccessIndexerVersionSync({ aptosConfig, ledgerVersion: Number(lastTxn.version) });
+    } catch (_e) {
+      throw new WaitForTransactionError(
+        `Transaction ${txnHash} timed out waiting for indexer to sync with ledger version ${lastTxn.version}`,
```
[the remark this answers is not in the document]
0xjinn: Ah good call.

banool:
I think it should be possible to optionally provide a list of processors you wait for and just wait for the min of those to catch up, rather than all of them implicitly.

src/internal/general.ts
```diff
@@ -130,3 +130,19 @@ export async function queryIndexer<T>(args: {
   });
   return data;
 }
+
+export async function getIndexerLastSuccessVersion(args: { aptosConfig: AptosConfig }): Promise<number> {
```
0xmaayan, line 134: looks like the query accepts `where` condition but we dont support it in the api/internal functions

src/internal/general.ts
```diff
@@ -130,3 +130,19 @@ export async function queryIndexer<T>(args: {
...
+    query: GetProcessorStatus,
+  };
+
+  const data = await queryIndexer<GetProcessorStatusQuery>({
+    aptosConfig,
+    query: graphqlQuery,
+    originMethod: "getIndexerLastSuccessVersion",
+  });
+
+  return data.processor_status[0].last_success_version;
```
0xmaayan, line 147: this is the design we follow please make sure you follow it also
https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/types/indexer.ts#L45
https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/internal/account.ts#L456

0xjinn: Yep will update to following:

internal/general.ts
Update the use Promise\<GetProcessorStatusLastSuccessResponse\> from `export type GetProcessorStatusLastSuccessResponse = GetProcessorStatusQuery["processor_status"][0];`

api/general.ts
Keep the Promise<number> - Because this is suppose the return the latest version number

0xjinn: Actually never mind. Will create 2 different reponse `GetProcessorStatusResponse` and `GetIndexerLastSuccessVersionResponse`

src/internal/general.ts
0xjinn: Will remove.

I was thinking about having it there to allow flexibility in the future, but will remove where clause for now

src/internal/transaction.ts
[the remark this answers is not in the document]
0xjinn: Updated with new error message

0xjinn:
Update:

I removed the e2e test right now, because it does not work. There is no way to mock the func the way it is. See issue #56 for details

src/internal/transaction.ts
0xjinn: Changed to 200ms

0xjinn:
> I think it should be possible to optionally provide a list of processors you wait for and just wait for the min of those to catch up, rather than all of them implicitly.

Currently I get the list order by `asc` based on the last_success. So its always at index 0.

src/api/general.ts
```diff
@@ -171,4 +172,9 @@ export class General {
       ...args,
     });
   }
+
+  async getIndexerLastSuccessVersion(): Promise<number> {
+    const response = await getIndexerLastSuccessVersion({ aptosConfig: this.config });
+    return response;
```
heliuchuan, line 178:
```suggestion
  async getIndexerLastSuccessVersion(): Promise<number> {
    return getIndexerLastSuccessVersion({ aptosConfig: this.config });
```
can just return

src/internal/general.ts
```diff
@@ -130,3 +131,24 @@ export async function queryIndexer<T>(args: {
...
+  const { aptosConfig } = args;
+
+  const graphqlQuery = {
+    query: GetProcessorStatus,
+  };
+
+  const data = await queryIndexer<GetProcessorStatusQuery>({
+    aptosConfig,
+    query: graphqlQuery,
+    originMethod: "getIndexerLastSuccessVersion",
```
heliuchuan, line 145: this should be getProcessorStatuses i think

0xjinn: Good catch.

gregnazario (approved):

heliuchuan (approved):

src/api/general.ts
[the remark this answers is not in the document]
0xjinn: This is probably commented on an outdated source code.

PR #49, opened 2023-10-14
[hdkey] Consolidate and add testing for seed phrase derivation

gregnazario:
### Description
Cleans up and adds proper testing for seed phrase derivation.  This also opens up the path for
derivation of other keys.

Additionally, this removes `isValidPath` from the Account class, as it doesn't really belong there, but instead it is separate on it's own.

### Test Plan
Added testing for valid and invalid seed phrases, ensured that derivation changes based on the path, and that the key derivation is correct.

### DIFF
--- a/src/utils/hdKey.ts
+++ b/src/utils/hdKey.ts
@@ -1,79 +1,104 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-import nacl from "tweetnacl";
 import { hmac } from "@noble/hashes/hmac";
 import { sha512 } from "@noble/hashes/sha512";
-import { hexToBytes } from "@noble/hashes/utils";
+import * as bip39 from "@scure/bip39";
 
-export type Keys = {
+export type DerivedKeys = {
   key: Uint8Array;
   chainCode: Uint8Array;
 };
 
-const pathRegex = /^m(\/[0-9]+')+$/;
+/**
+ * Aptos derive path is 637
+ *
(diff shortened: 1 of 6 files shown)


src/utils/hdKey.ts
```diff
@@ -1,79 +1,104 @@
...
+ * Aptos derive path is 637
+ *
+ * See https://github.com/satoshilabs/slips/blob/master/slip-0044.md
+ */
+export const APTOS_PATH_REGEX = /^m\/44'\/637'\/[0-9]+'\/[0-9]+'\/[0-9]+'?$/;
 
-const replaceDerive = (val: string): string => val.replace("'", "");
-
-const HMAC_KEY = "ed25519 seed";
+export const ED25519_KEY = "ed25519 seed";
```
0xmaayan, line 20: do we also need support for Secp256k1

0xmaayan (approved):
thoughts on renaming hdKey file? have no suggestion but hdKey doesnt really mean anything to me lol

gregnazario:
> thoughts on renaming hdKey file? have no suggestion but hdKey doesnt really mean anything to me lol

Yeah, I also was very confused until I looked it up, it's hierarchical derivation keys.  Though, its painfully long of a name, so I think we'll keep the file name to hd.

src/utils/hdKey.ts
gregnazario: We may but, I'll have to do more research into what the output value needs to be from it.

PR #50, opened 2023-10-14
Improve events queries

0xmaayan:
### Description
Full event queries support (requested by external partner).

We only had one event query, and with no orderBy or pagination support.

Added `getEvents` to fetch all events and support filtering, pagination and orderBy.
In addition, added support for `getAccountEventsByCreationNumber` and `getAccountEventsByEventType` queries.

All queries support v1 and v2 events.

### Test Plan
`pnpm jest`
We have reached the `longTestTimeout` spot. Will work on fixing it later.


src/internal/event.ts
```diff
@@ -10,41 +10,90 @@
...
- * @param args.address - The account address
- * @param args.creationNumber - The event creation number
- * @returns Promise<GetEventsByCreationNumberResponse>
- */
-export async function getEventsByCreationNumber(args: {
+export async function getAccountEventsByCreationNumber(args: {
   aptosConfig: AptosConfig;
   address: HexInput;
   creationNumber: AnyNumber;
+  options?: {
```
heliuchuan, line 23: does it make sense to paginate this?  

"Event types are globally identifiable by an account address and monotonically increasing creation_number, one per event type emitted to the given account. This API returns events corresponding to that that event type."

0xmaayan: check out this [query](https://cloud.hasura.io/public/graphiql?endpoint=https%3A%2F%2Findexer-devnet.staging.gcp.aptosdev.com%2Fv1%2Fgraphql&query=query+MyQuery+%7B%0A++events%28where%3A+%7Bcreation_number%3A+%7B_eq%3A+%223%22%7D%7D%29+%7B%0A++++type%0A++++transaction_version%0A++++transaction_block_height%0A++++sequence_number%0A++++event_index%0A++++data%0A++++creation_number%0A++++account_address%0A++%7D%0A%7D%0A)

0xmaayan: sorry this query was without account address - I guess you are right

0xmaayan: done

src/api/event.ts
```diff
@@ -16,13 +17,66 @@ export class Event {
...
+   *
+   * @returns GetEventsQuery response type
    */
-  async getEventsByCreationNumber(args: { address: string; creationNumber: AnyNumber }): Promise<GetEventsResponse> {
-    return getEventsByCreationNumber({ aptosConfig: this.config, ...args });
+  async getEvents(args?: {
+    options?: {
+      where?: EventsBoolExp;
+      pagination?: IndexerPaginationArgs;
+      orderBy?: OrderBy<GetEventsResponse[0]>;
```
gregnazario, line 77: I'm curious what the [0] is in this context since I'm learning these random things in typescript

src/internal/event.ts
```diff
@@ -10,41 +10,80 @@
...
+
+  const customOptions = {
+    where: whereCondition,
+    pagination: options?.pagination,
+    orderBy: options?.orderBy,
+  };
+
+  const data = await getEvents({ aptosConfig, options: customOptions });
+
+  return data;
```
gregnazario, line 62: Can drop the await and return, and just return the future directly

```suggestion
  return getEvents({ aptosConfig, options: customOptions });
```

tests/e2e/api/event.test.ts
```diff
@@ -6,22 +6,77 @@ import { FUND_AMOUNT, INDEXER_WAIT_TIME } from "../../unit/helper";
...
+      amount: FUND_AMOUNT,
+    });
+
+    await sleep(INDEXER_WAIT_TIME);
+    const events = await aptos.getEvents({
+      options: { where: { account_address: { _eq: testAccount1.accountAddress.toString() } } },
+    });
+
+    expect(events[0].account_address).toBe(testAccount1.accountAddress.toString());
+  }, 8000);
```
gregnazario, line 81: Nit: let's make timeouts constants so they read better

gregnazario (approved):

src/api/event.ts
heliuchuan: So GetEventsResponse is technically array - e.g. {prop1: number, prop2: string ....}[]

We want the type {prop1: number, prop2: string ....} .  The [0] allows us to reference the underlying type.

I asked the same thing to @0xmaayan recently lol.

0xmaayan: yes, this is a design choice. 
The response is an array of items. We want to order by every item in the array and therefore we use the first element keys to know what we can order by.

heliuchuan (approved):

PR #52, opened 2023-10-14
[api] Fix view functions to return multiple values

gregnazario:
### Description
If a view function returns a tuple, it will return multiple values. This changes the assumption that there is only one return value.

### Test Plan
Adjusted the existing test, but need a new contract to show a view function that has more than one return value

### DIFF
--- a/src/api/general.ts
+++ b/src/api/general.ts
@@ -124,11 +124,11 @@ export class General {
    * };
    * `
    *
-   * @returns a Move value
+   * @returns an array of Move values
    */
-  async view(args: { payload: ViewRequest; options?: LedgerVersion }): Promise<MoveValue> {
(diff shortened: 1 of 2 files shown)


heliuchuan (approved):

src/api/general.ts
```diff
@@ -124,11 +124,11 @@ export class General {
...
    * `
    *
-   * @returns a Move value
+   * @returns an array of Move values
    */
-  async view(args: { payload: ViewRequest; options?: LedgerVersion }): Promise<MoveValue> {
+  async view(args: { payload: ViewRequest; options?: LedgerVersion }): Promise<Array<MoveValue>> {
     const data = await view({ aptosConfig: this.config, ...args });
-    return data[0];
+    return data;
```
0xmaayan, line 131: thanks! I had this change in mind at some point but apparently forgot about it

0xmaayan (approved):

PR #53, opened 2023-10-14
Add a simple fee payer transaction example

0xmaayan:
### Description
* Add simple fee payer transaction example
* Modify existing simple transfer example

Might be better to encourage to use `getAccountCoinsData` query to check account balance instead of `getAccountResource`. Even better to have a query to check account balance that return the amount prop returned from `getAccountResource`.

Anyway, we should wait for https://github.com/aptos-labs/aptos-ts-sdk/pull/48#pullrequestreview-1677681796 so we don't need to use sleep.

### Test Plan
run all tests


examples/javascript/simple_sponsored_transaction.js
```diff
@@ -0,0 +1,97 @@
...
+
+  // Bob should have the transfer amount
+  if (bobBalanceAfter[0].amount !== TRANSFER_AMOUNT) throw new Error("Bob's balance after transfer is incorrect");
+
+  // Alice should have the initial balance minus tranfer amount
+  if (aliceBalanceAfter[0].amount !== ALICE_INITIAL_BALANCE - TRANSFER_AMOUNT)
+    throw new Error("Alice's balance after transfer is incorrect");
+
+  // Sponser should have the initial balance minus gas
+  if (sponserBalanceAfter[0].amount >= SPONSER_INITIAL_BALANCE)
```
xbtmatt, line 91: if we want to be *really* clear, you can also check the actual gas used in the tx and throw an error if it doesn't match

xbtmatt:
need to `ctrl + f` replace `sponser` with `sponsor`:D

examples/typescript/simple_sponsored_transaction.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  // Submit the transaction to chain
+  const committedTxn = await aptos.submitTransaction({
+    transaction,
+    senderAuthenticator: senderSignature,
+    secondarySignerAuthenticators: { feePayerAuthenticator: sponserSignature },
+  });
+
+  console.log(`Submitted transaction: ${committedTxn.hash}`);
+  await aptos.waitForTransaction({ txnHash: committedTxn.hash });
```
gregnazario, line 76: I think after Jin's change we can skip the sleep

examples/typescript/simple_sponsored_transaction.ts
```diff
@@ -0,0 +1,104 @@
...
+  console.log(`Bob's address is: ${bobAddress}`);
+  console.log(`Sponser's address is: ${sponserAddress}`);
+
+  // Fund Alice and Sponser accounts
+  await aptos.fundAccount({ accountAddress: alice.accountAddress.toString(), amount: ALICE_INITIAL_BALANCE });
+  await aptos.fundAccount({ accountAddress: sponser.accountAddress.toString(), amount: SPONSER_INITIAL_BALANCE });
+
+  // Show account balances
+  const aliceBalanceBefore = await aptos.getAccountCoinsData({ accountAddress: aliceAddres });
+  const sponserBalanceBefore = await aptos.getAccountCoinsData({ accountAddress: sponserAddress });
```
gregnazario, line 41:
```suggestion
  const sponsorBalanceBefore = await aptos.getAccountCoinsData({ accountAddress: sponserAddress });
```

examples/typescript/simple_sponsored_transaction.ts
```diff
@@ -0,0 +1,104 @@
...
+    "This example will create three accounts (Alice, Bob and Sponser), fund Alice and Sponser, transfer between Alice and Bob with Sponser to pay the gas fee.",
+  );
+
+  // Setup the client
+  const aptos = new Aptos();
+
+  // Create three accounts
+  const alice = Account.generate();
+  const bob = Account.generate();
+  const sponser = Account.generate();
```
gregnazario, line 23:
```suggestion
  const sponsor = Account.generate();
```

examples/typescript/simple_sponsored_transaction.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  // Variables to hold Alice and Sponser accounts address
+  const aliceAddres = alice.accountAddress.toString();
+  const bobAddress = bob.accountAddress.toString();
+  const sponserAddress = sponser.accountAddress.toString();
```
gregnazario, line 28:
```suggestion
  const sponsorAddress = sponsor.accountAddress.toString();
```

examples/typescript/simple_sponsored_transaction.ts
```diff
@@ -0,0 +1,104 @@
...
+  const aptos = new Aptos();
+
+  // Create three accounts
+  const alice = Account.generate();
+  const bob = Account.generate();
+  const sponser = Account.generate();
+
+  // Variables to hold Alice and Sponser accounts address
+  const aliceAddres = alice.accountAddress.toString();
+  const bobAddress = bob.accountAddress.toString();
```
gregnazario, line 27:
```suggestion
  const aliceAddress = alice.accountAddress.toString();
  const bobAddress = bob.accountAddress.toString();
```

examples/typescript/simple_sponsored_transaction.ts
```diff
@@ -0,0 +1,104 @@
...
+  const sponserAddress = sponser.accountAddress.toString();
+
+  console.log("=== Addresses ===\n");
+  console.log(`Alice's address is: ${aliceAddres}`);
+  console.log(`Bob's address is: ${bobAddress}`);
+  console.log(`Sponser's address is: ${sponserAddress}`);
+
+  // Fund Alice and Sponser accounts
```
gregnazario, line 35:
```suggestion
  // Fund Alice and Sponsor accounts
```

examples/typescript/simple_sponsored_transaction.ts
```diff
@@ -0,0 +1,104 @@
+/**
+ * Example to submit a simple sponsored transaction where Alice transfers APT coin to Bob
+ * with a sponser account to pay for the gas fee
+ */
+import { Account, Aptos, U64 } from "aptos";
+
+const ALICE_INITIAL_BALANCE = 100_000_000;
+const SPONSER_INITIAL_BALANCE = 100_000_000;
```
gregnazario, line 8:
```suggestion
const SPONSOR_INITIAL_BALANCE = 100_000_000;
```

examples/typescript/simple_sponsored_transaction.ts
```diff
@@ -0,0 +1,104 @@
...
+  // Show account balances
+  const aliceBalanceBefore = await aptos.getAccountCoinsData({ accountAddress: aliceAddres });
+  const sponserBalanceBefore = await aptos.getAccountCoinsData({ accountAddress: sponserAddress });
+
+  console.log("\n=== Balances ===\n");
+  console.log(`Alice's balance is: ${aliceBalanceBefore[0].amount}`);
+  console.log(`Bobs's balance is: ${BOB_INITIAL_BALANCE}`);
+  console.log(`Sponser's balance is: ${sponserBalanceBefore[0].amount}`);
+
+  if (aliceBalanceBefore[0].amount !== ALICE_INITIAL_BALANCE) throw new Error("Alice's balance is incorrect");
```
gregnazario, line 48: Fyi if there's an assert library or something you think is better, you can add it.  I just wanted to keep dependencies down / keep it simple

examples/typescript/simple_sponsored_transaction.ts
```diff
@@ -0,0 +1,104 @@
...
+  // Alice and Sponser to sign the transaction (order matters)
+  const senderSignature = aptos.signTransaction({ signer: alice, transaction });
+  const sponserSignature = aptos.signTransaction({ signer: sponser, transaction });
+
+  // Submit the transaction to chain
+  const committedTxn = await aptos.submitTransaction({
+    transaction,
+    senderAuthenticator: senderSignature,
+    secondarySignerAuthenticators: { feePayerAuthenticator: sponserSignature },
+  });
```
gregnazario, line 73: One thing in the future for fee payer, we should separate scopes and pretend complete separate spaces for the two users to be a little more explicit about who end up sending it in the end.

Since the UX has the sponsor sign second, at the moment the sponsor would probably actually send the transaction.

examples/typescript/simple_sponsored_transaction.ts
```diff
@@ -0,0 +1,104 @@
...
+  // Alice should have the initial balance minus tranfer amount
+  if (aliceBalanceAfter[0].amount !== ALICE_INITIAL_BALANCE - TRANSFER_AMOUNT)
+    throw new Error("Alice's balance after transfer is incorrect");
+
+  // Sponser should have the initial balance minus gas
+  if (sponserBalanceAfter[0].amount >= SPONSER_INITIAL_BALANCE)
+    throw new Error("Sponser's balance after transfer is incorrect");
+
+  console.log(`Alice's final balance is: ${ALICE_INITIAL_BALANCE - TRANSFER_AMOUNT}`);
+  console.log(`Bob's balance is: ${TRANSFER_AMOUNT}`);
```
gregnazario, line 95: Just for typo reasons I'd use the actual value rather than the value you expect.  If someone changes the if statement this value will be wrong

examples/typescript-esm/simple_sponsored_transaction.ts
gregnazario, line 1: Unless there's something majorly different, other than just showing it can be done, we may not need all the examples for ESM

gregnazario (approved):
Looks good, I just commented on the typescript one.

We may think about just writing the examples mostly for one, unless we expect there to be anything major we'll learn.

Additionally *sponsor* everywhere :)

examples/javascript/simple_sponsored_transaction.js
0xmaayan: interestingly, 
Sponsor initial balance = 100_000_000
Transaction gas used = 1010
Expected sponsor final balance = 99998990
Sponsor final balance fron Indexer = 99899000
Sponsor final balance from fullnode = 99899000

Not sure where the extra 10 is from..?

examples/typescript/simple_sponsored_transaction.ts
0xmaayan: I actually find throwing is nicer and enough, expect/assert doesnt give nice info

examples/typescript/simple_sponsored_transaction.ts
0xmaayan: yes, we should mock 2 servers and have the transaction sent between to sign and submit

examples/typescript/simple_sponsored_transaction.ts
0xmaayan: done

examples/typescript-esm/simple_sponsored_transaction.ts
0xmaayan: agreed, done

0xjinn:
Probably want to update the README file in /example as well @0xmaayan

PR #55, opened 2023-10-14
Add getTokenData query

heliuchuan:
### Description
Adds the getTokenData query

### Test Plan
There is a commented out test against MAINNET.  Once we have an example/ability to create a token locally via sdk we will enable the test for LOCAL.


tests/e2e/api/token.test.ts
```diff
@@ -0,0 +1,20 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+// import { AptosConfig, Aptos } from "../../../src";
+// import { Network } from "../../../src/utils/apiEndpoints";
+
+describe("token api", () => {
+  test("it gets token data for a token's address", async () => {
+    // const config = new AptosConfig({ network: Network.MAINNET });
+    // const aptos = new Aptos(config);
+    // const tokenAddress = "0x0cb098d8f875f38dcb4109e2638e3e24055a585ed2143e9ba76e002fea303795";
```
gregnazario, line 11: We should probably do a mint and lookup with the local testnet

gregnazario (commented):
Do we have minting already in / can we do that flow?  Otherwise we should update it when we add it

src/internal/token.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { AptosConfig } from "../api/aptos_config";
+import { GetTokenDataResponse, HexInput } from "../types";
+import { GetTokenDataQuery } from "../types/generated/operations";
+import { GetTokenData } from "../types/generated/queries";
+import { queryIndexer } from "./general";
+
+export async function getTokenData(args: {
+  aptosConfig: AptosConfig;
+  tokenAddress: HexInput;
```
0xmaayan, line 19: lets add Hex validation
```
Hex.fromHexInput(creatorAddress).toString();
```

0xmaayan (approved):

PR #59, opened 2023-10-14
Add transfer_coin example using transferCoinTransaction

0xjinn:
### Description

Similar to the transfer example, but instead using transferCoinTransaction

### Test Plan

pnpm transfer_coin

### DIFF
--- a/examples/typescript/transfer_coin.ts
+++ b/examples/typescript/transfer_coin.ts
@@ -27,7 +27,9 @@ const balance = async (aptos: Aptos, name: string, address: AccountAddress) => {
 };
 
 const example = async () => {
-  console.log("This example will create two accounts (Alice and Bob), fund them, and transfer between them using transferCoinTransaction.");
+  console.log(
+    "This example will create two accounts (Alice and Bob), fund them, and transfer between them using transferCoinTransaction.",
+  );
 
   // Setup the client
   const config = new AptosConfig({ network: Network.DEVNET });
--- a/examples/typescript/transfer_coin.ts
+++ b/examples/typescript/transfer_coin.ts
@@ -1,0 +1,93 @@
+/**
+ * This example shows how to use the Aptos client to create accounts, fund them, and transfer between them.
+ * Similar to ./simple_transfer.ts, but uses transferCoinTransaction to generate the transaction.
+ */
+
+import { Account, AccountAddress, Aptos, AptosConfig, Network } from "aptos";
+
+const COIN_STORE = `0x1::coin::CoinStore<0x1::aptos_coin::AptosCoin>`;
+const ALICE_INITIAL_BALANCE = 100_000_000;
+const BOB_INITIAL_BALANCE = 1_000_000;
+const TRANSFER_AMOUNT = 1_000_000;
+
+/**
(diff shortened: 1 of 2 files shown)


examples/typescript/transfer_coin.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Prints the balance of an account
+ * @param aptos
+ * @param name
+ * @param address
+ * @returns {Promise<*>}
+ *
+ */
+const balance = async (aptos: Aptos, name: string, address: AccountAddress) => {
+  let resource = await aptos.getAccountResource({ accountAddress: address.toUint8Array(), resourceType: COIN_STORE });
```
0xmaayan, line 22: I start to think we want to use `getAccountCoinsData` query instead of this.. imo is cleaner

examples/typescript/transfer_coin.ts
```diff
@@ -0,0 +1,95 @@
...
+  return amount;
+};
+
+const example = async () => {
+  console.log(
+    "This example will create two accounts (Alice and Bob), fund them, and transfer between them using transferCoinTransaction.",
+  );
+
+  // Setup the client
+  const config = new AptosConfig({ network: Network.DEVNET });
```
0xmaayan, line 35: devnet is default for now - might want to change it
```suggestion
  const config = new AptosConfig();
```

examples/typescript/transfer_coin.ts
```diff
@@ -0,0 +1,95 @@
...
+  console.log("=== Addresses ===\n");
+  console.log(`Alice's address is: ${alice.accountAddress.toString()}`);
+  console.log(`Bob's address is: ${bob.accountAddress.toString()}`);
+
+  // Fund the accounts
+  console.log("\n=== Funding accounts ===\n");
+
+  // Fund alice account
+  const aliceFundTxn = await aptos.faucet.fundAccount({
+    accountAddress: alice.accountAddress.toUint8Array(),
```
0xmaayan, line 51: any reason in using `toUint8Array()` instead of `toString()`?
```suggestion
    accountAddress: alice.accountAddress.toUint8Array(),
```

examples/typescript/transfer_coin.ts
```diff
@@ -0,0 +1,95 @@
...
+    amount: ALICE_INITIAL_BALANCE,
+  });
+  console.log("Alice's fund transaction: ", aliceFundTxn);
+
+  // Fund bob account
+  const bobFundTxn = await aptos.faucet.fundAccount({
+    accountAddress: bob.accountAddress.toUint8Array(),
```
0xmaayan, line 58: same

0xmaayan (approved):

examples/typescript/transfer_coin.ts
0xjinn: hmm I am actually fine with both haha. Can update if needed :)

examples/typescript/transfer_coin.ts
0xjinn: I think either is fine. Though I will prefer the actual `AccountAddress` object

PR #61, opened 2023-10-15
Add transaction simulation test

0xmaayan:
### Description
Add tests for transaction simulation for both ED25519 and Secp256k1
* single signer transaction with script / entry function / multisig payloads
* multi agent transaction with script / entry function payloads
* fee payer transaction with script / entry function / multisig payloads and as multi agent

### Test Plan
pnpm test


gregnazario (approved):

xbtmatt (approved):

src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -328,17 +317,34 @@ export function generateSignedTransactionForSimulation(args: SimulateTransaction
...
+      accountAuthenticator.public_key,
+      accountAuthenticator.signature,
+    );
+  } else if (accountAuthenticator instanceof AccountAuthenticatorSecp256k1) {
+    transactionAuthenticator = new TransactionAuthenticatorSecp256k1(
+      accountAuthenticator.public_key,
+      accountAuthenticator.signature,
+    );
+  } else {
+    throw new Error("Invalud public key");
```
xbtmatt, line 332:
```suggestion
    throw new Error("Invalid public key");
```

tests/e2e/transaction/transaction_simulation.test.ts
```diff
@@ -0,0 +1,413 @@
...
+          transaction: rawTxn,
+        });
+        expect(response.success).toBeTruthy();
+      });
+      test("with multisig payload", async () => {
+        const rawTxn = await aptos.generateTransaction({
+          sender: senderAccount.accountAddress.toString(),
+          data: {
+            multisigAddress: secondarySignerAccount.accountAddress,
+            function: `0x${senderAccount.accountAddress.toStringWithoutPrefix()}::transfer::transfer`,
```
gregnazario, line 75: We should use `toString()` instead of `0x${toStringWithoutPrefix()}`

0xmaayan: it is a typing thing, `function` expects `0x${string}::${string}::${string}`, and using `toString()` typescript complains type mismatch

gregnazario: Oh, that's super annoying... we should probably fix that.

PR #65, opened 2023-10-16
Add generate transaction api tests

0xmaayan:
### Description
Add generate transaction api function tests
* generate single signer transaction with script / entry function / multisig payloads
* generate multi agent transaction with script / entry function payloads
* generate fee payer transaction with script / entry function / multisig payloads

### Test Plan
pnpm test


tests/e2e/transaction/generate_transaction.test.ts
```diff
@@ -0,0 +1,197 @@
...
+import { longTestTimeout } from "../../unit/helper";
+import { fundAccounts, singleSignerScriptBytecode } from "./helper";
+
+describe("generate transaction", () => {
+  const config = new AptosConfig({ network: Network.LOCAL });
+  const aptos = new Aptos(config);
+  const senderAccount = Account.generate();
+  const recieverAccounts = [Account.generate(), Account.generate()];
+  const secondarySignerAccount = Account.generate();
+  const feePayerAccount = Account.generate();
```
0xjinn, line 17: I wonder, should we do some test against different scheme type of account?

For example:
senderAccount = Ed25519
recieverAccount = Secp256k1
...

0xmaayan: yes. we can add it on simulation/submission tests, generate transaction doesnt really care for the signing scheme

0xjinn (approved):

PR #66, opened 2023-10-16
Add transaction submission api test

0xmaayan:
### Description
Need https://github.com/aptos-labs/aptos-ts-sdk/pull/65 to land first.

Add transaction submission api function tests

secp256k1 is skipped for now because of a user txn processor issue https://github.com/aptos-labs/aptos-ts-sdk/pull/42

* submit single signer transaction with script / entry function / multisig payloads
* submit multi agent transaction with script / entry function payloads
* submit fee payer transaction with script / entry function / multisig payloads

### Test Plan
pnpm test

### DIFF
--- a/tests/e2e/transaction/transaction_submission.test.ts
+++ b/tests/e2e/transaction/transaction_submission.test.ts
@@ -10,7 +10,7 @@ describe("transaction submission", () => {
   const config = new AptosConfig({ network: Network.LOCAL });
   const aptos = new Aptos(config);
   const senderAccount = Account.generate();
-  const recieverAccounts = [Account.generate(), Account.generate()];
+  const receiverAccounts = [Account.generate(), Account.generate()];
   const senderSecp256k1Account = Account.generate(SigningScheme.Secp256k1Ecdsa);
   const secondarySignerAccount = Account.generate();
   const feePayerAccount = Account.generate();
@@ -18,7 +18,7 @@ describe("transaction submission", () => {
     await fundAccounts(aptos, [
       senderAccount,
       senderSecp256k1Account,
-      ...recieverAccounts,
+      ...receiverAccounts,
       secondarySignerAccount,
       feePayerAccount,
     ]);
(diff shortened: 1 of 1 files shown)


xbtmatt:
At some point we should probably try to use the same little "test harness" for these lol, there's a ton of setup

tests/e2e/transaction/transaction_submission.test.ts
```diff
@@ -3,100 +3,362 @@
...
+        const response = await aptos.submitTransaction({
+          transaction,
+          senderAuthenticator,
+          secondarySignerAuthenticators: { feePayerAuthenticator: feePayerSignerAuthenticator },
         });
-        const bobauthenticator = aptos.signTransaction({
-          signer: bob,
-          transaction: rawTxn,
+
+        await waitForTransaction({ aptosConfig: config, txnHash: response.hash });
```
xbtmatt, line 139: I think in general we should be doing a check here right?

The node api will tell you if you have a fee payer  signature, etc

```typescript
const responseSignature = response.signature as TransactionFeePayerSignature;
expect(responseSignature.type).toEqual("fee_payer_signature");
```

this is what I do in mine

the 5 types:

```
ed25519_signature
secp256k1_ecdsa_signature
multi_agent_signature
fee_payer_signature
multi_ed25519_signature
```

tests/e2e/transaction/transaction_submission.test.ts
```diff
@@ -3,100 +3,362 @@
...
+  const senderAccount = Account.generate();
+  const recieverAccounts = [Account.generate(), Account.generate()];
+  const senderSecp256k1Account = Account.generate(SigningScheme.Secp256k1Ecdsa);
+  const secondarySignerAccount = Account.generate();
+  const feePayerAccount = Account.generate();
+  beforeAll(async () => {
+    await fundAccounts(aptos, [
+      senderAccount,
+      senderSecp256k1Account,
+      ...recieverAccounts,
```
xbtmatt, line 21:
```suggestion
      ...receiverAccounts,
```

ctrl+f `reciever` nit :D

tests/e2e/transaction/transaction_submission.test.ts
0xmaayan: hmm I guess it depends on what you are testing. those tests tests that transaction is submitted. We have some other tests that tests the signature in `sign_transaction`

PR #71, opened 2023-10-16
Make type_arguments optional and publically rename to typeArguments

heliuchuan:
### Description
This makes it so that the the developer doesn't have to define type_arguments when building transactions.  This is the majority use case.  This also changes it to camelCase for consistency for public facing params.  Snake case is still required for serialization.

### Test Plan
pnpm test


0xmaayan (approved):
nice!

PR #80, opened 2023-10-17
[docs and code] Pass one on Standardization

gregnazario:
### Description
This attempts to standardize some input variables, comments, and other fields.  Note that it's almost 3 AM for me by the time I finished this, so multiple passes will need to be taken.

### Test Plan
I regenerated the docs, ran the lint and formatter.  My local node kept crashing, so I'm waiting on CI for any other issues


0xmaayan (approved):
blind stamp

PR #82, opened 2023-10-17
Remove unused pagination arg in getAccountEventsByCreationNumber

heliuchuan:
### Description
see title

### Test Plan
pnpm test

### DIFF
--- a/src/api/event.ts
+++ b/src/api/event.ts
@@ -27,10 +27,6 @@ export class Event {
   async getAccountEventsByCreationNumber(args: {
     address: string;
     creationNumber: AnyNumber;
-    options?: {
-      pagination?: PaginationArgs;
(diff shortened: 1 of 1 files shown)


src/api/event.ts
```diff
@@ -27,10 +27,6 @@ export class Event {
   async getAccountEventsByCreationNumber(args: {
     address: string;
     creationNumber: AnyNumber;
-    options?: {
-      pagination?: PaginationArgs;
-      orderBy?: OrderBy<GetEventsResponse[0]>;
-    };
```
0xjinn, line 33: I been thinking, instead of removing it, should we add pagination for all /api query func? So they are consistent. Any thought?

0xmaayan: not all queries need pagination. for example, this one.

0xmaayan (approved):

PR #84, opened 2023-10-17
Introduce Single Sender transaction authenticator

0xmaayan:
### Description
This PR adds support to the new Single Sender introduced in [AIP-55](https://github.com/aptos-foundation/AIPs/pull/263)
We mostly have things represented in a way that reflects the rust code.

We introduce 2 new modules
- `AnyPublicKey` that holds a reference to a `PublicKey`, we use it to represent any Single Sender key.
- `AnySignature` that holds a reference to a `Signature`, we use it to represent any Single Sender signature.

When generating new keys (i.e creating an Account) we default to generate the new keys as `SingleSigner` but also support an optional `legacy` argument to generate keys based on the legacy ED25519 scheme.

When want to get the account from an existing and known private key, i.e `await aptos.deriveAccountFromPrivateKey`, we derive the public and auth keys from the provided private key and then generating an Account based on the Account configured signing scheme - Legacy ED25519 or Single Sender. **Note that this function is now an async function since we need to make api calls to chain.**

The transaction builder resolves to the correct signing scheme. 

- When a single signer transaction is being simulated, it checks for the provided public key instance, if it is an instance of `AnyPublicKey` we create a `AccountAuthenticatorSingleKey` based on the `PublicKey` the class holds reference to.
- When a single signer transaction is being signed, it checks for the account signing scheme and, if it is `SingleKey` we create a `AccountAuthenticatorSingleKey` based on the `PublicKey` the account uses.
- When a single signer transaction is being submitted, it checks for the provided account authenticator, if it is an instance of a `SingleKeyAuthenticator` we create a `SingleSenderTransactionAuthenticator` to create a `SignedTransaction` with.

Also not, we remove `fromPrivateKeyAndAddress` static method as `fromPrivateKey` basically does it now

### Test Plan
Indexer doesnt support the new Single Sender yet, so tests are breaking. Was able to successfully run it locally without indexer
```
 PASS  tests/e2e/transaction/transactionSimulation.test.ts (8.032 s)
 PASS  tests/e2e/transaction/transactionArguments.test.ts (10.768 s)
 PASS  tests/e2e/transaction/generateTransaction.test.ts
 PASS  tests/e2e/transaction/signTransaction.test.ts (5.155 s)
 PASS  tests/e2e/transaction/transactionBuilder.test.ts (15.517 s)
 PASS  tests/e2e/transaction/transactionSubmission.test.ts (17.563 s)
```

### Related Links
[AIP-55](https://github.com/aptos-foundation/AIPs/pull/263)


src/core/crypto/ed25519.ts
```diff
@@ -76,6 +76,93 @@ export class Ed25519PublicKey extends PublicKey {
...
+  serialize(serializer: Serializer): void {
+    serializer.serializeBytes(this.key.toUint8Array());
+  }
+
+  static deserialize(deserializer: Deserializer): Ed25519PublicKey {
+    const bytes = deserializer.deserializeBytes();
+    return new Ed25519PublicKey(bytes);
+  }
+
+  serializeForSingleKey(serializer: Serializer): void {
```
0xmaayan, line 157: can ignore

src/core/crypto/ed25519.ts
```diff
@@ -221,4 +308,14 @@ export class Ed25519Signature extends Signature {
     const bytes = deserializer.deserializeBytes();
     return new Ed25519Signature(bytes);
   }
+
+  serializeForSingleKey(serializer: Serializer): void {
```
0xmaayan, line 312: can ignore

src/core/crypto/ed25519.ts
```diff
@@ -76,6 +76,93 @@ export class Ed25519PublicKey extends PublicKey {
     const bytes = deserializer.deserializeBytes();
     return new Ed25519PublicKey(bytes);
   }
+
+  serializeForSingleKey(serializer: Serializer): void {
```
0xmaayan, line 80: can ignore

src/core/crypto/ed25519.ts
```diff
@@ -76,6 +76,93 @@ export class Ed25519PublicKey extends PublicKey {
...
+    this.serialize(serializer);
+  }
+
+  static load(deserializer: Deserializer): Ed25519PublicKey {
+    const bytes = deserializer.deserializeBytes();
+    return new Ed25519PublicKey(bytes);
+  }
+}
+
+export class SingleKey extends PublicKey {
```
0xmaayan, line 91: will be in a separate file

src/core/crypto/secp256k1.ts
```diff
@@ -74,6 +74,16 @@ export class Secp256k1PublicKey extends PublicKey {
     const bytes = deserializer.deserializeBytes();
     return new Secp256k1PublicKey(bytes);
   }
+
+  serializeForSingleKey(serializer: Serializer): void {
```
0xmaayan, line 78: ignore

src/core/crypto/secp256k1.ts
```diff
@@ -224,4 +234,14 @@ export class Secp256k1Signature extends Signature {
     const hex = deserializer.deserializeBytes();
     return new Secp256k1Signature(hex);
   }
+
+  serializeForSingleKey(serializer: Serializer): void {
```
0xmaayan, line 238: ignore

src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -319,27 +317,30 @@ export function generateSignedTransactionForSimulation(args: SimulateTransaction
...
 }
 
 export function getAuthenticatorForSimulation(publicKey: PublicKey) {
-  if (publicKey instanceof Ed25519PublicKey) {
-    return new AccountAuthenticatorEd25519(
-      new Ed25519PublicKey(publicKey.toUint8Array()),
-      new Ed25519Signature(new Uint8Array(64)),
-    );
+  // TODO add support for AnyMultiKey
+  if (publicKey instanceof AnyPublicKey) {
```
0xmaayan, line 330: need to test it

src/core/account.ts
```diff
@@ -97,62 +99,77 @@ export class Account {
...
+      authenticationKey: authKey.toString(),
+    });
+
+    try {
+      await getInfo({
+        aptosConfig: config,
+        accountAddress: potentialAddress1.toString(),
+      });
+      return true;
+    } catch (error: any) {
```
xbtmatt, line 170: I think we should only swallow the error if it's `resource_not_found` or whatever, that way if it's some other API error, we're not assuming the account wasn't found and throwing an actual error

src/core/account.ts
```diff
@@ -73,54 +84,108 @@ export class Account {
...
-   * @param scheme optional SigningScheme - type of SigningScheme to use. Default to Ed25519
+   * @param args.scheme optional. SigningScheme - type of SigningScheme to use. Default to Ed25519
    * Currently only Ed25519 and Secp256k1 are supported
+   * @param args.legacy optional. If set to true, would create a legacy Ed25519 signing keys. Default
+   * is Single Sender signing key
    *
    * @returns Account with the given signing scheme
    */
-  static generate(scheme?: SigningScheme): Account {
+  static generate(args?: { scheme?: SigningSchemeInput; legacy?: boolean }): Account {
```
gregnazario, line 94: Hmm, why is it SigningSchemeInput and not SigningScheme, is SigningScheme something else now?

src/core/crypto/singleKey.ts
```diff
@@ -0,0 +1,78 @@
...
+import { HexInput } from "../../types";
+import { Hex } from "../hex";
+import { PublicKey } from "./asymmetricCrypto";
+import { Ed25519Signature, Ed25519PublicKey } from "./ed25519";
+
+export class SingleKey extends PublicKey {
+  /**
+   * Length of an Ed25519 public key
+   */
+  static readonly LENGTH: number = 65;
```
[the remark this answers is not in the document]
0xmaayan: yea, copy pasta. should even be here

src/core/account.ts
```diff
@@ -73,54 +84,108 @@ export class Account {
...
+    if (privateKey instanceof Ed25519PrivateKey) {
+      // lookup single sender ed25519
+      const SingleSenderTransactionAuthenticatorAuthKey = AuthenticationKey.fromBytesAndScheme({
+        publicKey,
+        scheme: SigningScheme.SingleKey,
+      });
+      const isSingleSenderTransactionAuthenticator = await Account.lookupAddress(
+        SingleSenderTransactionAuthenticatorAuthKey,
+        aptosConfig,
+      );
```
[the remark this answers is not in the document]
0xmaayan: good point. any preferences in using only Key + Address + Scheme computing instead of using lookupAddress? do we need to consider key rotation?

src/core/account.ts
```diff
@@ -73,54 +84,108 @@ export class Account {
...
+      await getInfo({
+        aptosConfig,
+        accountAddress: accountAddress.toString(),
+      });
+      return true;
+    } catch (error: any) {
+      // account not found
+      if (error.code === 404) {
+        return false;
+      }
```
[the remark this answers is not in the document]
0xjinn: Yep, if not found in the table, it will return the same address as the auth key

src/core/account.ts
0xmaayan: `SigningSchemeInput` is for the user to provide the signing scheme they want to use. i.e `ED25519` , `Secp256k1`, `MultiED25519`, etc.

The inner `SigningScheme` now uses the new flow `SingleSender` and later `MultiSender` so we can create it based on that.

src/core/account.ts
```diff
@@ -45,23 +48,31 @@ export class Account {
...
+   * @param args.legacy optional. If set to true, would create a legacy Ed25519 signing keys. Default
+   * is Single Sender signing key
    *
    * This method is private because it should only be called by the factory static methods.
    * @returns Account
    */
-  private constructor(args: { privateKey: PrivateKey; address: AccountAddress }) {
-    const { privateKey, address } = args;
+  private constructor(args: { privateKey: PrivateKey; address: AccountAddress; legacy?: boolean }) {
+    const { privateKey, address, legacy } = args;
```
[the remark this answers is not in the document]
0xmaayan: so we have 3 ways to derive keys.
- "generate()" which generates a new key set for you - here the thought is to default to the new single sender flow with the "legacy" open
- "fromPrivateKey()" - that solves the old/new flow for you based on the provided private key
- "fromPath()" - which it seems only support old ED25519 flow

The `constructor `  is a private method, and called only from the 2 mentioned functions. So it looks like we technically only support "legacy" when you want to generate new keys, wtbs - when a wallet generates  new keys, do they have to use the old legacy flow or can use the new single sender flow?

src/api/transaction.ts
```diff
@@ -104,7 +104,7 @@ export class Transaction {
    */
   async waitForTransaction(args: {
     transactionHash: HexInput;
-    options?: { timeoutSecs?: number; checkSuccess?: boolean };
+    options?: { timeoutSecs?: number; checkSuccess?: boolean; indexerVersionCheck?: boolean };
```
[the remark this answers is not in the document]
0xmaayan: it is only here b/c indexer doesnt support single sender yet

src/core/account.ts
```diff
@@ -133,10 +198,12 @@ export class Account {
    */
   static fromDerivationPath(args: { path: string; mnemonic: string }): Account {
     const { path, mnemonic } = args;
-
     const { key } = derivePrivateKeyFromMnemonic(KeyType.ED25519, path, mnemonic);
     const privateKey = new Ed25519PrivateKey(key);
-    return Account.fromPrivateKey(privateKey);
+    const publicKey = privateKey.publicKey();
+    const authKey = Account.authKey({ publicKey });
```
0xjinn, line 204: nit - Looks like we missed one when reverting obj for param back to regular param. I clean this update later on.

src/core/account.ts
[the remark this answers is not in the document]
0xmaayan: that doesnt help in this use case where I dont want it to return anything as I rely on whether the account had found to understand the signing scheme

src/core/account.ts
```diff
@@ -73,54 +84,108 @@ export class Account {
...
+   * @param aptosConfig AptosConfig type
+   *
+   * @returns Promise<Account>
    */
-  static fromPrivateKey(privateKey: PrivateKey): Account {
-    const publicKey = privateKey.publicKey();
-    const authKey = Account.authKey({ publicKey });
-    const address = new AccountAddress({ data: authKey.toUint8Array() });
-    return Account.fromPrivateKeyAndAddress({ privateKey, address });
+  static async fromPrivateKey(privateKey: PrivateKey, aptosConfig: AptosConfig): Promise<Account> {
```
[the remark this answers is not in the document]
gregnazario: We don't lookup the address onchain, but we do for the Ed25519 keys.

src/core/account.ts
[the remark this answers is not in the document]
0xjinn: Ahh ok, i see what you mean. Though I wonder if we should change this name? `lookupAddress` that returns bool seems odd.

src/core/crypto/anyPublicKey.ts
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Verifies a signed data with a public key
+   *
+   * @param args.message message
+   * @param args.signature The signature
+   * @returns true if the signature is valid
+   */
+  verifySignature(args: { message: HexInput; signature: AnySignature }): boolean {
+    const { message, signature } = args;
+    if (this.publicKey instanceof Ed25519PublicKey && signature.signature instanceof Ed25519Signature) {
```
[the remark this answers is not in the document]
0xmaayan: can do something like that
```
if (this.isED25519Signature(signature)) {
.....
} else if (this.isSecp256k1Signature(signature)) {
.....
}
```

src/core/account.ts
```diff
@@ -133,10 +198,12 @@ export class Account {
...
   static fromDerivationPath(args: { path: string; mnemonic: string }): Account {
     const { path, mnemonic } = args;
-
     const { key } = derivePrivateKeyFromMnemonic(KeyType.ED25519, path, mnemonic);
     const privateKey = new Ed25519PrivateKey(key);
-    return Account.fromPrivateKey(privateKey);
+    const publicKey = privateKey.publicKey();
+    const authKey = Account.authKey({ publicKey });
+    const address = new AccountAddress({ data: authKey.toUint8Array() });
+    return new Account({ privateKey, address, legacy: true });
```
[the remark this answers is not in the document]
0xmaayan: this function only supports legacy ED25519 currently, is it meant to support both legacy and new single sender?

src/internal/transaction.ts
```diff
@@ -103,7 +103,7 @@ export async function waitForTransaction(args: {
   const { aptosConfig, transactionHash, options } = args;
   const timeoutSecs = options?.timeoutSecs ?? DEFAULT_TXN_TIMEOUT_SEC;
   const checkSuccess = options?.checkSuccess ?? true;
-  const indexerVersionCheck = options?.indexerVersionCheck ?? true;
+  const indexerVersionCheck = options?.indexerVersionCheck ?? false;
```
0xmaayan, line 106: need to remove

src/internal/faucet.ts
```diff
@@ -35,7 +35,11 @@ export async function fundAccount(args: {
 
   const txnHash = data.txn_hashes[0];
 
-  await waitForTransaction({ aptosConfig, transactionHash: txnHash, options: { timeoutSecs } });
+  await waitForTransaction({
```
0xmaayan, line 38: probably can be removed after https://github.com/aptos-labs/aptos-ts-sdk/pull/110#pullrequestreview-1693839440

gregnazario (approved):

src/api/account.ts
```diff
@@ -357,4 +358,22 @@ export class Account {
...
   }
+
+  /**
+   * Derives an account by providing a private key.
+   * This functions resolves the provided private key type and derives the public key from it.
+   *
+   * If the privateKey is a Secp256k1 type, it derives the account using the derived public key and
+   * auth key using the SingleKey scheme locally.
+   *
+   * If the privateKey is a ED25519 type, it looksup the auth key on chain, and uses it to resolve
```
xbtmatt, line 369:
```suggestion
   * If the privateKey is a ED25519 type, it looks up the authentication key on chain, and uses it to resolve
```

src/api/account.ts
```diff
@@ -357,4 +358,22 @@ export class Account {
...
+   * whether it is a Legacy ED25519 key or a Unified ED25519 key. It then derives the account based
```
xbtmatt, line 370: Love the name `Unified Ed25519` key! Much clearer

src/core/account.ts
```diff
@@ -45,23 +62,31 @@ export class Account {
    *
    * @param args.privateKey PrivateKey - private key of the account
    * @param args.address AccountAddress - address of the account
+   * @param args.legacy optional. If set to true, would create a Legacy signing keys. Default
+   * is Unified signing key
```
xbtmatt, line 66:
```suggestion
   * @param args.legacy optional. If set to true, the keypair generated is a Legacy keypair. Defaults
   * to generating a Unified keypair
```

src/core/account.ts
```diff
@@ -71,56 +96,70 @@ export class Account {
...
-    return Account.fromPrivateKeyAndAddress({ privateKey, address });
-  }
-
-  /**
-   * Derives an account with provided private key and address
-   * This is intended to be used for account that has it's key rotated
+   * @param privateKey PrivateKey - private key of the account
+   * @param address The account address
+   * @param legacy optional. If set to true, would create a Legacy signing keys. Default
+   * is Unified signing key
```
xbtmatt, line 156:
```suggestion
   * @param legacy optional. If set to true, the keypair generated is a Legacy keypair. Defaults
   * to generating a Unified keypair
```

tests/e2e/transaction/signTransaction.test.ts
```diff
@@ -1,139 +1,197 @@
...
     });
-
-    describe.skip("Secp256k1", () => {
+    describe("Single Sender Secp256k1", () => {
+      test("it signs a script transaction", async () => {
+        const rawTxn = await aptos.generateTransaction({
+          sender: singleSignerSecp256k1Account.accountAddress.toString(),
+          data: {
+            bytecode: singleSignerScriptBytecode,
+            functionArguments: [new U64(1), recieverAccounts[0].accountAddress],
```
xbtmatt, line 95:
```suggestion
            functionArguments: [new U64(1), receiverAccounts[0].accountAddress],
```

ctrl + h `reciever` and replace with `receiver` :p 😁

src/api/account.ts
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { AptosConfig } from "./aptosConfig";
-import { AccountAddress } from "../core";
+import { AccountAddress, PrivateKey, Account as AccountModule } from "../core";
```
xbtmatt, line 5:
```suggestion
import { AccountAddress, PrivateKey, Account } from "../core";
```

I'm assuming this snuck in there somehow

src/api/account.ts
```diff
@@ -357,4 +358,22 @@ export class Account {
...
+   * auth key using the SingleKey scheme locally.
+   *
+   * If the privateKey is a ED25519 type, it looksup the auth key on chain, and uses it to resolve
+   * whether it is a Legacy ED25519 key or a Unified ED25519 key. It then derives the account based
+   * on that.
+   *
+   * @param args.privateKey An account private key
+   * @returns Account type
+   */
+  async deriveAccountFromPrivateKey(args: { privateKey: PrivateKey }): Promise<AccountModule> {
```
xbtmatt, line 376:
```suggestion
  async deriveAccountFromPrivateKey(args: { privateKey: PrivateKey }): Promise<Account> {
```

src/api/account.ts
0xmaayan: well Account is the name of the current class, had to use a different name

PR #86, opened 2023-10-17
[refactor] File case consistency

gregnazario:
### Description
This ensures that all files are using camelCase.  Not kebob-case, and not snake_case.  This is part of the consistency so that users don't have to find which path to use.

### Test Plan

CI


src/api/aptos.ts
```diff
@@ -2,17 +2,17 @@
...
 
 import { Account } from "./account";
-import { AptosConfig } from "./aptos_config";
+import { AptosConfig } from "./aptosConfig";
 import { Coin } from "./coin";
 import { Collection } from "./collection";
 import { Event } from "./event";
 import { Faucet } from "./faucet";
-import { FungibleAsset } from "./fungible_asset";
+import { FungibleAsset } from "./fungibleAsset";
```
0xjinn, line 10: I just updated this from my recent PR. 😅 Probably conflict

0xjinn (approved):

0xmaayan (approved):

PR #87, opened 2023-10-18
Add createCollection API to the SDK

heliuchuan:
### Description
This adds a function for users to invoke 0x4::aptos_token::create_collection via the SDK

### Test Plan

pnpm test

### DIFF
--- a/src/api/collection.ts
+++ b/src/api/collection.ts
@@ -2,7 +2,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { AptosConfig } from "./aptosConfig";
-import { getCollectionId, getCollectionData } from "../internal/collection";
+import {
+  CreateCollectionOptions,
+  createCollectionTransaction,
+  getCollectionId,
+  getCollectionData,
+} from "../internal/collection";
+import { Account } from "../core";
+import { GenerateTransactionOptions, SingleSignerTransaction } from "../transactions/types";
 import { GetCollectionDataResponse, HexInput, TokenStandard } from "../types";
 
 /**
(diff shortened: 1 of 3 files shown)


0xmaayan (dismissed):
this should be under `digitalAsset ` class as all `aptos_token` operations. we can skip it for now as this is a small implementation of a bigger item.
Or at least have this api to be under digitalAsset

gregnazario:
> this should be under `digitalAsset ` class as all `aptos_token` operations. we can skip it for now as this is a small implementation of a bigger item. Or at least have this api to be under digitalAsset

Yes, let's combine `collection` and `token` under `digitalAsset`

gregnazario:
@heliuchuan just for reference, I addressed the merge conflicts, we can get these two PRs in, and then combine them into one digitalAssets, which should just be an easy rename.

tests/e2e/api/collection.test.ts
```diff
@@ -1,44 +1,54 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-// import { Aptos, AptosConfig, Network } from "../../../src";
+import { Account, Aptos, AptosConfig, Network } from "../../../src";
+import { waitForTransaction } from "../../../src/internal/transaction";
+import { FUND_AMOUNT } from "../../unit/helper";
+
+// use it here since all tests use the same configuration
+const config = new AptosConfig({ network: Network.DEVNET });
```
gregnazario, line 9: I believe we will want to change these to LOCAL, but let's evaluate that at a combined level

gregnazario (approved):

src/api/collection.ts
```diff
@@ -15,6 +22,43 @@ export class Collection {
...
+      creator: Account;
+      description: string;
+      name: string;
+      uri: string;
+      options?: GenerateTransactionOptions;
+    } & CreateCollectionOptions,
+  ): Promise<SingleSignerTransaction> {
+    return createCollectionTransaction({ aptosConfig: this.config, ...args });
+  }
+
```
gregnazario, line 61: Returning a SingleSigner may be weird if we also support Fee payer, but I'm not sure how that model is going to look cc @0xmaayan

PR #91, opened 2023-10-18
Initial ANS contract apis (registerDomain and getOwnerAddress)

kaw2k:
### Description
This is a start to porting over the ANS contract from the previous SDK. I want to make sure I am aligned with the direction of things before committing too much code. This PR:
* Ports over the script to fetch the ANS contract and publish it locally
* ~Extends the base config to allow for custom ANS router contract addresses (for use in local / devnet / custom networks).~
* Implements [`register_domain`](https://explorer.aptoslabs.com/account/0x5f8fd2347449685cf41d4db97926ec3a096eaf381332be4f1318ad4d16a8497c/modules/run/domains/register_domain?network=testnet) and [`get_ownder_addr`](https://explorer.aptoslabs.com/account/0x5f8fd2347449685cf41d4db97926ec3a096eaf381332be4f1318ad4d16a8497c/modules/view/router/get_owner_addr?network=testnet). 

~**Note** There is a bit of copy pasta going on. I have not added jsdocs or payed too much head to them yet. I will finalize them after feedback .~

### Test Plan
Currently there are two e2e tests that use both `registerDomain` and `getOwnerAddress`. More tests coming as I get through the rest of the contract.

### Related Links
* [The previous implementation](https://github.com/aptos-labs/aptos-core/blob/cda8f80baa9d70b9c35f320a2acb7247f7835949/ecosystem/typescript/sdk/src/plugins/ans_client.ts#L1) 
* [The contract on testnet](https://explorer.aptoslabs.com/account/0x5f8fd2347449685cf41d4db97926ec3a096eaf381332be4f1318ad4d16a8497c/modules/run/router/register_domain?network=testnet)


package.json
```diff
@@ -26,7 +26,8 @@
     "lint": "eslint 'src/**/*.ts' 'tests/**/*.ts'",
     "test": "pnpm jest",
     "unit-test": "pnpm jest tests/unit",
-    "e2e-test": "pnpm jest tests/e2e",
+    "e2e-test": "pnpm run _e2e-test:publish-contracts && jest tests/e2e",
+    "_e2e-test:publish-contracts": "ts-node -r dotenv/config ./scripts/publishANSContracts.mts",
```
kaw2k, line 30: Pulling and building the contract isn't a cheap operation. Just wanted to highlight that for consideration.

tsconfig.json
```diff
@@ -16,5 +16,8 @@
     "strict": true,
     "target": "es2020"
   },
-  "include": ["src", "tests"]
+  "include": ["src", "tests"],
+  "ts-node": {
+    "esm": true
+  }
```
kaw2k, line 22: `ts-node` was fighting me with the publish script. If anyone has more familiarity there, open to feedback.

src/api/name.ts
```diff
@@ -0,0 +1,44 @@
...
+      );
+    }
+  }
+
+  async mintDomainName(args: {
+    sender: Account;
+    domainName: string;
+    registrationDuration: 31536000;
+    targetAddress?: MoveAddressType;
+    toAddress?: MoveAddressType;
```
kaw2k, line 39: I settled on `MoveAddressType` here, would `HexInput` be more appropriate?

src/api/name.ts
```diff
@@ -0,0 +1,44 @@
...
+          "Note, there is no ANS contract published on devnet.",
+        ].join(" "),
+      );
+    }
+  }
+
+  async mintDomainName(args: {
+    sender: Account;
+    domainName: string;
+    registrationDuration: 31536000;
```
kaw2k, line 37: This is an oddity of the contract at the moment. Callers can provide increments of a year in seconds. However at the moment we only accept one year. I chose to keep the SDK as close to the contract as possible, to reduce confusion if the caller switches between the contract and the SDK. This will be relaxed to `number` (with a proper description in jsdocs) when the contract supports registration times outside of a year.

src/api/name.ts
```diff
@@ -0,0 +1,44 @@
...
+      console.warn(
+        [
+          "No ANS Router address found. This most likely means you are using",
+          "a custom network, a local network, or the devnet. If you plan on using",
+          "the Aptos Names API, you must provide a router address in the AptosConfig",
+          "(such as { network: Network.LOCAL, ansRouter: '0x123...' })",
+          "Note, there is no ANS contract published on devnet.",
+        ].join(" "),
+      );
+    }
```
kaw2k, line 31: I don't want to block the use of custom networks, local, or devnet, that don't use the `Name` class. I decided to warn upon instantiation and throw errors if the user tries to call a class method without properly configuring the router.

src/internal/name.ts
```diff
@@ -0,0 +1,77 @@
...
+  testnet: "0x5f8fd2347449685cf41d4db97926ec3a096eaf381332be4f1318ad4d16a8497c",
+  mainnet: "0x867ed1f6bf916171b1de3ee92849b8978b7d1b9e0a8cc982a3d19d535dfd9c0c",
+  devnet: "" as MoveAddressType,
+  local: "" as MoveAddressType,
+  custom: "" as MoveAddressType,
+};
+
+function getRouterAddress(aptosConfig: AptosConfig): MoveAddressType {
+  const address = aptosConfig.ansRouter || RouterAddressLookup[aptosConfig.network];
+  if (!address) throw new Error("To use the Name API, you must provide a router address in the AptosConfig.");
```
kaw2k, line 30: Is there a more specific `Error` to throw?

src/internal/name.ts
```diff
@@ -0,0 +1,77 @@
...
+}): Promise<SingleSignerTransaction> {
+  const routerAddress = getRouterAddress(aptosConfig);
+  validateMultipleOfYearInSeconds(registrationDuration);
+
+  const transaction = await generateTransaction({
+    aptosConfig,
+    sender: sender.accountAddress.toString(),
+    data: {
+      function: `${routerAddress}::router::register_domain`,
+      typeArguments: [],
```
kaw2k, line 65: Wasn't sure what to put here.

src/internal/name.ts
```diff
@@ -0,0 +1,77 @@
...
+      arguments: [
+        new MoveString(domainName),
+        new U64(registrationDuration),
+        new MoveOption(targetAddress ? AccountAddress.fromHexInput(targetAddress) : null),
+        new MoveOption(toAddress ? AccountAddress.fromHexInput(toAddress) : null),
```
kaw2k, line 70: Just double checking I was correctly using `MoveOption` here. The contract doesn't require these fields.

package.json
0xmaayan: note: we also need to make sure we run it on CI https://github.com/aptos-labs/aptos-core/blob/main/.github/actions/run-ts-sdk-e2e-tests/action.yaml#L38

src/api/ans.ts
```diff
@@ -0,0 +1,63 @@
...
+   * ```
+   *
+   * @param args
+   * @param args.domainName - A string of the domain name to retrieve
+   * @param args.subdomainName - A string of the subdomain name to retrieve
+   *
+   * @returns MoveAddressType if the name is owned, undefined otherwise
+   */
+  async getOwnerAddress(args: { domainName: string; subdomainName?: string }): Promise<MoveAddressType | undefined> {
+    return getOwnerAddress({ ...args, aptosConfig: this.config });
```
0xmaayan, line 35: dont think it makes any difference, but it is a standard
```suggestion
    return getOwnerAddress({ aptosConfig: this.config, ...args });
```

src/api/ans.ts
```diff
@@ -0,0 +1,63 @@
...
+   */
+  async registerDomain(args: {
+    sender: Account;
+    domainName: string;
+    registrationDuration: 31536000;
+    targetAddress?: HexInput;
+    toAddress?: HexInput;
+    options?: GenerateTransactionOptions;
+  }): Promise<SingleSignerTransaction> {
+    return registerDomain({ ...args, aptosConfig: this.config });
```
0xmaayan, line 61: same

src/api/aptos.ts
```diff
@@ -61,6 +64,7 @@ export class Aptos {
     this.event = new Event(this.config);
     this.faucet = new Faucet(this.config);
     this.general = new General(this.config);
+    this.ans = new ANS(this.config);
```
0xmaayan, line 67: can we order it alphabetically

src/api/aptos.ts
```diff
@@ -107,6 +112,7 @@ applyMixin(Aptos, Collection, "collection");
 applyMixin(Aptos, Event, "event");
 applyMixin(Aptos, Faucet, "faucet");
 applyMixin(Aptos, General, "general");
+applyMixin(Aptos, ANS, "ans");
```
0xmaayan, line 115: same

src/internal/ans.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  const transaction = await generateTransaction({
+    aptosConfig,
+    sender: sender.accountAddress.toString(),
+    data: {
+      function: `${routerAddress}::router::register_domain`,
+      arguments: [
+        new MoveString(domainName),
+        new U64(registrationDuration),
+        new MoveOption(targetAddress ? AccountAddress.fromHexInput(targetAddress) : null),
```
0xmaayan, line 76: cc @xbtmatt

src/api/ans.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Retrieve the owner address of a domain name or subdomain name.
+   *
+   * ```ts
+   * getOwnerAddress({domainName: "aptos", subdomainName: "test"})
+   * // Will return the owner address of "test.aptos.apt" or undefined
+   * ```
+   *
+   * @param args
```
0xmaayan, line 28:
```suggestion

```

src/api/ans.ts
```diff
@@ -0,0 +1,63 @@
...
+   * @returns MoveAddressType if the name is owned, undefined otherwise
+   */
+  async getOwnerAddress(args: { domainName: string; subdomainName?: string }): Promise<MoveAddressType | undefined> {
+    return getOwnerAddress({ ...args, aptosConfig: this.config });
+  }
+
+  /**
+   * Registers a new domain name
+   *
+   * @param args
```
0xmaayan, line 41:
```suggestion

```

.github/actions/run-tests/action.yaml
```diff
@@ -17,6 +17,13 @@ runs:
       with:
         version: 8.9.0
 
+    # Set up the necessary env vars for the test suite.
+    - run: echo "ANS_TEST_ACCOUNT_PRIVATE_KEY=0x37368b46ce665362562c6d1d4ec01a08c8644c488690df5a17e13ba163e20221" >> .env
+      shell: bash
+    - run: echo "ANS_TEST_ACCOUNT_ADDRESS=0x585fc9f0f0c54183b039ffc770ca282ebd87307916c215a3e692f2f8e4305e82" >> .env
+      shell: bash
```
kaw2k, line 24: @0xmaayan I am not sure if this is correct, will test in CI

src/internal/ans.ts
```diff
@@ -0,0 +1,119 @@
...
+// eslint-disable-next-line eqeqeq
+const Option = <T>(value: T | undefined | null): MoveValue => (value != undefined ? Some(value) : None());
+
+const unwrapOption = <T>(option: any): T | undefined => {
+  if (!!option && typeof option === "object" && "vec" in option && Array.isArray(option.vec)) {
+    return option.vec[0];
+  }
+
+  return undefined;
+};
```
kaw2k, line 59: @0xmaayan (cc @gregnazario)

The ANS contract has numerous optional types as arguments and as results, so I included some helper functions. I got a little turned in circles reconciling when to use `MoveOption` vs when not to. Looks like `view` expects more primitive types? 

What are your opinions on these helpers? Should I keep them or get rid of them? Move them to a different location?

gregnazario: Oh, that's a big mess we can see what we can do here to clean this up.  I'm rewriting the parsing logic

0xmaayan: cc @xbtmatt

scripts/publishANSContracts.mts
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * This script runs when testing locally and on CI (as part of sdk-release.yaml) using `pnpm test`.
+ */
+
+// URLs at which the node and faucet are running.
+const APTOS_NODE_URL = process.env.APTOS_NODE_URL || "http://localhost:8080/v1";
+const APTOS_FAUCET_URL = process.env.APTOS_FAUCET_URL || "http://localhost:8081";
+
+// Env vars that configure how we run the CLI.
+const DOCKER_IMAGE = process.env.DOCKER_IMAGE;
```
0xmaayan, line 20: I dont think we use docker anymore to run the CLI, cc @banool

tests/e2e/api/ans.test.ts
```diff
@@ -0,0 +1,101 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { Aptos, Network, Account, AnyRawTransaction, Ed25519PrivateKey, U8, AptosConfig } from "../../../src";
+import { LOCAL_ANS_ACCOUNT_ADDRESS, LOCAL_ANS_ACCOUNT_PK } from "../../../src/internal/ans";
+import { generateTransaction } from "../../../src/internal/transactionSubmission";
+
+describe("ANS", () => {
+  const config = new AptosConfig({ network: Network.LOCAL });
+  const aptos = new Aptos(config);
+  const contractAccount = Account.fromPrivateKey(new Ed25519PrivateKey(LOCAL_ANS_ACCOUNT_PK));
```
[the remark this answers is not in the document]
kaw2k: Just double checked, they are the same!

src/internal/ans.ts
kaw2k: @gregnazario did you get a chance to address this logic?

gregnazario: Yes, the new SDK inputs should automatically do all this unwrapping as long as your ABI of your function is the same

src/api/ans.ts
```diff
@@ -30,7 +30,7 @@ export class ANS {
    *
    * @returns MoveAddressType if the name is owned, undefined otherwise
    */
-  async getOwnerAddress(args: { domainName: string; subdomainName?: string }): Promise<MoveAddressType | undefined> {
+  async getOwnerAddress(args: { domainName: string; subdomainName?: string }): Promise<AccountAddress | undefined> {
     return getOwnerAddress({ aptosConfig: this.config, ...args });
```
gregnazario: Hmm, just thinking about this from a user perspective.  Would this want to just be `name` and parse the `domain` and `subdomain` out of it?  Might be less confusing / more adaptable.

For example, if I just throw my ANS name test.greg.apt into a website, I hope it'll be able to just figure it out, without the website builder having to always write that parsing logic themselves.

.github/actions/run-tests/action.yaml
gregnazario: Where I moved the ANS installation to, you won't need this.

src/api/ans.ts
```diff
@@ -0,0 +1,61 @@
...
+   * ```
+   *
+   * @param args.domainName - A string of the domain name to retrieve
+   * @param args.subdomainName - A string of the subdomain name to retrieve
+   *
+   * @returns MoveAddressType if the name is owned, undefined otherwise
+   */
+  async getOwnerAddress(args: { domainName: string; subdomainName?: string }): Promise<AccountAddress | undefined> {
+    return getOwnerAddress({ aptosConfig: this.config, ...args });
+  }
```
gregnazario, line 35: I stared at this a bit.

Can we make an easy mode, where we take in say... `test.greg.apt` and parse it out into it's parts?

Otherwise, the dapp or dev will have to write a wrapper to do this every time they take an ANS name from a user.  I doubt that they'll already be split into pieces.

src/api/ans.ts
```diff
@@ -0,0 +1,61 @@
...
+   * the sender's address will be used.
+   * @param args.toAddress optional - The address to send the domain name to. If not provided,
+   * the transaction will be sent to the router.
+   *
+   * @returns InputSingleSignerTransaction
+   */
+  async registerDomain(args: {
+    sender: Account;
+    domainName: string;
+    registrationDuration: 31536000;
```
gregnazario, line 54: This is hardcoded?  What happens if it isn't this number? (basically seems strange to give a number as an input parameter if there isn't any choice)

src/internal/ans.ts
```diff
@@ -0,0 +1,121 @@
...
+import { MoveOption, MoveString, U64 } from "../bcs";
+import { Account, AccountAddress } from "../core";
+import { InputGenerateTransactionOptions, InputSingleSignerTransaction } from "../transactions/types";
+import { HexInput, MoveAddressType, MoveValue } from "../types";
+import { Network } from "../utils/apiEndpoints";
+import { view } from "./general";
+import { generateTransaction } from "./transactionSubmission";
+
+export const LOCAL_ANS_ACCOUNT_PK =
+  (process.env.ANS_TEST_ACCOUNT_PRIVATE_KEY as MoveAddressType) ||
```
gregnazario, line 21: Probably good to share these with the tests directly.

src/internal/ans.ts
```diff
@@ -0,0 +1,121 @@
...
+  [Network.TESTNET]: "0x5f8fd2347449685cf41d4db97926ec3a096eaf381332be4f1318ad4d16a8497c",
+  [Network.MAINNET]: "0x867ed1f6bf916171b1de3ee92849b8978b7d1b9e0a8cc982a3d19d535dfd9c0c",
+  [Network.LOCAL]: LOCAL_ANS_ACCOUNT_ADDRESS,
+  [Network.CUSTOM]: null,
+  [Network.DEVNET]: null,
+};
+
+function getRouterAddress(aptosConfig: AptosConfig): MoveAddressType {
+  const address = NetworkToAnsContract[aptosConfig.network];
+  if (!address) throw new Error("The ANS contract is not deployed to your network");
```
gregnazario, line 37: I would say the network here

src/internal/ans.ts
```diff
@@ -0,0 +1,121 @@
...
+  return address;
+}
+
+function validateMultipleOfYearInSeconds(seconds: number): void {
+  if (seconds % 31536000 !== 0 || seconds === 0) {
+    throw new Error("Registration duration must be a multiple of 1 year in seconds, greater than 0");
+  }
+}
```
gregnazario, line 45: This allows for 0, so I think you want to change to || seconds !== 0?

src/internal/ans.ts
gregnazario: This needs more love in Remote ABI for view functions, but for any entry functions you can remove this.

src/internal/ans.ts
```diff
@@ -0,0 +1,121 @@
...
+  domainName,
+  registrationDuration,
+  targetAddress,
+  toAddress,
+  options,
+}: {
+  aptosConfig: AptosConfig;
+  sender: Account;
+  domainName: string;
+  registrationDuration: 31536000;
```
gregnazario, line 73: This is weird with a fixed number, maybe not even provide it for now?

src/internal/ans.ts
```diff
@@ -0,0 +1,121 @@
...
+    aptosConfig,
+    payload: {
+      function: `${routerAddress}::router::get_owner_addr`,
+      functionArguments: [domainName, Option(subdomainName)],
+    },
+  });
+
+  const owner = unwrapOption(res[0]) as MoveAddressType | undefined;
+
+  return owner ? AccountAddress.fromHexInput(owner) : undefined;
```
gregnazario, line 120: We should just use the type directly and avoid casting

```suggestion
      functionArguments: [domainName, Option(subdomainName)],
    },
  });

  const owner = unwrapOption<MoveAddressType>(res[0]);

  return owner ? AccountAddress.fromHexInput(owner) : undefined;
```

tsconfig.json
```diff
@@ -16,5 +16,5 @@
     "strict": true,
     "target": "es2020"
   },
-  "include": ["src", "tests"]
+  "include": ["src", "tests"],
```
gregnazario, line 19: fyi, I think the last comma might have to go (maybe I messed that up) since json technically doesn't have extra commas

src/api/ans.ts
kaw2k: This is a quirk of the contract currently. The duration must be in increments of a year (in seconds). Currently the contract only accepts a single year, but will expand to multi-year registrations once they figure out their pricing model. 

When making these contracts my first impression was to stick close to the contracts, that way there would be consistency and developers wouldn't get confused if they were referencing the contract. After talking yesterday, perhaps it is better to favor simplicity. 

I could change this from inputting seconds to inputing years, and do the math for users.

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
0xmaayan, line 1: is this expected?

src/api/ans.ts
```diff
@@ -0,0 +1,58 @@
...
+   * getOwnerAddress({domainName: "aptos", subdomainName: "test"})
+   * // Will return the owner address of "test.aptos.apt" or undefined
+   * ```
+   *
+   * @param args.domainName - A string of the domain name to retrieve
+   * @param args.subdomainName - A string of the subdomain name to retrieve
+   *
+   * @returns MoveAddressType if the name is owned, undefined otherwise
+   */
+  async getOwnerAddress(args: { name: ANSName }): Promise<AccountAddress | undefined> {
```
0xmaayan, line 31: do we want to return the full AccountAddress object? or just the address (as string)?

src/internal/ans.ts
```diff
@@ -0,0 +1,192 @@
...
+
+export async function registerName({
+  aptosConfig,
+  sender,
+  name,
+  expiration,
+  toAddress,
+  targetAddress,
+  options,
+}: RegisterNameParameters) {
```
0xmaayan, line 149: add function return type

src/api/ans.ts
```diff
@@ -0,0 +1,58 @@
...
+   * - subdomain:independent: The name will expire at the given date.
+   * @param args.transferable  - Determines if the subdomain being minted is soul-bound. Applicable only to subdomains.
+   * @param args.targetAddress optional - The address the domain name will resolve to. If not provided,
+   * the sender's address will be used.
+   * @param args.toAddress optional - The address to send the domain name to. If not provided,
+   * the transaction will be sent to the router.
+   *
+   * @returns InputSingleSignerTransaction
+   */
+  async registerName(args: Omit<RegisterNameParameters, "aptosConfig">): ReturnType<typeof registerName> {
```
0xmaayan, line 55: return `Promise<InputSingleSignerTransaction>`? not sure what `ReturnType<typeof registerName>` is for

tests/e2e/ans/ans.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      const name = randomString();
+      const alice = Account.generate();
+      await aptos.fundAccount({
+        accountAddress: alice.accountAddress.toString(),
+        amount: 500_000_000,
+      });
+
+      await signAndSubmit(
+        alice,
+        await aptos.ans.registerName({
```
0xmaayan, line 134:
```suggestion
        await aptos.registerName({
```

tests/e2e/ans/ans.test.ts
```diff
@@ -0,0 +1,175 @@
...
+          sender: alice,
+          name,
+          // Force the year to be absent
+          expiration: { policy: "domain" } as any,
+        }),
+      ).rejects.toThrow();
+
+      // Testing to make sure that the subdomain policy is enforced
+      await expect(
+        aptos.ans.registerName({
```
0xmaayan, line 115:
```suggestion
        aptos.registerName({
```

tests/e2e/ans/publishANSContracts.ts
```diff
@@ -0,0 +1,106 @@
+import { execSync } from "child_process";
```
0xmaayan, line 1: lets put this file under `~/scripts/` , we would need to run it before all tests

tests/e2e/ans/ans.test.ts
kaw2k: Any particular reason to drop the namespace in this particular instance (and the one bellow)?

tests/e2e/ans/publishANSContracts.ts
kaw2k: @gregnazario improved the flow here! [He moved the invocation of the script to the e2e test file itself](https://github.com/aptos-labs/aptos-ts-sdk/blob/02173dbf29a168564b8ab587aa3a3fc1675703d9/tests/e2e/ans/ans.test.ts#L24-L25). Now, whenever this test runs it will ensure it has a published contract, regardless of the command used to make it run. Keeping it colocated to the test makes sense with this change!

pnpm-lock.yaml
kaw2k: Nope, not sure how this got updated. I'll revert this file.

tests/e2e/ans/publishANSContracts.ts
0xmaayan: how then you promise that this file runs before the actual `ans.test.ts` file?

tests/e2e/ans/ans.test.ts
0xmaayan: no particular reason with those specific cases, it is in general that we can drop the namespace - same as we do with everything. Just didn't want to spam the PR with the same comment everywhere

src/internal/ans.ts
```diff
@@ -0,0 +1,192 @@
...
+import { AptosConfig } from "../api/aptosConfig";
+import { MoveOption, MoveString, U64 } from "../bcs";
+import { Account, AccountAddress } from "../core";
+import { InputGenerateTransactionOptions, InputSingleSignerTransaction } from "../transactions/types";
+import { HexInput, MoveAddressType, MoveValue } from "../types";
+import { Network } from "../utils/apiEndpoints";
+import { view } from "./general";
+import { generateTransaction } from "./transactionSubmission";
+
+export type ANSName = string | { domainName: string; subdomainName?: string };
```
0xmaayan, line 20: why do accept object as an ans name input?

tests/e2e/ans/publishANSContracts.ts
0xmaayan: ohh I see you run it in beforeAll in the other test, got it.

src/internal/ans.ts
kaw2k: It was my initial implementation, then we shifted to a more concise `string` variant. I thought to keep it around in case the developer has a preference of one way or the other, but can consolidate into just the string if you prefer.

kaw2k: Removed the type and are just using strings now!

src/internal/ans.ts
```diff
@@ -0,0 +1,191 @@
...
+
+const unwrapOption = <T>(option: any): T | undefined => {
+  if (!!option && typeof option === "object" && "vec" in option && Array.isArray(option.vec)) {
+    return option.vec[0];
+  }
+
+  return undefined;
+};
+
+export async function getOwnerAddress({
```
0xmaayan, line 101: format we use
```
getOwnerAddress(args:{aptosConfig: AptosConfig; name: string;}){
 const {aptosConfig, name} = args;
}
```

src/internal/ans.ts
```diff
@@ -0,0 +1,191 @@
...
+    | { policy: "subdomain:follow-domain" }
+    | { policy: "subdomain:independent"; expirationDate: Date };
+  transferable?: boolean;
+  toAddress?: HexInput;
+  targetAddress?: HexInput;
+  options?: InputGenerateTransactionOptions;
+}
+
+export async function registerName({
+  aptosConfig,
```
0xmaayan, line 141: same comment on the input format

src/internal/ans.ts
```diff
@@ -0,0 +1,191 @@
...
+function getRouterAddress(aptosConfig: AptosConfig): MoveAddressType {
+  const address = NetworkToAnsContract[aptosConfig.network];
+  if (!address) throw new Error(`The ANS contract is not deployed to ${aptosConfig.network}`);
+  return address;
+}
+
+const Some = <T>(value: T): MoveValue => ({ vec: [value] } as any);
+const None = (): MoveValue => ({ vec: [] } as any);
+// != here is intentional, we want to check for null and undefined
+// eslint-disable-next-line eqeqeq
```
0xmaayan, line 90: lol whats `eqeqeq`?

kaw2k: Ha, that rule enforces the use of `===` instead of `==`.  `val != undefined` is an old trick to test for only `null` and `undefined` without being true for other falsy values like `""` or `false`

src/internal/ans.ts
```diff
@@ -0,0 +1,191 @@
...
+      function: `${routerAddress}::router::get_owner_addr`,
+      functionArguments: [domainName, Option(subdomainName)],
+    },
+  });
+
+  const owner = unwrapOption<MoveAddressType>(res[0]);
+
+  // We ideally want fromString or fromHexInput but the view function currently
+  // truncates leading 0s
+  return owner ? AccountAddress.fromHexInputRelaxed(owner).toString() : undefined;
```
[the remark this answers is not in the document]
kaw2k: We are getting the address from a `view` function, that is trimming off leading `0`s. The conversion here adds them back in.

tests/e2e/ans/publishANSContracts.ts
gregnazario: Tried to put it under scripts... it was failing with some really messy esm vs cjs problems, so this keeps it compatible with both, but it just does it before the test (and only once)

tests/e2e/ans/ans.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      });
+
+      const bob = Account.generate();
+      await aptos.fundAccount({
+        accountAddress: bob.accountAddress.toString(),
+        amount: 500_000_000,
+      });
+
+      const name = randomString();
+
```
gregnazario: Let's followup in a future PR, but we should test some border conditions on the input strings to make sure they work. e.g. -hello hello- he-llo

src/internal/ans.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  if (second && !isValidANSSegment(second)) {
+    throw new Error(`${second} is not valid. ${VALIDATION_RULES_DESCRIPTION}`);
+  }
+
+  return {
+    domainName: second || first,
+    subdomainName: second ? first : undefined,
+  };
```
gregnazario, line 63: Ah javascript....

src/internal/ans.ts
gregnazario: Interesting, I've been doing !== undefined && != null.

Personally, since most of us aren't pro web developers working on this, I'd stray away from things that can be too clever, but this works fine for me.

src/internal/ans.ts
```diff
@@ -0,0 +1,177 @@
...
+// eslint-disable-next-line eqeqeq
+const Option = <T>(value: T | undefined | null): MoveValue => (value != undefined ? Some(value) : None());
+
+const unwrapOption = <T>(option: any): T | undefined => {
+  if (!!option && typeof option === "object" && "vec" in option && Array.isArray(option.vec)) {
+    return option.vec[0];
+  }
+
+  return undefined;
+};
```
gregnazario, line 99: We'll have to come back and revisit this for view functions as a whole.  I'm hoping this PR https://github.com/aptos-labs/aptos-core/pull/10761 will allow for arguments to be parsed the same as entry functions

gregnazario (approved):

tests/e2e/ans/ans.test.ts
kaw2k: I just added a couple to the `isValidANSName` test

PR #95, opened 2023-10-18
Skip all secp256k1 tests

0xmaayan:
### Description
Skip all secp256k1 tests until next devnet release when single signer is supported


xbtmatt (approved):
Leaving the rest of the stuff in like all the classes, right?

PR #99, opened 2023-10-18
Support publish module API

0xmaayan:
### Description
This adds an API function to publish a move module with the SDK. It is a used function with some of our external partners.
Note: later, when we will have aptos-cli integrated within the SDK, this process would be much more simple!

### Test Plan
pnpm test

### DIFF
--- a/src/internal/transactionSubmission.ts
+++ b/src/internal/transactionSubmission.ts
@@ -6,6 +6,7 @@
  */
 
 import { AptosConfig } from "../api/aptosConfig";
+import { MoveVector } from "../bcs";
 import { postAptosFullNode } from "../client";
 import { Account } from "../core/account";
 import { AccountAuthenticator } from "../transactions/authenticator/account";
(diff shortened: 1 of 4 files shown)


src/internal/transactionSubmission.ts
```diff
@@ -160,3 +167,37 @@ export async function submitTransaction(args: {
   });
   return data;
 }
+
+export async function signAndSubmitTransaction(args: {
+  aptosConfig: AptosConfig;
+  signer: Account;
```
0xjinn, line 173: I like the idea of using `Account` here. Similar concept to where I would like to replace many `address: HexInput` to `address: AccountAddress` :)

0xjinn (approved):

src/internal/transactionSubmission.ts
0xmaayan: I think we use `Account` for all transactions submission api functions

PR #101, opened 2023-10-18
Fix examples

0xmaayan:

### DIFF
--- a/examples/javascript/simple_transfer.js
+++ b/examples/javascript/simple_transfer.js
@@ -32,7 +32,8 @@ const example = async () => {
   console.log("This example will create two accounts (Alice and Bob), fund them, and transfer between them.");
 
   // Setup the client
-  const sdk = new aptos.Aptos();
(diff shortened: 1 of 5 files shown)


examples/javascript/simple_transfer.js
```diff
@@ -32,7 +32,8 @@ const example = async () => {
   console.log("This example will create two accounts (Alice and Bob), fund them, and transfer between them.");
 
   // Setup the client
-  const sdk = new aptos.Aptos();
+  const config = new aptos.AptosConfig({ network: aptos.Network.DEVNET });
```
0xjinn, line 35: I don't think this is taking in the obj as param anymore? And default is devnet already?

0xmaayan: It does and always is. This is a configuration object.
Yes devnet is default but i think it helps to show the usage of the config option

0xjinn: ohh I see, so its taking in `AptosSettings` as an object.

0xjinn (approved):

PR #102, opened 2023-10-19
[BCS classes] Added helper factory methods for Vectors and Options to instantiate Objects/AccountAddresses, + documentation

xbtmatt:
Adding factory methods for MoveObjects and MoveVectors with unit tests & documentation.

**EDIT:** removed the restriction on inputs (requiring vector/option T types to be `EntryFunctionArgumentTypes` and just back to `Serializable`. Can consider it later (it was for ABI purposes)

### Test Plan
```typescript
pnpm jest
```


0xmaayan:
why do we do that?

PR #104, opened 2023-10-19
[typeTag] Add new type tag parser for struct tags

gregnazario:
### Description
The typetag parser will need to be adjusted for generics and other special cases in an upcoming pull request.

This adds a much simpler parser and has comments for every choice made.

### Test Plan
I've put a pretty comprehensive set of tests, including some that are disabled that aren't properly handled in either parser (it can probably be updated in the new one).

Additionally, replaced structTag parsing with this implementation for now.

### DIFF
--- a/src/transactions/typeTag/typeTagParser.ts
+++ b/src/transactions/typeTag/typeTagParser.ts
@@ -1,15 +1,11 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-/* eslint-disable @typescript-eslint/no-unused-vars */
-/* eslint-disable class-methods-use-this */
-/* eslint-disable max-classes-per-file */
 import {
   StructTag,
   TypeTag,
   TypeTagAddress,
   TypeTagBool,
-  TypeTagParserError,
   TypeTagStruct,
   TypeTagU128,
   TypeTagU16,
@@ -19,7 +15,7 @@ import {
   TypeTagU8,
   TypeTagVector,
 } from "./typeTag";
-import { AccountAddress } from "../../core";
+import { AccountAddress, ParsingError } from "../../core";
 import { Identifier } from "../instances";
 
 function isValidIdentifier(str: string) {
@@ -43,6 +39,18 @@ function consumeWhitespace(tagStr: string, pos: number) {
   return i;
 }
 
+type TypeTagState = {
+  savedExpectedTypes: number;
+  savedStr: string;
+  savedTypes: Array<TypeTag>;
+};
+
+export class TypeTagParserError extends ParsingError<string> {
+  constructor(invalidReason: string, typeTagStr: string) {
+    super(`Failed to parse typeTag '${typeTagStr}'`, invalidReason);
+  }
+}
+
 /**
  * All types are made of a few parts they're either:
  * 1. A simple type e.g. u8
@@ -56,16 +64,16 @@ function consumeWhitespace(tagStr: string, pos: number) {
  * 4. Generics for types in ABIs are filled in with placeholders e.g T1, T2, T3
  */
 export function parseTypeTag(typeStr: string) {
-  const saved: Array<{ savedExpectedTypes: number; savedStr: string; savedTypes: Array<TypeTag> }> = [];
+  const saved: Array<TypeTagState> = [];
   // This represents the internal types for a type tag e.g. '0x1::coin::Coin<innerTypes>'
   let innerTypes: Array<TypeTag> = [];
   // This represents the current parsed types in a comma list e.g. 'u8, u8'
   let curTypes: Array<TypeTag> = [];
   // This represents the current character index
-  let cur = 0;
+  let cur: number = 0;
   // This represents the current working string as a type or struct name
-  let currentStr = "";
-  let expectedTypes = 1;
+  let currentStr: string = "";
+  let expectedTypes: number = 1;
 
   // Iterate through each character, and handle the border conditions
   while (cur < typeStr.length) {
@@ -93,12 +101,12 @@ export function parseTypeTag(typeStr: string) {
       // Pop off stack outer type, if there's nothing left, there were too many '>'
       const savedPop = saved.pop();
       if (savedPop === undefined) {
-        throw new TypeTagParserError("Failed to parse typeTag, unexpected '>'");
+        throw new TypeTagParserError(typeStr, "Unexpected '>'");
       }
 
       // If the expected types don't match the number of commas, then we also fail
       if (expectedTypes !== curTypes.length) {
-        throw new TypeTagParserError("Failed to parse typeTag, type argument count mismatch");
+        throw new TypeTagParserError(typeStr, "Type argument count mismatch");
       }
 
       // Add in the new created type, shifting the current types to the inner types
@@ -120,7 +128,7 @@ export function parseTypeTag(typeStr: string) {
         expectedTypes += 1;
       }
     } else if (isValidWhitespaceCharacter(char)) {
-      // This means we should save what we have and everything else should skip until the next ,
+      // This means we should save what we have and everything else should skip until the next
       let parsedTypeTag = false;
       if (currentStr.length !== 0) {
         const newType = parseTypeTagInner(currentStr, innerTypes);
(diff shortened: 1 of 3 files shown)


src/transactions/typeTag/typeTagParser.ts
```diff
@@ -0,0 +1,246 @@
...
+
+/**
+ * All types are made of a few parts they're either:
+ * 1. A simple type e.g. u8
+ * 2. A standalone struct e.g. 0x1::account::Account
+ * 3. A nested struct e.g. 0x1::coin::Coin<0x1234::coin::MyCoin>
+ *
+ * There are a few more special cases that need to be handled, however.
+ * 1. Multiple generics e.g 0x1::pair::Pair<u8, u16>
+ * 2. Spacing in the generics e.g. 0x1::pair::Pair< u8 , u16>
```
xbtmatt, line 54: Could you normalize all inputs by removing all whitespace and then re-adding it only after commas?

gregnazario: Yeah, also normalization by removing spaces and then reparsing is dangerous, if you jam two words together when you forget to have commas.  (not like anyone should be naming their types like that)

src/transactions/typeTag/typeTagParser.ts
```diff
@@ -0,0 +1,246 @@
...
+ * 3. Nested generics of different depths e.g. 0x1::pair::Pair<0x1::coin::Coin<0x1234::coin::MyCoin>, u8>
+ * 4. Generics for types in ABIs are filled in with placeholders e.g T1, T2, T3
+ */
+export function parseTypeTag(typeStr: string) {
+  const saved: Array<{ savedExpectedTypes: number; savedStr: string; savedTypes: Array<TypeTag> }> = [];
```
0xmaayan, line 59: probably a better name
```suggestion
 type Saved = {
  savedExpectedTypes: number; 
  savedStr: string; 
  savedTypes: Array<TypeTag>
 }
  const saved: Array<Saved> = [];
```

src/transactions/typeTag/typeTagParser.ts
```diff
@@ -0,0 +1,246 @@
...
+  // This represents the internal types for a type tag e.g. '0x1::coin::Coin<innerTypes>'
+  let innerTypes: Array<TypeTag> = [];
+  // This represents the current parsed types in a comma list e.g. 'u8, u8'
+  let curTypes: Array<TypeTag> = [];
+  // This represents the current character index
+  let cur = 0;
```
0xmaayan, line 65: can we type each?
```
let cur: number = 0;
let currentStr: string = "";
let expectedTypes: number = 1;
```

src/transactions/typeTag/typeTagParser.ts
```diff
@@ -0,0 +1,246 @@
...
+      // Process last type, if there is no type string, then don't parse it
+      if (currentStr !== "") {
+        const newType = parseTypeTagInner(currentStr, innerTypes);
+        curTypes.push(newType);
+      }
+
+      // Pop off stack outer type, if there's nothing left, there were too many '>'
+      const savedPop = saved.pop();
+      if (savedPop === undefined) {
+        throw new TypeTagParserError("Failed to parse typeTag, unexpected '>'");
```
0xmaayan, line 96: can we log the typeTag?

src/transactions/typeTag/typeTagParser.ts
```diff
@@ -0,0 +1,246 @@
+// Copyright © Aptos Foundation
```
0xmaayan, line 1: I would rename the file to just `parser`, it is already under `typeTag` folder

src/transactions/typeTag/typeTagParser.ts
```diff
@@ -0,0 +1,246 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
0xmaayan, line 4: why?

src/transactions/typeTag/typeTagParser.ts
```diff
@@ -0,0 +1,246 @@
...
+/* eslint-disable class-methods-use-this */
```
0xmaayan, line 5: make them static?

src/transactions/typeTag/typeTagParser.ts
```diff
@@ -0,0 +1,246 @@
...
+    } else if (char === ">") {
+      // Process last type, if there is no type string, then don't parse it
+      if (currentStr !== "") {
+        const newType = parseTypeTagInner(currentStr, innerTypes);
+        curTypes.push(newType);
+      }
+
+      // Pop off stack outer type, if there's nothing left, there were too many '>'
+      const savedPop = saved.pop();
+      if (savedPop === undefined) {
```
0xmaayan, line 95: do we also need to check for `null`?

src/transactions/typeTag/typeTagParser.ts
```diff
@@ -0,0 +1,246 @@
...
+        const newType = parseTypeTagInner(currentStr, innerTypes);
+
+        // parse type tag and push it on the types
+        innerTypes = [];
+        curTypes.push(newType);
+        currentStr = "";
+        expectedTypes += 1;
+      }
+    } else if (isValidWhitespaceCharacter(char)) {
+      // This means we should save what we have and everything else should skip until the next ,
```
0xmaayan, line 123: nit
```suggestion
      // This means we should save what we have and everything else should skip until the next
```

src/transactions/typeTag/typeTagParser.ts
```diff
@@ -0,0 +1,246 @@
...
+        curTypes.push(newType);
+        currentStr = "";
+        parsedTypeTag = true;
+      }
+
+      // Skip ahead on any more whitespace
+      cur = consumeWhitespace(typeStr, cur);
+
+      // The next space MUST be a comma, or a closing > if there was something parsed before
+      // e.g. `u8 u8` is invalid but `u8, u8` is valid
```
0xmaayan, line 139: nice

src/transactions/typeTag/typeTagParser.ts
```diff
@@ -0,0 +1,246 @@
...
+/**
+ * Parses a type tag with internal types associated
+ * @param str
+ * @param types
+ */
+function parseTypeTagInner(str: string, types: Array<TypeTag>): TypeTag {
+  switch (str) {
+    case "bool":
+      if (types.length > 0) {
+        throw new TypeTagParserError("Invalid typeTag, bool should not have a type argument");
```
0xmaayan, line 177: not sure that I understand the error message..?

tests/unit/typeTagParser.test.ts
```diff
@@ -0,0 +1,207 @@
...
+const MODULE_NAME = new Identifier("tag");
+const STRUCT_NAME = new Identifier("Tag");
+
+const structTagType = (typeTags?: Array<TypeTag>) =>
+  new TypeTagStruct(new StructTag(AccountAddress.ONE, MODULE_NAME, STRUCT_NAME, typeTags ?? []));
+
+describe("TypeTagParser", () => {
+  test("invalid types", () => {
+    // Missing 8
+    expect(() => parseTypeTag("8")).toThrow();
```
0xmaayan, line 34: can also add the expected message it throws to the test
```
toThrow("Missing 8")
```

0xmaayan (approved):
Looks great! left some comments

src/transactions/typeTag/typeTagParser.ts
gregnazario: Maybe we make a better name but basically:

This is not a type
```
bool<u8>
```
This is the only type that's a bool
```
bool
```

src/transactions/typeTag/typeTagParser.ts
gregnazario: `pop` returns `undefined` or `value` as far as I could see.

src/transactions/typeTag/typeTagParser.ts
gregnazario: I should probably remove these, I believe this was copied in

src/transactions/typeTag/typeTagParser.ts
gregnazario: sure

tests/unit/typeTagParser.test.ts
gregnazario: The messages need to be standardized a bit (which is why I didn't add any of them).

Will take a look at adding them.

PR #105, opened 2023-10-20
[Rename] Change `arguments` in the transaction builder to `functionArguments`, change references in docs

xbtmatt:
### Description
Because `arguments` is a reserved keyword in typescript, when you try to use the spread operator with it in `strict` mode, it doesn't let you.

For example:
```typescript
const transactionBuilder = async(args: {
    aptos: Aptos,
    sender: Account,
    moduleFunction: string,
    arguments: Array<EntryFunctionArgumentTypes>,
    typeArguments?: TypeTag[],
}): Promise<UserTransactionResponse> => {
    const { aptos, sender, moduleFunction, arguments, typeArguments } = args;
    // ..
```

The compiler won't let you destructure `args` because `arguments` is reserved. You could of course do `arguments: functionArguments`, but perhaps we should just avoid the reserved keyword entirely to avoid issues

I've also changed each instance of `type_arguments` in the docs to `typeArguments`

I also made sure not to change `arguments` to `functionArguments` in API calls where the JSON structure is required

### Test Plan
```shelll
pnpm jest
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to the Aptos TypeScript SDK will be captured in this file. This changelog is written by hand for now. It adheres to the format set out by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Unreleased
-- Changed all instances of `arguments` to `functionArguments` to avoid the reserved keyword in `strict` mode.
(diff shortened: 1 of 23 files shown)


gregnazario (commented):
Can we add a CHANGELOG?

0xjinn (approved):
LGTM

PR #106, opened 2023-10-20
Support custom client provider

0xmaayan:
### Description
This PR adds the ability for a user to specify and provide a custom client. It can be configured as 
```
const config = new AptosConfig({client: { provider: myCustomClient }})
```

`myCustomClient` should be a function with the signature
```
async function myCustomClient<Req, Res>(requestOptions: ClientRequest<Req>): Promise<ClientResponse<Res>>
```

Also, added an example on how one can configure a custom client with both built-in `fetch` api and a custom npm package.

Note:
We will probably want to unify `clientConfig` and `client` on `AptosConfig` to something like
```
export interface Client {
  provider<Req, Res>(requestOptions: ClientRequest<Req>): Promise<ClientResponse<Res>>;
  config: ClientConfig
}
export class AptosConfig {
readonly client: Client;
}
```
But that needs some more work and wanted to keep the scope of this PR light.

### Test Plan
pnpm test


src/client/core.ts
```diff
@@ -88,7 +83,7 @@ export async function aptosRequest<Req, Res>(
       throw new AptosApiError(
         options,
         result,
-        response.data.errors[0].message ?? `Unhandled Error ${response.status} : ${response.statusText}`,
+        (response.data as any).errors[0].message ?? `Unhandled Error ${response.status} : ${response.statusText}`,
```
gregnazario, line 86: Interesting, why `any`?

src/types/index.ts
```diff
@@ -117,7 +117,7 @@ export type AnyNumber = number | bigint;
  * behavior and interaction with the Aptos network
  */
 export type AptosSettings = {
-  readonly network: Network;
+  readonly network?: Network;
```
gregnazario, line 120: Why is network optional?

examples/typescript/custom_client.ts
```diff
@@ -0,0 +1,95 @@
...
+    config: response,
+    request,
+  };
+}
+
+const example = async () => {
+  console.log("This example demostrate how one can config for a custom client ot be used by the SDK");
+
+  async function withSuperagentClient() {
+    const config = new AptosConfig({ client: { provider: superagnetCustomClient } });
```
gregnazario, line 71: *superagent*

examples/typescript/custom_client.ts
```diff
@@ -0,0 +1,95 @@
+/**
+ * Example to demostrate how one can config the SDK to use a custom client.
```
gregnazario, line 2: demonstrate

gregnazario (commented):
So, this is pretty nice, let's you drop in any client that you want, but then you still need the dependency for the original client that we have?

src/types/index.ts
0xmaayan: good catch. So technically `network` shouldnt be required to pass in when initializing `AptosConfig` as `Aptos` resolves to a default network - Devnet. It was working fine by now because the `args` `AptosConfig` accepts is optional.

I guess discussion on whether we should default to a network makes sense or not, but if `network` is required, when want to support a custom client, need to
```
const config = new AptosConfig({ network:Network.DEVNET, client: { provider: customClient } });
```

0xmaayan:
> So, this is pretty nice, let's you drop in any client that you want, but then you still need the dependency for the original client that we have?

yes... I guess in another world, we could have avoid it by having the user specify if they use a custom client or the one SDK supports when they `pnpm i sdk` and then the sdk builds package.json on the fly and decides whether to include `aptos-client` dependency or not. It makes it a bit tricky

src/client/core.ts
0xmaayan: before, `request` returned `Promise<any>` and therefore ts will resolve to any type and you can pretty much do whatever you want. now, we return a `ClientResponse<Res>` type where the `data` in `ClientResponse` is `Res`. Indexer is a bit weird in a way that errors are part of the data key returned by the api, so since we dont really want to do `type ClientResponse = {.... data:Res & {errors?:any} } }` we just convert it here to `any`

0xmaayan: btw, we also do it in the `if` block

src/client/core.ts
```diff
@@ -88,7 +83,7 @@ export async function aptosRequest<Req, Res>(
       throw new AptosApiError(
         options,
         result,
-        response.data.errors[0].message ?? `Unhandled Error ${response.status} : ${response.statusText}`,
+        (result.data as any).errors[0].message ?? `Unhandled Error ${response.status} : ${response.statusText}`,
```
gregnazario, line 86: This may be a good time to type this output or, type guard it so result.data is any so we don't keep redoing it over and over below.

e.g.
```
const data = result.data as any;

if (data.errors) {
  throw new data.errors[0].message ...
}
```
etc.

Either that, or type it if possible, but I guess it's kinda the any randomness.

Interesting that it doesn't need `?` before each sub item.

gregnazario (approved):

src/types/index.ts
xbtmatt: I definitely do like being able to do `new AptosConfig();`, but maybe it's not explicit enough...

PR #108, opened 2023-10-21
add troubleshoot in readme on import error

0xaptosj:
### Description

We should document this import error due to tsconfig, im putting it in readme for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,16 @@ To run the SDK tests, simply run from the root of this repository:
 pnpm test
 ```
 
+## Troubleshooting
+
+If you see import error when you do this
+
+```javascript
+import { Aptos, AptosConfig, Network } from "@aptos-labs/ts-sdk";
+```
+
+It could be your `tsconfig.json` is not incompatible, make sure your `moduleResolution` is set to `node` instead of `bundler`.
+
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


0xaptosj:
i don't have merge access, if it looks good please merge for me

xbtmatt (commented):
Hm, I think we might want to be more general about how we approach troubleshooting. This is more like a single item in a troubleshooting faq

Maybe we should just link them to a full working example template

README.md
```diff
@@ -131,6 +131,16 @@ To run the SDK tests, simply run from the root of this repository:
 pnpm test
 ```
 
+## Troubleshooting
+
+If you see import error when you do this
+
+```javascript
```
gregnazario, line 138: Nit I think this should be typescript since it only applies to typescript

gregnazario (approved):

gregnazario:
I agree, we should probably make a docs page somewhere for troubleshooting.

We have a few flushed out examples with TypeScript already, which a user could copy paste.  We'd likely just want to focus on the onboarding experience for create aptos dapp

README.md
0xaptosj: fixed

xbtmatt (approved):

PR #109, opened 2023-10-23
[account] Add ability for custom typed resource calls

gregnazario:
### Description
Add ability for direct typed resource calls from the API without having to do conversion.

This should let you read Move resources directly, without having to have your own conversion code.  This also should allow for general type safety, keeping in mind that the underlying resource could possibly not match if it's not correct.

I'm debating how useful this is without something that generates types ahead of time from the ABI, but it also should be pretty easy to generate those.

### Test Plan
I added a test that does a direct comparison of the output objects of `getAccountResource` and the typed one, and they are the same, though now there is proper typing to handle it

### DIFF
--- a/src/internal/account.ts
+++ b/src/internal/account.ts
@@ -155,12 +155,12 @@ export async function getResources(args: {
   });
 }
 
-export async function getResource(args: {
+export async function getResource<T extends {}>(args: {
   aptosConfig: AptosConfig;
   accountAddress: HexInput;
   resourceType: MoveResourceType;
   options?: LedgerVersion;
-}): Promise<MoveResource> {
+}): Promise<T> {
   const { aptosConfig, accountAddress, resourceType, options } = args;
   const { data } = await getAptosFullNode<{}, MoveResource>({
     aptosConfig,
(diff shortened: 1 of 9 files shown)


0xmaayan (changes requested):

src/internal/account.ts
```diff
@@ -171,6 +171,22 @@ export async function getResource(args: {
   return data;
 }
 
+export async function getResourceTyped<T>(args: {
```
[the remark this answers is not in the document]
gregnazario: Can you have optional types / overloading?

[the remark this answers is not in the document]
0xmaayan: no. but we can just support the type one...?

src/internal/account.ts
```diff
@@ -155,20 +155,20 @@ export async function getResources(args: {
   });
 }
 
-export async function getResource(args: {
+export async function getResource<T extends {}>(args: {
```
0xmaayan, line 158: wonder if we can `T = {}` to default it to `{}` and then dont have to provide any type

src/api/abi.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+export class ABI {}
```
0xmaayan, line 4: do we use it?

0xmaayan (approved):
Very nice! just a comment/question on default type to `{}`

src/api/abi.ts
gregnazario: Oh, this shouldn't be here

src/internal/account.ts
gregnazario: I went with `T extends {} = any`.  This allows typing to provided by the user, but still allows blind usage via any.

PR #110, opened 2023-10-24
add missing indexerVersionCheck check var to waitForTransaction in api level

0xjinn:


0xmaayan (approved):
might worth having it as a custom type
```
export type WaitForTransactionOptions = {
 timeoutSecs?: number
 checkSuccess?: boolean
 indexerVersionCheck?: boolean
}
```

gregnazario (commented):
Are there any other functions that need it propagated through?  Faucet?  Others?

PR #112, opened 2023-10-24
Add Swap example

0xjinn:
### Description

First version of the `swap` dapp or example. This version relies on the user to compile AND publish the `facoin` and `swap` package. 

### Test Plan

```
➜  typescript git:(main) ✗ pnpm swap 0x7f3d84369e337f51bf738ceac7bc661f6ca190989d92405495d716db40b1033f 0x88fef4956bd9155c217671ad14adb0ede4b19971a0cef82e4b94595575b96413

> ts-test@1.0.0 swap /Users/jinhou/Projects/aptos-ts-sdk/examples/typescript
> ts-node swap.ts "0x7f3d84369e337f51bf738ceac7bc661f6ca190989d92405495d716db40b1033f" "0x88fef4956bd9155c217671ad14adb0ede4b19971a0cef82e4b94595575b96413"

This example will create a main user account called 'Admin', it will be used to deploy Liquidity pool and two new fungible assets. 
After creating the Dog and Cat coin, and the liquidity pool, it will swap one token for another. 
Note: This example requires you to have the 'swap' module published before running. 
If you haven't published the 'swap' module, please publish the package using 
'aptos move create-resource-account-and-publish-package --seed 0 --address-name=swap --named-addresses deployer=[admin] --profile [admin]' first. 
[admin] is the account profile you will be using for this example. 

====== Account info ======

Admin's address is: 0xc0b7fa8173bb62556c1cd138c2d84fdabe86f5874ec6a3f36b3177c1c9b857e6
Admin's private key is: 0x88fef4956bd9155c217671ad14adb0ede4b19971a0cef82e4b94595575b96413
Swap address is: 0x7f3d84369e337f51bf738ceac7bc661f6ca190989d92405495d716db40b1033f

====== Create Fungible Asset -> (Dog and Cat coin) ======

Follow the steps to publish the Dog and Cat Coin module with Admin's address, and press enter. 
1. cd to /aptos-ts-sdk/examples/typescript/facoin folder 
2. run 'aptos move publish --named-address FACoin=[admin_address] --private-key=[private_key]
Fungible Asset:  [
  {
    icon_uri: 'http://example.com/favicon.ico',
    project_uri: 'http://example.com',
    supply_aggregator_table_handle_v1: null,
    supply_aggregator_table_key_v1: null,
    creator_address: '0xc0b7fa8173bb62556c1cd138c2d84fdabe86f5874ec6a3f36b3177c1c9b857e6',
    asset_type: '0x17c94be9519bcc24afc779041423b14b59c3af91f0a084fa9004a7e0a1434ef6',
    decimals: 8,
    last_transaction_timestamp: '2023-10-24T09:22:19',
    last_transaction_version: 6408698,
    name: 'CAT Coin',
    symbol: 'CAT Coin',
    token_standard: 'v2'
  },
  {
    icon_uri: 'http://example.com/favicon.ico',
    project_uri: 'http://example.com',
    supply_aggregator_table_handle_v1: null,
    supply_aggregator_table_key_v1: null,
    creator_address: '0xc0b7fa8173bb62556c1cd138c2d84fdabe86f5874ec6a3f36b3177c1c9b857e6',
    asset_type: '0x41d4a401c2e3ccab4576b138ecfdb83add0cd7d135e00a65e994433e1a2d1165',
    decimals: 8,
    last_transaction_timestamp: '2023-10-24T09:22:19',
    last_transaction_version: 6408698,
    name: 'DOG Coin',
    symbol: 'DOG Coin',
    token_standard: 'v2'
  }
]
Cat FACoin asset type: 0x17c94be9519bcc24afc779041423b14b59c3af91f0a084fa9004a7e0a1434ef6
Dog FACoin asset type: 0x41d4a401c2e3ccab4576b138ecfdb83add0cd7d135e00a65e994433e1a2d1165

====== Mint Dog and Cat Coin ======

minting 20_000_000 Dog coin...
Minting dog coin successful. -  0xd1e410127aa92d2f5f144b7c9aea52612473ea1c337a7a888b40f854768aea49
minting 30_000_000 Cat coin...
Minting cat coin successful. -  0xaf8c27a3ec8ed97bbc6a2b4373980022dfb0e67ef907ea33ac3083dc366e43c9

====== Current Balance ======

Admin's Dog coin balance: 20000000.
Admin's Cat coin balance: 30000000.

====== Create Liquidity Pool ======

initializing Lquidity Pool......
Init LP Pool success. -  0x66c43d41f8acc6a51e1151729fd776b614926a7443bd612e1b91e4818ac473af
Creating liquidity pool......
Creating liquidity pool successful. -  0x7399f7b15fb165dacb65de4b88307614d34faa0f17f4fec596640aaa10be4f51
Getting optimal LP amount......
Optimal LP amount:  [ '200000', '300000', '243948' ]
Adding liquidity......
Add liquidity succeed. -  0xf65c1c01d74c62aba7ed7dca397f2fbe452820a6bffd419d76bba8d646581257
Done.

====== Swap 100 Dog coins for Cat coins ======

Swaping 100 Dog coin to Cat coin......
Swap succeed. -  0x0f6683cc8e37d1ff929ecb2a0a51e5b153c3def9a05410765edec33c97522979
Swap finished.

====== Current Balance ======

Admin's Dog coin balance: 19799900.
Admin's Cat coin balance: 29700149.
```


examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+const getOptimalLpAmount = async (
+  aptos: Aptos,
+  swap: AccountAddress,
+  token1Addr: HexInput,
+  token2Addr: HexInput,
+): Promise<void> => {
+  const payload: ViewRequestData = {
+    function: `0x${swap.toStringWithoutPrefix()}::router::optimal_liquidity_amounts`,
+    arguments: [
+      `0x${Hex.fromHexInput(token1Addr).toStringWithoutPrefix()}`,
```
movekevin, line 41: Does just token1Addr.toString() here?

examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+      `0x${Hex.fromHexInput(token2Addr).toStringWithoutPrefix()}`,
+      false,
+      "200000",
```
movekevin, line 44: Do numbers work here? This works in v1

examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+  aptos: Aptos,
+  swap: AccountAddress,
+  deployer: Account,
+  token1Addr: HexInput,
+  token2Addr: HexInput,
+): Promise<string> => {
+  const rawTxn = await aptos.generateTransaction({
+    sender: deployer.accountAddress.toString(),
+    data: {
+      function: `0x${swap.toStringLongWithoutPrefix()}::router::add_liquidity_entry`,
```
movekevin, line 64: I think the 0 prefix should still work? toStringLongWithoutPrefix is a bit long instead of toString

examples/typescript/facoin/Move.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "facoin"
+version = "1.0.0"
+authors = []
+
+[addresses]
+aptos_framework = "0x1"
```
gregnazario, line 7: This is not needed FYI, it is inherited from AptosFramework.

examples/typescript/facoin/publish.json
```diff
@@ -0,0 +1,17 @@
...
...054a1fded03f79f397343287df06e7436ba521b22fadc35d85f0a50d7fd5ddf7f1738c0f0bbcfcd6fd1eaac34ce3768903b8d8dd99501915655adf44b917ec22082b08cf8bba979d380cf673e6e67e3b6a9d697d173b8f8c23891f6715dcbb8eb3cb50c098801375ac319ab86b2129eb60d110602de5e2ab5544b9066dc0663e5866b1740ee6faf25c84de9a3b50b752bf2575a2a53e388bdbfc66eafed55971791991fc173927976864c25a2a252bfcca7396194e451fbf63beac0eb763d2bf914977bebfc3a68dc086fe4962aa3fefb0380a222ff0cb58fdcaddd3df69aeb0b2c48a758d20c4a3b5fb6b1764241b46f9e74ddb751a72b483737da79735813c9f5a2180b08652bf864e7a2c019e9bcd2999ba9138e435331daddd909556e28109e0543990c8061489a68c594c505edef6ba331badfcde67e3732e4ee277841a53ba887fc88d1838ebae6f23a1ca95ea875e0fe454c021ddd88c4e861f77175a543fb8498dcfa8ca0fcd44d427d471b0ec8cf3bfd423a94642e742d3b5dd6dd79472ee0514ae9c70d4c4b28ff1186bc57a9fee51056ef5c3cd8d9d9d98d51c933ac9fde614cff141626ed6a4dd990bb2656daa566fc7b4969fbfaa65dbb3655afb6aeb6fe010ca47f45191e000000000300000000000000000000000000000000000000000000000000000000000000010e4170746f734672616d65776f726b00000000000000000000000000000000000000000000000000000000000000010b4170746f735374646c696200000000000000000000000000000000000000000000000000000000000000010a4d6f76655374646c696200"
```
gregnazario, line 7: Interesting, what is this file?  A format file for deploying packages?

Perhaps we want to make a github issue to autogenerate this from the CLI for easier deployment via TS and others.

examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+const readline = require("readline").createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+const getOptimalLpAmount = async (
+  aptos: Aptos,
+  swap: AccountAddress,
+  token1Addr: HexInput,
+  token2Addr: HexInput,
```
gregnazario, line 36: We should probably take a look into making a proper AccountAddress input (AccountAddress | HexString), since I suspect people won't be passing in byte arrays for the account address.  Since the swap is an `AccountAddress` having this mixture of types may be a little strange.

examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+): Promise<void> => {
+  const payload: ViewRequestData = {
+    function: `0x${swap.toStringWithoutPrefix()}::router::optimal_liquidity_amounts`,
```
gregnazario, line 39: Good callout here for improvements to the SDK to remove the 0x in the pattern

examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+    arguments: [
+      `0x${Hex.fromHexInput(token1Addr).toStringWithoutPrefix()}`,
+      `0x${Hex.fromHexInput(token2Addr).toStringWithoutPrefix()}`,
+      false,
+      "200000",
+      "300000",
+      "200",
+      "300",
```
gregnazario, line 47: Are these U64s?

I suspect one of the things we should probably fix on the API side for View functions is the encoding, and allow BCS encoding of inputs, and use the ABI for translation.

examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+    ],
+  };
+  const result: any = await aptos.view({ payload });
```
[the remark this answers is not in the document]
gregnazario: We should be able to drop the `any` here.

examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+const getAssetType = async (aptos: Aptos, owner: Account): Promise<any> => {
+  const data = await aptos.getFungibleAssetMetadata({
+    options: {
+      where: {
+        creator_address: { _eq: owner.accountAddress.toString() },
+      },
+    },
+  });
+
+  if (data.length !== 2) throw new Error("Expected two Fungible Asset.");
```
gregnazario, line 121:
```suggestion
  if (data.length !== 2) throw new Error("Expected two Fungible Assets.");
```

Did you also want to verify the fungible assets are different?

I don't think a liquidity pool with the same exact assets make any sense?

examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+    cat: data[0].asset_type,
+    dog: data[1].asset_type,
+  };
+};
+
+const getFaBalance = async (aptos: Aptos, owner: Account, assetType: string): Promise<number> => {
+  const data = await aptos.getCurrentFungibleAssetBalances({
+    options: {
+      where: {
+        owner_address: { _eq: owner.accountAddress.toString() },
```
gregnazario, line 134: We should note here, I don't think that addresses are stored in shorter mode, making sure that the account addresses match?

examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  const aptos = new Aptos();
+  // Create three accounts
+  const admin = Account.fromPrivateKey(new Ed25519PrivateKey(process.argv[3]));
+  const swapAddress = AccountAddress.fromHexInput(process.argv[2]);
+
+  console.log("====== Account info ======\n");
+  console.log(`Admin's address is: ${admin.accountAddress.toString()}`);
+  console.log(`Admin's private key is: ${admin.privateKey.toString()}`);
```
gregnazario, line 229: Note, we should not print out private keys in examples.  Being as we should show proper best practices.
 
Instead, you can show the public key, or just the address.

examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+  // Prerequisite check
+  if (process.argv.length !== 4) {
+    console.log("Required usage: pnpm swap [swap_address] [user_private_key]");
+    process.exit(1);
+  }
+
+  const aptos = new Aptos();
+  // Create three accounts
+  const admin = Account.fromPrivateKey(new Ed25519PrivateKey(process.argv[3]));
+  const swapAddress = AccountAddress.fromHexInput(process.argv[2]);
```
gregnazario, line 225: Nit: From a usability side, I would probably use a file to keep the private key.

examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+
+  console.log("====== Account info ======\n");
+  console.log(`Admin's address is: ${admin.accountAddress.toString()}`);
+  console.log(`Admin's private key is: ${admin.privateKey.toString()}`);
+  console.log(`Swap address is: ${swapAddress.toString()}`);
+  // Fund Admin account
+  await aptos.fundAccount({ accountAddress: admin.accountAddress.toString(), amount: 100_000_000 });
```
gregnazario, line 232: This is a good example of how we should be taking in including `AccountAddress`

gregnazario (commented):
I have some best practices and general use questions, but otherwise looks pretty good

examples/typescript/swap.ts
movekevin: +1. If we can minimize the need to know how all these similar types differ, it'd be great!

examples/typescript/swap.ts
gregnazario: I took a look, it probably doesn't for view functions, it does string to u64 conversion from JSON on the API, and requires it here.

examples/typescript/swap.ts
movekevin: Not sure you mean. There's still a 0x at the start of the format string

examples/typescript/swap.ts
movekevin: +1

examples/typescript/swap.ts
movekevin: Or use .env file. Can be easily loaded with dotenv and then use process.env to extract the values. This is common safe practice in crypto

examples/typescript/swap.ts
gregnazario: There's something we have to fix in the SDK, which adds the protection of some inputs as `0x${string}`, however if you don't type the output of `swap.toStringWithoutPrefix{}` to `0x${string}`, it won't accept it (even if it always outputs `0x12345`)

We tried something fancy, and it made it kinda messy

movekevin (commented):
What would say are the highlights in terms of how v2 is easier or harder to write the same code as v1?

examples/typescript/swap.ts
movekevin: Does the API not accept module addresses without 0x prefix?

gregnazario:
I would also note, that private key you just used for this example (since it's now logged here), you don't use again @0xjinn

examples/typescript/swap.ts
gregnazario: It does, but we're trying to get rid of the variability

examples/typescript/swap.ts
movekevin: Or better we could provide an eq implementation that ignores 0x prefix (similar to equalIgnoreCase)?

movekevin:
This is very insightful so far. Amazing job @0xjinn !

examples/typescript/swap.ts
gregnazario: This is GraphQL, so it's direct to the DB, without having a service that fronts the indexer and does conversions, it's going to have to match exactly

examples/typescript/facoin/sources/fa_coin_cat.move
```diff
@@ -0,0 +1,158 @@
+/// A 2-in-1 module that combines managed_fungible_asset and coin_example into one module that when deployed, the
+/// deployer will be creating a new managed fungible asset with the hardcoded supply config, name, symbol, and decimals.
+/// The address of the asset can be obtained via get_metadata(). As a simple version, it only deals with primary stores.
+module FACoin::cat {
```
0xmaayan, line 4: why are we using a new move module and not an existing one?

examples/typescript/swap.ts
```diff
@@ -0,0 +1,275 @@
...
+
+const getOptimalLpAmount = async (
+  aptos: Aptos,
+  swap: AccountAddress,
+  token1Addr: HexInput,
+  token2Addr: HexInput,
+): Promise<void> => {
+  const payload: ViewRequestData = {
+    function: `0x${swap.toStringWithoutPrefix()}::router::optimal_liquidity_amounts`,
+    arguments: [
```
0xmaayan, line 40: it was renamed to `functionArguments`, make sure you are working against the latest local sdk

examples/typescript/swap.ts
0xmaayan: it is not about what will work or not. it is about the strict type we hold in the sdk. we ask the `function` to be of type `0x${string}::${string}::${string}` but since `swap.toString()` is just a string (tho it does have 0x) typescript will complain that type dont match. 
It is just something we need to change in the sdk

examples/typescript/swap.ts
movekevin: Why not arguments to make it shorter? 😜

0xmaayan: user feedback

examples/typescript/swap.ts
0xjinn: Yep correct. Hence I did created 2 different module (Cat and Dog) to separate out the two different fungible asset. 

Good call here, will check and make sure these two assets are differents.

examples/typescript/swap.ts
0xjinn: Going to remove this line. I was gonna log the private key for other purpose, but not anymore 😄

examples/typescript/facoin/sources/fa_coin_cat.move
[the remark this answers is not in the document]
movekevin: That shouldn't matter :)

examples/typescript/swap.ts
movekevin: The query can be modified to do an OR with the WHERE query right?

0xjinn: I believe the `owner_address` that stored in DB are always prefixed with `0x`

examples/typescript/facoin/sources/fa_coin_cat.move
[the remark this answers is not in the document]
0xjinn: Oh interesting! Though I feel like the user might get confused, and cause some confusion. 

@movekevin Shall we keep it as two separate module?

0xjinn:
Address comments

1. Rebase to main to update `arguments` to `functionArugments`
2. Remove the private key console.log
3. update FA token from HexInput to AccountAddress
4. others.

examples/typescript/swap.ts
xbtmatt: @movekevin it was because `arguments` is a reserved keyword in `strict` mode, so you couldn't easily deconstruct an args struct without having to rename:

```typescript
const { ..., arguments: functionArguments, ... } = args;
// use functionArguments...
``` 

if it leads to issues down the road we can change it back :p

examples/typescript/swap.ts
```diff
@@ -0,0 +1,247 @@
...
+  swap: AccountAddress,
+  deployer: Account,
+  fromToken: AccountAddress,
+  toToken: AccountAddress,
+  amountIn: number,
+  amountOutMin: number,
+  recipient: AccountAddress,
+): Promise<string> => {
+  const rawTxn = await aptos.generateTransaction({
+    sender: deployer.accountAddress.toString(),
```
movekevin, line 73: Is this required even with the signAndSubmitTransaction call on line 79?

0xjinn: Yep this is required. I think this is basically generating raw txn in bytecode, and we need the address to specify where the module gonna live in.

movekevin (approved):

examples/typescript/swap.ts
gregnazario: This is actually fixed now, so we can get rid of the 0x weird hack.

examples/typescript/swap.ts
gregnazario: Fixed now.

examples/typescript/swap.ts
gregnazario: owner_address is stored as a full hex, with a 0x in front.  So, this likely needs to be .toLongString()

gregnazario (approved):
Please fix the long string check

examples/typescript/swap.ts
0xjinn: The `.toString()` here actually coming from Hex.toString() that dport implemented, so it should always contains the `0x` prefix 😄

gregnazario: It doesn't just need the `0x`, it actually needs the full `0's` so you'll need to use `toLongString()`, double check with the indexer folks

0xjinn: ohh interesting. I thought `toString()` is already converting to long string which contains all of the leading zeros if there any. 

Ok, we probably need to implement a `toLongString()` then.

gregnazario: https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/core/accountAddress.ts#L142-L144

0xjinn: yeah I just realized we already have one on `AccountAddress`, I was looking at the wrong class `Hex` lol all fixed now #134

PR #113, opened 2023-10-24
[types] Remove requirement on 0x for input

gregnazario:
### Description
This removes needing to do silliness like:

`0x${HexInput.toStringWithoutPrefix()}...`

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -639,9 +639,9 @@ export type MoveUint32Type = number;
 export type MoveUint64Type = string;
 export type MoveUint128Type = string;
 export type MoveUint256Type = string;
-export type MoveAddressType = `0x${string}`;
-export type MoveObjectType = `0x${string}`;
-export type MoveStructType = `0x${string}::${string}::${string}`;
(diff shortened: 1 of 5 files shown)


src/types/index.ts
```diff
@@ -639,9 +639,9 @@ export type MoveUint32Type = number;
 export type MoveUint64Type = string;
 export type MoveUint128Type = string;
 export type MoveUint256Type = string;
-export type MoveAddressType = `0x${string}`;
-export type MoveObjectType = `0x${string}`;
-export type MoveStructType = `0x${string}::${string}::${string}`;
+export type MoveAddressType = string;
+export type MoveObjectType = string;
+export type MoveStructType = `${string}::${string}::${string}`;
```
0xjinn, line 644: I suspect user (like me) will dig in and check what these types are. 

Shall we add a note here for the user to know, that there should be `0x` prefix?

0xjinn (approved):

src/types/index.ts
gregnazario: We can probably document that on each type

PR #114, opened 2023-10-24
Include client config override and set version for metrics

0xmaayan:
### Description
- Add clientConfig to header client config override
- Set metric version to 2.0.0 to distinguish between sdk v1 and v2

### DIFF
--- a/src/client/post.ts
+++ b/src/client/post.ts
@@ -69,7 +69,7 @@ export async function post<Req, Res>(options: PostRequestOptions): Promise<Aptos
       acceptType: acceptType?.valueOf(),
       params,
       overrides: {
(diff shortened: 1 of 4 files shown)


src/client/post.ts
```diff
@@ -69,7 +69,7 @@ export async function post<Req, Res>(options: PostRequestOptions): Promise<Aptos
       acceptType: acceptType?.valueOf(),
       params,
       overrides: {
-        ...aptosConfig,
+        ...aptosConfig.clientConfig,
```
banool, line 72: wow funky, I'm surprised the type checker was okay with this

banool (approved):

PR #115, opened 2023-10-24
Add new query to fetch specific FA metadata

0xjinn:
### Description

When working with `swap` example, I noticed I have to make a call to fetch a specific metadata quite often. I am assuming users who works with Fungible Asset alot will need work with this query alot.

### Test Plan

pnpm test
pnpm fmt

### DIFF
--- a/src/api/fungibleAsset.ts
+++ b/src/api/fungibleAsset.ts
@@ -5,7 +5,6 @@ import {
   GetCurrentFungibleAssetBalancesResponse,
   GetFungibleAssetActivitiesResponse,
   GetFungibleAssetMetadataResponse,
-  GetFungibleAssetMetadatasResponse,
   PaginationArgs,
 } from "../types";
 import { AptosConfig } from "./aptosConfig";
@@ -43,7 +42,7 @@ export class FungibleAsset {
       pagination?: PaginationArgs;
       where?: FungibleAssetMetadataBoolExp;
     };
-  }): Promise<GetFungibleAssetMetadatasResponse> {
+  }): Promise<GetFungibleAssetMetadataResponse> {
     return getFungibleAssetMetadata({ aptosConfig: this.config, ...args });
   }
 
@@ -52,11 +51,15 @@ export class FungibleAsset {
    *
    * This query returns the fungible asset metadata for a specific fungible asset.
    *
-   * @param assetType The asset type of the fungible asset
-   * @returns getFungibleAssetMetadata A list of fungible asset metadata
+   * @param assetType The asset type of the fungible asset.
+   * e.g
+   * "0x1::aptos_coin::AptosCoin" for Aptos Coin
+   * "0xc2948283c2ce03aafbb294821de7ee684b06116bb378ab614fa2de07a99355a8" - address format if this is fungible asset
+   *
+   * @returns getFungibleAssetMetadata A fungible asset metadata item
    */
   async getFungibleAssetMetadataByAssetType(assetType: string): Promise<GetFungibleAssetMetadataResponse> {
-    const data = await getFungibleAssetMetadata({
+    return getFungibleAssetMetadata({
       aptosConfig: this.config,
       options: {
         where: {
@@ -64,8 +67,6 @@ export class FungibleAsset {
         },
       },
     });
-
-    return data[0];
   }
 
   /**
--- a/src/api/fungibleAsset.ts
+++ b/src/api/fungibleAsset.ts
@@ -58,8 +58,8 @@ export class FungibleAsset {
    *
    * @returns getFungibleAssetMetadata A fungible asset metadata item
    */
-  async getFungibleAssetMetadataByAssetType(assetType: string): Promise<GetFungibleAssetMetadataResponse> {
-    return getFungibleAssetMetadata({
+  async getFungibleAssetMetadataByAssetType(assetType: string): Promise<GetFungibleAssetMetadataResponse[0]> {
+    const data = await getFungibleAssetMetadata({
       aptosConfig: this.config,
       options: {
         where: {
@@ -67,6 +67,8 @@ export class FungibleAsset {
         },
       },
     });
+
+    return data[0];
   }
 
   /**
--- a/src/api/fungibleAsset.ts
+++ b/src/api/fungibleAsset.ts
@@ -5,6 +5,7 @@ import {
   GetCurrentFungibleAssetBalancesResponse,
   GetFungibleAssetActivitiesResponse,
   GetFungibleAssetMetadataResponse,
+  GetFungibleAssetMetadatasResponse,
   PaginationArgs,
 } from "../types";
 import { AptosConfig } from "./aptosConfig";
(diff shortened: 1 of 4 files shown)


src/api/fungibleAsset.ts
```diff
@@ -42,10 +43,31 @@ export class FungibleAsset {
...
 
+  /**
+   * Queries the current specific fungible asset metadata
+   *
+   * This query returns the fungible asset metadata for a specific fungible asset.
+   *
+   * @param assetType The asset type of the fungible asset
+   * @returns getFungibleAssetMetadata A list of fungible asset metadata
+   */
+  async getFungibleAssetMetadataByAssetType(assetType: string): Promise<GetFungibleAssetMetadataResponse> {
```
0xmaayan, line 58: does it return a list? if so, we should support `options.pagination`. 
I would also follow the existing format, and create a `getFungibleAssetMetadataByAssetType` in `internal/fungibleAsset`

0xmaayan, line 58:
```suggestion
  async getFungibleAssetMetadataByAssetType(args:{assetType: string, options?:...}): Promise<GetFungibleAssetMetadataResponse> {
```

src/types/indexer.ts
```diff
@@ -58,7 +58,8 @@ export type GetDelegatedStakingActivitiesResponse = GetDelegatedStakingActivitie
 export type GetCollectionDataResponse = GetCollectionDataQuery["current_collections_v2"][0];
 export type GetTokenDataResponse = GetTokenDataQuery["current_token_datas_v2"][0];
 export type GetProcessorStatusResponse = GetProcessorStatusQuery["processor_status"];
-export type GetFungibleAssetMetadataResponse = GetFungibleAssetMetadataQuery["fungible_asset_metadata"];
+export type GetFungibleAssetMetadatasResponse = GetFungibleAssetMetadataQuery["fungible_asset_metadata"];
+export type GetFungibleAssetMetadataResponse = GetFungibleAssetMetadataQuery["fungible_asset_metadata"][0];
```
gregnazario, line 62: This can be... confusing at best.

Best to have it be `GetFungibleAssetMetadataResponse` and then `Array<GetFungibleAssetMetadataResponse>`

A single character inside a 32 character string is easy to miss

src/api/fungibleAsset.ts
```diff
@@ -42,10 +43,31 @@ export class FungibleAsset {
...
+    const data = await getFungibleAssetMetadata({
+      aptosConfig: this.config,
+      options: {
+        where: {
+          asset_type: { _eq: assetType },
+        },
+      },
+    });
+
+    return data[0];
```
gregnazario, line 68: is there more than one?  Then let's return all and leave to the user to handle it, or provide a assertion or error that handles when there's more than one.

src/api/fungibleAsset.ts
gregnazario, line 58: Let's document AssetType on what would it be:
1. Currency code
2. Struct type

0xjinn: So if we query by `asset type`, it should be 1 single item of the metadata. Noticed I added a new type `GetFungibleAssetMetadataResponse` for this case, and updated the list one as `GetFungibleAssetMetadatasResponse`. (The letter "s" might be hard to see haha)

And I was thinking on how to handle this. Since we already have a general query that can take an `where` condition in /internal. I think we can take advantage of that, and have many of the /api query depends on that one.
For example:

```
/internal
getFungibleAssetMetadata

/api
getFungibleAssetMetadata - same as above
getFungibleAssetMetadataByAssetType - calling the above with a `where` clause (_eq: asset_type)
getFungibleAssetMetadataByOwner - calling the above with a `where` clause (_eq: owner_address)
```

Any thoughts?

tests/e2e/api/fungibleAsset.test.ts
```diff
@@ -26,6 +26,11 @@ describe("FungibleAsset", () => {
     expect(data[0].asset_type).toEqual(APTOS_COIN);
   });
 
+  test("it should fetch a specific fungible asset metadata", async () => {
```
0xmaayan, line 29: so not sure how useful this test is, we create only one asset. maybe worth trying to get by an asset that doesnt exist and then make sure it returns nothing

0xjinn: Good call!

src/api/fungibleAsset.ts
0xmaayan: then lets update the function comment `etFungibleAssetMetadata A list of fungible asset metadata`.

And yes we should definitely take advantage on the existing query. I still think it makes sense to have on the user facing api function `options` that holds `where` condition so they can filter on other things other than the asset type, then on the internal function, we should respect their custom filter and concat the assetType filer

movekevin: What is assetType here? FungibleAsset only goes by address while coin goes by the full struct tag. If this is only meant for fungible asset, this should always be an address

0xjinn: @movekevin Yep you got it. This one is interesting, because it is being used interchangeably. 

"asset_type": "0x1::aptos_coin::AptosCoin" 
"asset_type": "0xc2948283c2ce03aafbb294821de7ee684b06116bb378ab614fa2de07a99355a8"

Both of the above are valid.

0xjinn: @0xmaayan 

Hmm, but I think the asset_type is unique, so there is no need to do additional filter?

src/types/indexer.ts
0xjinn: So I think the problem here is that `GetFungibleAssetMetadataQuery["fungible_asset_metadata"]` is an array already - https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/types/generated/operations.ts#L470-L485

0xjinn:
Update

1. Instead of returning a single element, I updated to fallback to array, which is easier to manage IMO
2. Updated e2e to test the test where asset_type does not exist

src/api/fungibleAsset.ts
```diff
@@ -46,6 +46,29 @@ export class FungibleAsset {
...
+   *
+   * @param assetType The asset type of the fungible asset.
+   * e.g
+   * "0x1::aptos_coin::AptosCoin" for Aptos Coin
+   * "0xc2948283c2ce03aafbb294821de7ee684b06116bb378ab614fa2de07a99355a8" - address format if this is fungible asset
+   *
+   * @returns getFungibleAssetMetadata A fungible asset metadata item
+   */
+  async getFungibleAssetMetadataByAssetType(assetType: string): Promise<GetFungibleAssetMetadataResponse> {
+    return getFungibleAssetMetadata({
```
0xmaayan, line 62: So look like you return an array even though we explicitly query for an item. I think it makes more sense to return and item, and if it is not found it would simply resolve to `undefined` 
```
async getFungibleAssetMetadataByAssetType(assetType: string): Promise<GetFungibleAssetMetadataResponse[0]> {
    const res = await getFungibleAssetMetadata({
      aptosConfig: this.config,
      options: {
        where: {
          asset_type: { _eq: assetType },
        },
      },
    });
    return res[0];
  }
```

0xjinn: Updated.

PR #120, opened 2023-10-25
consolidate move struct and resource type

### LINKED ISSUE #117: [consistency] Document and reduce number of duplicated types
Looks like we have lots of duplicitous types:
MoveResourceType === MoveStructType, though one has docs another doesn’t
There’s a representation of a MoveModuleId and a ModuleId class, and a StructType which is a ModuleId plus a struct name
There’s also a representation of a FunctionId  which uses the StructType
It may be good to consolidate these in some way

Additionally, it would be good to document each accordingly.

0xjinn:
### Description

resolve #117 

Removing the MoveStructType and consolidate into MoveResourceType. 

We could keep both, and assign one to another. i.e 

```
export type MoveStructType = MoveResourceType;
export type MoveResourceType = `${string}::${string}::${string}`;
```

But I feel like it would just confuse the user. 

### Test Plan

pnpm fmt

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -249,7 +249,7 @@ export type GasEstimation = {
 };
 
 export type MoveResource = {
-  type: MoveResourceType;
+  type: MoveStructType;
   data: {};
 };
 
@@ -518,7 +518,7 @@ export type TransactionPayloadResponse = EntryFunctionPayloadResponse | ScriptPa
 
 export type EntryFunctionPayloadResponse = {
   type: string;
-  function: MoveResourceType;
+  function: MoveStructType;
   /**
    * Type arguments of the function
    */
@@ -683,7 +683,7 @@ export type MoveOptionType = MoveType | null | undefined;
 /**
  * String representation of a on-chain Move struct type.
  */
-export type MoveResourceType = `${string}::${string}::${string}`;
+export type MoveStructType = `${string}::${string}::${string}`;
 
 export type MoveType =
   | boolean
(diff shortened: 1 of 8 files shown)


xbtmatt:
I think `MoveStructType` makes more sense if we have to use one (because function names aren't `MoveResource`s technically), but honestly the type alias solution might be the most accurate. Maybe there's a more formal name for the `1::2::3` type format in the Move book that encompasses function names and structs/resources

0xmaayan:
> @xbtmatt Yeah I was thinking about using a alias option that I mentioned in the description. Not sure if that will confuse the user though.

Let's not use alias, that only creates confusion. We should go with a type name that makes sense to us

src/types/index.ts
```diff
@@ -667,7 +667,6 @@ export type MoveUint128Type = string;
 export type MoveUint256Type = string;
 export type MoveAddressType = string;
 export type MoveObjectType = string;
-export type MoveStructType = `${string}::${string}::${string}`;
 export type MoveOptionType = MoveType | null | undefined;
 /**
  * String representation of a on-chain Move struct type.
```
0xjinn, line 672: @xbtmatt Description will live here

0xmaayan (approved):

src/types/index.ts
xbtmatt: I think we should say something like...

```typescript
/**
* This is the format for a fully qualified struct, resource, or entry function in Move.
*/
```

Above it or something

PR #128, opened 2023-10-25
[typeTag] Use new typetag parser everywhere, remove old one

gregnazario:
### Description

Use the new TypeTag parser everywhere, with generic support.  This consistency is required for Remote ABI

### Test Plan
Existing tests.  Parser is already covered with separate tests, so duplicates were removed.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,7 +6,7 @@ export * from "./bcs";
 export * from "./client";
 export * from "./core";
 export * from "./transactions/types";
-export * from "./transactions/typeTag/typeTag";
+export * from "./transactions/typeTag";
 export * from "./types";
 export * from "./utils/apiEndpoints";
 export * from "./utils/hdKey";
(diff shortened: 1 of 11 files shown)


src/index.ts
```diff
@@ -7,6 +7,7 @@ export * from "./client";
 export * from "./core";
 export * from "./transactions/types";
 export * from "./transactions/typeTag/typeTag";
+export * from "./transactions/typeTag/parser";
```
0xmaayan, line 10: lets just have an index.ts file that exports each?

src/transactions/typeTag/typeTag.ts
```diff
@@ -38,6 +38,9 @@ export abstract class TypeTag extends Serializable {
         return TypeTagU32.load(deserializer);
       case TypeTagVariants.U256:
         return TypeTagU256.load(deserializer);
+      case TypeTagVariants.Generic:
+        // This is specifically a placeholder and does not represent a real type
```
0xmaayan, line 42: a bit more explanation on why?

0xmaayan (approved):

src/transactions/typeTag/typeTag.ts
gregnazario: So, this is not a real type (only for the ABI).  We will have to do some smarter stuff around the generics, which actually I have to go back and fix something here first.

PR #135, opened 2023-10-27
[Small change] Add 0x0 to the AccountAddress class as `AccountAddress.ZERO` in anticipation of optional fee payer address

xbtmatt:
### Description
While working on a few things, I realized we don't have a static field for the `0x0` address, and we'll be using it a lot soon

I added simple unit tests for it in `accountAddress.test.ts` and added it to the CHANGELOG


0xjinn (approved):
Good call! I need this for fee payer

PR #136, opened 2023-10-27
[Naming] Unifying type names so they reflect the rust enums more accurately

xbtmatt:
### Description
As I'm exporting types for the wallet adapter stuff, I'm coming across inconsistent type names, so I'm trying to make sure they make sense.

### Test Plan
```typescript
pnpm jest
```

### DIFF
--- a/src/transactions/instances/transactionPayload.ts
+++ b/src/transactions/instances/transactionPayload.ts
@@ -334,7 +334,7 @@ export class Script {
 export class MultiSig {
   public readonly multisig_address: AccountAddress;
 
-  public readonly transaction_payload?: TransactionPayloadMultiSig;
+  public readonly transaction_payload?: MultisigTransactionPayload;
 
   /**
    * Contains the payload to run a multi-sig account transaction.
@@ -344,7 +344,7 @@ export class MultiSig {
    * @param transaction_payload The payload of the multi-sig transaction. This is optional when executing a multi-sig
    *  transaction whose payload is already stored on chain.
    */
-  constructor(multisig_address: AccountAddress, transaction_payload?: TransactionPayloadMultiSig) {
+  constructor(multisig_address: AccountAddress, transaction_payload?: MultisigTransactionPayload) {
     this.multisig_address = multisig_address;
     this.transaction_payload = transaction_payload;
   }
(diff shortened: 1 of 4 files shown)


src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -35,15 +35,15 @@ import {
   TransactionAuthenticatorEd25519,
   TransactionAuthenticatorFeePayer,
   TransactionAuthenticatorMultiAgent,
-  SingleSenderTransactionAuthenticator,
+  TransactionAuthenticatorSingleSender,
```
gregnazario, line 38: This change makes sense for consistency, but why was it made in the first place without the type first?

gregnazario (commented):
Maybe we can flip all of them instead of having them the way they are now?

src/transactions/transaction_builder/transaction_builder.ts
xbtmatt: I'm not sure- I think Maayan and I were just trying to get it working and it slipped in like this.

Tbh I don't remember when it even became this, I was confused when I saw it

gregnazario: Not that they're both different ways, but that if you were to choose a name not thinking of either one.  Which one would you choose:

1. TypeTransactionAuthenticator
2. TransactionAuthenticatorType

That should be the one you choose (since we're just breaking anything over better UX)

xbtmatt: In general I prefer the more specific descriptor first, because when you're typing auto complete works faster that way haha

And visually it's easier to see what it is that way

src/transactions/instances/transactionPayload.ts
```diff
@@ -366,16 +366,21 @@ export class MultiSig {
...
   }
 }
 
 /**
- * Representation of a MultiSig Transaction Payload that can serialized and deserialized
+ * Representation of a MultiSig Transaction Payload from `multisig_account.move`
+ * that can be serialized and deserialized
+
+ * This class exists right now to represent an extensible transaction payload class for
+ * transactions used in `multisig_account.move`. Eventually, this class will be able to
```
0xmaayan, line 380: sounds too much information just to say "for now, support only entry function payload "

0xmaayan (approved):

PR #137, opened 2023-10-27
[tests] Removed forced types discrimination

hardsetting:
### Description
- Added "0x" to the return type of some methods from AccountAddress
- Leveraging generic signature to simplify test helper (pair programming with @xbtmatt 

### Test Plan
Unit and e2e tests are passing

PR #138, opened 2023-10-27
[Easy] Fix MoveStructType address format

0xjinn:
### Description

Since we loosen the type of MoveStructType, we do not need to specify the `0x` prefix. `toString()` of the address class should include the prefix already.

### Test Plan

pnpm test


xbtmatt:
Make sure you check out #137 before doing any more of these- there's a nice solution gabi and I found!

gregnazario (approved):

gregnazario:
> Make sure you check out #137 before doing any more of these- there's a nice solution gabi and I found!

This is only sort of helpful.

In the event someone stores their own string versions it will still be affected.  I had actually decided to not go forward with that, though it still does work all the same.

PR #140, opened 2023-10-28
[BCS] Adding toSimpleValue() function and changed MoveVector.U8 to allow empty array input

xbtmatt:
### Description
Adding a `toSimpleValue()` function to convert a BCS back to its basic simple/inner type

This helps us convert BCS arguments to unserialized JSON payloads

I also changed `MoveVector.U8` to accept an empty array as input by adding a length check as an additional OR clause instead of just a type check on the first element. 

### Test Plan
Added a few unit tests

### DIFF
--- a/src/bcs/serializable/entryFunctionBytes.ts
+++ b/src/bcs/serializable/entryFunctionBytes.ts
@@ -46,6 +46,10 @@ export class EntryFunctionBytes extends Serializable implements EntryFunctionArg
     serializer.serialize(this);
   }
 
+  /**
+   * This function converts a BCS class instance to its inner value representation as a simple typescript type.
+   * @returns the Uint8Array input that initialized this class instance
+   */
   toSimpleValue(): Uint8Array {
     return this.value.toSimpleValue();
   }
--- a/src/bcs/serializable/entryFunctionBytes.ts
+++ b/src/bcs/serializable/entryFunctionBytes.ts
@@ -46,8 +46,8 @@ export class EntryFunctionBytes extends Serializable implements EntryFunctionArg
     serializer.serialize(this);
   }
 
-  toInner(): Uint8Array {
-    return this.value.value;
+  toSimpleValue(): Uint8Array {
+    return this.value.toSimpleValue();
   }
 
   /**
--- a/src/bcs/serializable/entryFunctionBytes.ts
+++ b/src/bcs/serializable/entryFunctionBytes.ts
@@ -46,6 +46,10 @@ export class EntryFunctionBytes extends Serializable implements EntryFunctionArg
     serializer.serialize(this);
   }
 
+  toInner(): Uint8Array {
+    return this.value.value;
+  }
+
   /**
    * The only way to create an instance of this class is to use this static method.
    *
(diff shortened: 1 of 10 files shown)


src/bcs/serializable/moveStructs.ts
```diff
@@ -276,6 +284,10 @@ export class MoveOption<T extends Serializable & EntryFunctionArgument>
     serializer.serializeBytes(bcsBytes);
   }
 
+  toSimpleValue() {
+    return this.vec.toSimpleValue();
+  }
```
gregnazario, line 289: You don't want to do some special casing rather than return a vector?

Options are taken in as:
1. value
2. undefined / null

Vs this will get you the vector's value as an array (i'm assuming)

tests/unit/bcsHelper.test.ts
```diff
@@ -593,4 +593,70 @@ describe("Tests for the Serializable class", () => {
...
+      expect(u128.toSimpleValue()).toEqual(1n);
+      expect(u256.toSimpleValue()).toEqual(1n);
+      expect(bool.toSimpleValue()).toEqual(true);
+      expect(address.toSimpleValue()).toEqual(
+        new Uint8Array([
+          0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 2, 3, 4,
+        ]),
+      );
+      expect(str.toSimpleValue()).toEqual("This is a string. It's a simple string...");
+    });
```
gregnazario, line 621: Ideally test one way and then convert it back the other way?

tests/unit/bcsHelper.test.ts
```diff
@@ -593,4 +593,70 @@ describe("Tests for the Serializable class", () => {
...
+
+    it("correctly converts primitive types to inner nested types", () => {
+      const vector = MoveVector.U8([1, 2, 3]);
+      const option = MoveOption.U8(1);
+      expect(vector.toSimpleValue()).toEqual([1, 2, 3]);
+      expect(option.toSimpleValue()).toEqual([1]);
```
gregnazario, line 627: This should be just `1`, I think, or we need a better representation of Option like in Rust...

src/bcs/serializable/moveStructs.ts
xbtmatt: Yeah it gives you back [] instead of `undefined` or `null` which I think makes sense because that's what it is on-chain, that's what the node returns for view functions, and ultimately that's how it's represented in the class (it converts everything to a  MoveVector, it just accepts undefined/null as inputs)

xbtmatt: Also the typing for `undefined` and `null` make it a pain to work with :p

I suppose we can do it, but I figure it's better to keep it consistent/simple

tests/unit/bcsHelper.test.ts
xbtmatt: ah I see what you mean better now- I was just trying to output what it was before with the previous remote ABI, but if that's not what makes sense it doesn't have to be

I do have an `unwrap()` function and it will return `1` but that will throw if it's undefined/null

src/bcs/serializable/moveStructs.ts
gregnazario: We can also change it to be an array of 1 for option on the input, just seemed confusing to me

src/bcs/serializable/entryFunctionBytes.ts
```diff
@@ -46,6 +46,10 @@ export class EntryFunctionBytes extends Serializable implements EntryFunctionArg
     serializer.serialize(this);
   }
 
+  toSimpleValue(): Uint8Array {
```
0xjinn, line 49: Can we add comment and an example for this function? :)

PR #141, opened 2023-10-29
Implement `MultiKey` authentication key

0xmaayan:
### Description
- Implement `MultiKey` that conforms to [AIP-55](https://github.com/aptos-foundation/AIPs/pull/263).
- Add multiSig example that uses MultiKey class
- Fix verify signature logic on public key module
- Simplify code to use `senderAuthentication` when generating signed transaction as part of the transaction builder flow
- Add clarity and explanations around `Ed25519PublicKey`, `Secp256k1PublicKey` and `AnyPublicKey` classes

Follow up:
- Implement `verifySignature` in `MultiKey` class

### Test Plan
pnpm test

### DIFF
--- a/src/core/crypto/multiKey.ts
+++ b/src/core/crypto/multiKey.ts
@@ -1,0 +1,122 @@
+import { Hex } from "../hex";
+import { HexInput } from "../../types";
+import { Deserializer } from "../../bcs/deserializer";
+import { Serializer } from "../../bcs/serializer";
+import { AnyPublicKey } from "./anyPublicKey";
+import { AnySignature } from "./anySignature";
+import { PublicKey } from "./asymmetricCrypto";
+
+export class MultiKey extends PublicKey {
+  /**
+   * List of any public keys
+   */
+  public readonly publicKeys: AnyPublicKey[];
+
+  /**
+   * The minimum number of valid signatures required, for the number of public keys specified
+   */
+  public readonly signaturesRequired: number;
+
+  constructor(args: { publicKeys: PublicKey[]; signaturesRequired: number }) {
+    super();
+    const { publicKeys, signaturesRequired } = args;
+
+    // Validate number of public keys is greater than signature required
+    if (signaturesRequired < 0) {
+      throw new Error("The number of required signatures needs to be greater then 0");
+    }
+
+    // Validate number of public keys is greater than signature required
+    if (publicKeys.length < signaturesRequired) {
+      throw new Error(
+        `Provided ${publicKeys.length} public keys is smaller than the ${signaturesRequired} required signatures`,
+      );
+    }
+
+    const keys: AnyPublicKey[] = [];
+    publicKeys.forEach((publicKey) => {
+      if (publicKey instanceof AnyPublicKey) {
+        keys.push(publicKey);
+      } else {
+        // if public key is instance of a legacy authentication key, i.e
+        // Legacy Ed25519, convert it into AnyPublicKey
+        keys.push(new AnyPublicKey(publicKey));
+      }
+    });
+
+    this.publicKeys = keys;
+    this.signaturesRequired = signaturesRequired;
+  }
+
+  toUint8Array(): Uint8Array {
+    return this.bcsToBytes();
+  }
+
+  /**
+   * Create a bitmap that holds the mapping from the original public keys
+   * to the signatures passed in
+   *
+   * @param args.bits array of the index mapping to the matching public keys
+   * @returns Uint8array bit map
+   */
+  createBitmap(args: { bits: number[] }): Uint8Array {
+    const { bits } = args;
+    // Bits are read from left to right. e.g. 0b10000000 represents the first bit is set in one byte.
+    // The decimal value of 0b10000000 is 128.
+    const firstBitInByte = 128;
+    const bitmap = new Uint8Array([0, 0, 0, 0]);
+
+    // Check if duplicates exist in bits
+    const dupCheckSet = new Set();
+
+    bits.forEach((bit: number, idx: number) => {
+      if (idx + 1 > this.publicKeys.length) {
+        throw new Error(`Signature index ${idx + 1} is out of public keys range, ${this.publicKeys.length}.`);
+      }
+
+      if (dupCheckSet.has(bit)) {
+        throw new Error(`Duplicate bit ${bit} detected.`);
+      }
+
+      dupCheckSet.add(bit);
+
+      const byteOffset = Math.floor(bit / 8);
+
+      let byte = bitmap[byteOffset];
+
+      // eslint-disable-next-line no-bitwise
+      byte |= firstBitInByte >> bit % 8;
+
+      bitmap[byteOffset] = byte;
+    });
+
+    return bitmap;
+  }
+
+  /**
+   * Hex string representationog the multi key bytes
+   *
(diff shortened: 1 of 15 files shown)


src/core/crypto/multiKey.ts
```diff
@@ -0,0 +1,122 @@
...
+  public readonly signaturesRequired: number;
+
+  constructor(args: { publicKeys: PublicKey[]; signaturesRequired: number }) {
+    super();
+    const { publicKeys, signaturesRequired } = args;
+
+    // Validate number of public keys is greater than signature required
+    if (signaturesRequired < 0) {
+      throw new Error("The number of required signatures needs to be greater then 0");
+    }
```
gregnazario, line 27: Typo
```suggestion
    // Validate number of public keys is greater than signature required
    if (signaturesRequired < 1) {
      throw new Error("The number of required signatures needs to be greater then 0");
    }
```

src/core/crypto/multiKey.ts
```diff
@@ -0,0 +1,122 @@
...
+    const keys: AnyPublicKey[] = [];
+    publicKeys.forEach((publicKey) => {
+      if (publicKey instanceof AnyPublicKey) {
+        keys.push(publicKey);
+      } else {
+        // if public key is instance of a legacy authentication key, i.e
+        // Legacy Ed25519, convert it into AnyPublicKey
+        keys.push(new AnyPublicKey(publicKey));
+      }
+    });
```
gregnazario, line 45: Tbh, this is probably confusing for a user form the outside.  PublicKey vs AnyPublicKey.  We may want to work on naming.  Similarly I saw elsewhere with transaction types that gets somewhat confusing

src/core/crypto/multiKey.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    this.publicKeys = keys;
+    this.signaturesRequired = signaturesRequired;
+  }
+
+  toUint8Array(): Uint8Array {
+    return this.bcsToBytes();
+  }
```
gregnazario, line 53: Sometimes I wonder if we want an interface for this, as something that can be made into bytes.  Is this any different than serialize?

src/core/crypto/multiKey.ts
```diff
@@ -0,0 +1,122 @@
...
+        throw new Error(`Signature index ${idx + 1} is out of public keys range, ${this.publicKeys.length}.`);
+      }
+
+      if (dupCheckSet.has(bit)) {
+        throw new Error(`Duplicate bit ${bit} detected.`);
+      }
+
+      dupCheckSet.add(bit);
+
+      const byteOffset = Math.floor(bit / 8);
```
gregnazario, line 83: I don't think you have to floor if you shift, though I'm sure the linter hates it.
 `bit >> 3`

src/core/crypto/multiKey.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  /**
+   * Hex string representationog the multi key bytes
+   *
+   * @returns string
+   */
+  toString(): string {
+    return Hex.fromHexInput(this.toUint8Array()).toString();
+  }
```
gregnazario, line 103: Ew, we'll have to make a shortcut for this

src/core/crypto/multiKey.ts
```diff
@@ -0,0 +1,122 @@
...
+      byte |= firstBitInByte >> bit % 8;
+
+      bitmap[byteOffset] = byte;
+    });
+
+    return bitmap;
+  }
+
+  /**
+   * Hex string representationog the multi key bytes
```
gregnazario, line 97:
```suggestion
   * Hex string representation the multi key bytes
```

src/core/crypto/multiKey.ts
```diff
@@ -0,0 +1,122 @@
...
+   */
+  toString(): string {
+    return Hex.fromHexInput(this.toUint8Array()).toString();
+  }
+
+  // TODO
+  // eslint-disable-next-line class-methods-use-this, @typescript-eslint/no-unused-vars
+  verifySignature(args: { message: HexInput; signature: AnySignature }): boolean {
+    throw new Error("not implemented");
+  }
```
gregnazario, line 109: I guess we're going to do this later?

src/types/index.ts
```diff
@@ -948,17 +948,15 @@ export enum SigningScheme {
...
 
 export enum SigningSchemeInput {
   /**
    * For Ed25519PublicKey
    */
   Ed25519 = 0,
-  /**
-   * For MultiEd25519PublicKey
-   */
-  MultiEd25519 = 1,
```
gregnazario, line 961: We will have to provide some way to handle this old key type for compatibility reasons, let's circle back on it though

tests/e2e/transaction/transactionSubmission.test.ts
```diff
@@ -619,6 +633,57 @@ describe("transaction submission", () => {
...
+      const bitmap = multiKey.createBitmap({ bits: [0, 2] });
+
+      // account1 and account3 sign the transaction
+      const account1Authenticator = aptos.signTransaction({ signer: singleSignerED25519SenderAccount, transaction });
+      const account3Authenticator = aptos.signTransaction({ signer: singleSignerSecp256k1Account, transaction });
+
+      const multiKeyAuth = new AccountAuthenticatorMultiKey(
+        multiKey,
+        [
+          // TODO find a fix
```
gregnazario, line 672: We can probably just make a common function on all types of authenticators, or do something that will extract it no matter what.  Mixing classes and interfaces is messy.

Let's look at this in a couple days

gregnazario (approved):

src/core/crypto/multiKey.ts
0xmaayan: we use `AnyPublicKey` to reflect and match the Rust code. I explain it in different places and tbh, I think it is confusing no matter what so we need to make sure we document it well.

https://github.com/aptos-labs/aptos-ts-sdk/pull/141/files#diff-45edb2a64043b75aa2bf8796e392178277847b4e02a18ffae66a5adca7779a6fR9

https://github.com/aptos-labs/aptos-ts-sdk/pull/141/files#diff-819f43565d33e1eb1e2e53882e6bca77220613ee111ade4500b72862b102e2eeR14

https://github.com/aptos-labs/aptos-ts-sdk/pull/141/files#diff-98942ba717d1ba7eafd0b4cfa834f6307d611d3ba525aa2ae68a93938b97e0b5R14

src/types/index.ts
0xmaayan: This is removed from `SigningSchemeInput` but it still in `SigningScheme`. This is because you should pass in `MultiEd25519` scheme when creating an Account

tests/e2e/transaction/transactionSubmission.test.ts
0xmaayan: an optional solution that needs to look at is function overload. `signTransaction` return `AccountAuthenticator` type, and if one wants to access inner variables, it needs to tell ts what is the type they want to use.

src/core/crypto/multiKey.ts
0xmaayan: yes, from the PR description 
```
Follow up:

Implement verifySignature in MultiKey class
```

PR #143, opened 2023-10-29
[Documentation] Update documentation for authentication key/account address derivation

xbtmatt:
### Description

Trying to make it clearer (and more explicit) what a derivation/signature scheme is used for and how the authentication keys/account addresses are derived, and what the schemes mean.

### DIFF
--- a/src/core/account.ts
+++ b/src/core/account.ts
@@ -185,10 +185,10 @@ export class Account {
    * Instantiates an account given a private key and a specified account address.
    * This is primarily used to instantiate an `Account` that has had its authentication key rotated.
    *
-   * @param privateKey PrivateKey - private key of the account
-   * @param address The account address
-   * @param args.legacy optional. If set to true, the keypair authentication keys will be derived with a Legacy scheme.
-   * Defaults to deriving an authentication key with a Unified scheme
+   * @param args.privateKey PrivateKey - the underlying private key for the account
+   * @param args.address A specified account address, in case the authentication key has been rotated
+   * @param args.legacy An optional flag to indicate that the authentication key derivation should
+   * use the legacy Ed25519 scheme. Defaults to false
    *
    * @returns Account
    */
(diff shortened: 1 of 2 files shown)


src/core/account.ts
```diff
@@ -151,10 +151,10 @@ export class Account {
    * Instantiates an account given a private key and a specified account address.
    * This is primarily used to instantiate an `Account` that has had its authentication key rotated.
    *
-   * @param privateKey PrivateKey - private key of the account
-   * @param address The account address
-   * @param args.legacy optional. If set to true, the keypair generated is a Legacy keypair. Defaults
-   * to generating a Unified keypair
+   * @param args.privateKey PrivateKey - the underlying private key for the account
+   * @param args.address AccountAddress - The initial account address before the authentication key was rotated
```
0xmaayan, line 155: why `before`? `aptos.findByPrivateKey` eventually passes the current account (also after key rotation) address on chain

src/core/authenticationKey.ts
```diff
@@ -48,9 +48,14 @@ export class AuthenticationKey {
...
 
   /**
-   * Creates an AuthenticationKey from seed bytes and a scheme
+   * Derives an AuthenticationKey from the public key seed bytes and an explicit derivation scheme.
+   *
+   * This facilitates explicitly targeting a specific derivation scheme for an authentication key
+   * to use for an account.
+   *
+   * This is useful for deriving an account address from an account whose address
+   * was derived using a legacy derivation scheme.
```
0xmaayan, line 57: it is not only for "legacy derivation scheme", right?

xbtmatt: Yea, I probably should have said `This is particularly useful for...` but even then it's a little misleading.

I think it should probably just be the first two sentences,  I'll change it to that

src/core/account.ts
xbtmatt: Hmm, this is actually wrong. I think it's actually the opposite of this... I will just change this back to `The account address to use`

xbtmatt: Actually what about this:

`The account address the `Account` will appear as when signing a transaction`

0xmaayan (approved):

PR #144, opened 2023-10-30
[TypeTag] Adding explicit return types for TypeTag types, fix vectors `toString()` function, add aptosCoinStructTag

xbtmatt:
### Description
1. Adding more explicit return types to facilitate using pattern matching types
2. The `TypeTagVector.toString()` function was returning `vectoru64` instead of `vector<u64>`
3. added `aptosCoinStructTag()` because it's so commonly used
4. changed `stringStructTag()` to be `async function` instead of `const ... async()` to match the others

### Test Plan
```typescript
pnpm jest
```

### DIFF
--- a/src/transactions/typeTag/index.ts
+++ b/src/transactions/typeTag/index.ts
@@ -9,6 +9,7 @@ import { Serializable, Serializer } from "../../bcs/serializer";
 import { AccountAddress } from "../../core";
 import { Identifier } from "../instances/identifier";
 import { TypeTagVariants } from "../../types";
+import { APTOS_COIN } from "../../utils/const";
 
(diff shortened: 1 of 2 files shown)


src/transactions/typeTag/index.ts
```diff
@@ -288,7 +289,7 @@ export class TypeTagVector extends TypeTag {
 }
 
 export class TypeTagStruct extends TypeTag {
-  toString(): string {
+  toString(): `0x${string}::${string}::${string}${string}` {
```
gregnazario, line 292: I'd remove the second `${string}` at the end.

gregnazario (approved):
Please remove the extra `${string}` and merge it in

PR #145, opened 2023-10-30
[address] Make Address inputs also accept AccountAddress

gregnazario:
### Description
This should make processing AccountAddress easier.

HexInput was made for variable length hex inputs, which were uint8array, or a string.  But, it forgets about AccountAddress as a possible input / parsing for an input of a variable.  This adds a single function to parse it no matter what.

Note:
1. Do we want to remove Uint8array from a default input?  It doesn't seem like it would be common
2. Do we want to require 0x even for ones missing some 0s?  I think so, because this at a minimum should be adopted.
3. Do we just want to adjust the constructor?  So we don't have this `from` structure we use everywhere?

### Test Plan
All the tests pass, I still need to add a single test to check that AccountAddress works as an input as well.

### DIFF
--- a/src/api/account.ts
+++ b/src/api/account.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { AptosConfig } from "./aptosConfig";
-import { AccountAddress, PrivateKey, Account as AccountModule } from "../core";
+import { AccountAddress, PrivateKey, Account as AccountModule, AccountAddressInput } from "../core";
 import {
   AccountData,
   GetAccountCoinsDataResponse,
@@ -10,7 +10,6 @@ import {
   GetAccountOwnedObjectsResponse,
   GetAccountOwnedTokensFromCollectionResponse,
   GetAccountOwnedTokensQueryResponse,
-  HexInput,
   LedgerVersion,
   MoveModuleBytecode,
   MoveResource,
@@ -64,7 +63,7 @@ export class Account {
    * }
    * ```
    */
-  async getAccountInfo(args: { accountAddress: HexInput }): Promise<AccountData> {
+  async getAccountInfo(args: { accountAddress: AccountAddressInput }): Promise<AccountData> {
     return getInfo({ aptosConfig: this.config, ...args });
   }
 
@@ -83,7 +82,7 @@ export class Account {
    */
 
   async getAccountModules(args: {
-    accountAddress: HexInput;
+    accountAddress: AccountAddressInput;
     options?: PaginationArgs & LedgerVersion;
   }): Promise<MoveModuleBytecode[]> {
     return getModules({ aptosConfig: this.config, ...args });
@@ -107,7 +106,7 @@ export class Account {
    * ```
    */
   async getAccountModule(args: {
-    accountAddress: HexInput;
+    accountAddress: AccountAddressInput;
     moduleName: string;
     options?: LedgerVersion;
   }): Promise<MoveModuleBytecode> {
@@ -127,7 +126,7 @@ export class Account {
    * @returns The account transactions
    */
   async getAccountTransactions(args: {
-    accountAddress: HexInput;
+    accountAddress: AccountAddressInput;
     options?: PaginationArgs;
   }): Promise<TransactionResponse[]> {
     return getTransactions({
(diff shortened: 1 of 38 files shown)


xbtmatt:
Love this!!

1. I think it's ok if we leave it as an input, but I don't think I would notice much if we took it out.
2. Yea, it looks confusing/weird without it (iiuc)
3. Yeah I understand the purpose behind the private constructor, but for such a common class it's kind of confusing/prohibitive when we use `new` very frequently. in general I think we should aim for making 99% of what people will use easy/simple, and then leave stricter modes for the 1% in another function

`.from` is okay, too. But I do feel like letting people use a `new` constructor with multiple input types will lead to less confusion overall

src/core/accountAddress.ts
```diff
@@ -21,6 +21,8 @@ export enum AddressInvalidReason {
   INVALID_PADDING_ZEROES = "INVALID_PADDING_ZEROES",
 }
 
+export type AccountAddressInput = HexInput | AccountAddress;
```
0xmaayan, line 24: can we have it in the global type file? src/types/index.ts

gregnazario: Sure, will do

gregnazario: This is... impossible, and creates a dependency loop that's really I think impossible to get rid of since it then depends on itself

src/api/transactionSubmission.ts
```diff
@@ -47,7 +47,7 @@ export class TransactionSubmission {
   /**
    * Generates any transaction by passing in the required arguments
    *
-   * @param args.sender The transaction sender's account address as a HexInput
+   * @param args.sender The transaction sender's account address as a AccountAddressInput
```
xbtmatt, line 50:
```suggestion
   * @param args.sender The transaction sender's account address as an AccountAddressInput
```

src/internal/transactionSubmission.ts
```diff
@@ -26,11 +26,12 @@ import {
   InputGenerateTransactionPayloadDataWithRemoteABI,
 } from "../transactions/types";
 import { UserTransactionResponse, PendingTransactionResponse, MimeType, HexInput } from "../types";
+import { AccountAddress, AccountAddressInput } from "../core";
 
 /**
  * Generates any transaction by passing in the required arguments
  *
- * @param args.sender The transaction sender's account address as a HexInput
+ * @param args.sender The transaction sender's account address as a AccountAddressInput
```
xbtmatt, line 34:
```suggestion
 * @param args.sender The transaction sender's account address as an AccountAddressInput
```

src/core/accountAddress.ts
```diff
@@ -335,33 +337,35 @@ export class AccountAddress extends Serializable implements TransactionArgument
...
+   * Convenience method for creating an AccountAddress from all known inputs.
    *
-   * @returns An instance of AccountAddress.
+   * This handles, Uint8array, string, and AccountAddress itself
+   * @param input
    */
-  static fromHexInputRelaxed(hexInput: HexInput): AccountAddress {
-    if (hexInput instanceof Uint8Array) {
-      return new AccountAddress({ data: hexInput });
+  static from(input: AccountAddressInput): AccountAddress {
```
xbtmatt, line 361: Beautiful!

xbtmatt (approved):

src/core/accountAddress.ts
0xmaayan: do we use HexInput anywhere now? if not, can remove this type and just use this as
```
export type AccountAddressInput = string | Uint8Array | AccountAddress;
```

0xmaayan (approved):

src/api/account.ts
```diff
@@ -64,7 +63,7 @@ export class Account {
    * }
    * ```
    */
-  async getAccountInfo(args: { accountAddress: HexInput }): Promise<AccountData> {
+  async getAccountInfo(args: { accountAddress: AccountAddressInput }): Promise<AccountData> {
```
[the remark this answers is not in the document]
gregnazario: There's some we'll clean up on another round, this PR is already too big

src/core/accountAddress.ts
gregnazario: We use HexInput for other things like hashes, and bytes

PR #146, opened 2023-10-30
Derive account from private key and derivation path

0xmaayan:
### Description
Derive Account from
1. `privateKey` - given a private key, we calculate locally the public key and derive an account.
2. `derivationPath` - add secp256k1 support and organize the code to use a static function from the corresponding PrivateKey class. Discussed with @alinush and followed the proposal https://www.notion.so/HD-wallets-secp256k1-key-derivation-1f733e956f1442a985a40f8ebf36e6a8#62f3d557e12243c6943ee2982d1de743 

### Test Plan
pnpm test


src/utils/hdKey.ts
```diff
@@ -95,17 +100,30 @@ const mnemonicToSeed = (mnemonic: string): Uint8Array => {
...
+      throw new Error(`Signing scheme ${scheme} not supported`);
   }
-
   // Derive the master key from the mnemonic
-  const { key, chainCode } = deriveKey(keyType, mnemonicToSeed(seedPhrase));
+  const { key, chainCode } = deriveKey(keyType, seedPhrase);
+  console.log("key", key);
+  console.log("chainCode", Hex.fromHexInput(chainCode).toString());
+  console.log("--------");
   const segments = splitPath(path).map((el) => parseInt(el, 10));
```
alinush, line 127: @0xmaayan can you print the `key` and `chainCode` returned here? Rather than the ones above?

src/core/account.ts
```diff
@@ -146,6 +145,42 @@ export class Account {
...
+        publicKey = new AnyPublicKey(privateKey.publicKey());
+      }
+    } else {
+      throw new Error(`Unsupported private key ${privateKey}`);
+    }
+
+    const authKey = AuthenticationKey.fromPublicKey({ publicKey });
+    const address = new AccountAddress({ data: authKey.toUint8Array() });
+    return new Account({ privateKey, address, legacy });
+  }
```
gregnazario, line 182: Is there a reason why not to make this and `PrivateKeyAndAddress` the same function, and just have an optional address?  This is pretty much the constructor just that it autogenerates the account address.

Maybe we can make all of these (except from mnemonic) into the constructor and remove any choices needed to be made between these 3 very similar paths?

It would be good to explore how to deal with auto lookups the best way (I'm aware of the function on `Aptos`) and reduce down the number of ways to create an Account as much as possible

src/core/crypto/hdKey.ts
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Note that for secp256k1, last two components must be non-hardened.
+ *
+ * @param path path string (e.g. `m/44'/637'/0'/0/0`).
+ */
+export function isValidBIP32Path(path: string): boolean {
+  if (!APTOS_BIP32_REGEX.test(path)) {
+    return false;
+  }
+  return true;
```
gregnazario, line 42: Can reduce to `!!APTOS_BIP32_REGEX.test(path)` or maybe even just `APTOS_BIP32_REGEX.test(path)`

src/core/crypto/hdKey.ts
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Note that for ed25519, all components must be hardened.
+ *
+ * @param path path string (e.g. `m/44'/637'/0'/0'/0'`).
+ */
+export function isValidHardenedPath(path: string): boolean {
+  if (!APTOS_HARDENED_REGEX.test(path)) {
+    return false;
+  }
+  return true;
```
gregnazario, line 59: Similarly

src/core/crypto/secp256k1.ts
```diff
@@ -173,6 +175,26 @@ export class Secp256k1PrivateKey extends PrivateKey {
...
+   * @param mnemonics the mnemonic seed phrase
+   */
+  static fromDerivationPath(path: string, mnemonics: string): Uint8Array {
+    if (!isValidBIP32Path(path)) {
+      throw new Error(`Invalid derivation path ${path}`);
+    }
+    const { privateKey } = HDKey.fromMasterSeed(mnemonicToSeed(mnemonics)).derive(path);
+
+    // library returns privateKey as Uint8Array | null
+    if (privateKey == null) {
```
gregnazario, line 192: Unless empty byte array is possible
```suggestion
    if (privateKey === null) {
```

gregnazario (commented):
Some comments, I would like to see if our favorite cryptographer @alinush is good with this

src/core/account.ts
0xmaayan: tbh, I dont see any reason to have one function for both options. In general I am not a fan of optional arguments and to try and squish as much arguments into one function. It becomes messy really fast and explain what is going on is just confusing. Also, those 2 functions are pretty different in their usage, `fromPrivateAndAddress` cares for key rotation while `fromPrivateKey` doesnt.

gregnazario (approved):

src/core/crypto/ed25519.ts
alinush, line 1: Is this the right place to add the note about supporting hardened derivation only (since Ed25519 lacks a key homomorphism, so non-hardened derivation cannot work).

src/core/crypto/hdKey.ts
```diff
@@ -0,0 +1,99 @@
...
+export type DerivedKeys = {
+  key: Uint8Array;
+  chainCode: Uint8Array;
+};
+
+/**
+ * Aptos derive path is 637
+ */
+export const APTOS_HARDENED_REGEX = /^m\/44'\/637'\/[0-9]+'\/[0-9]+'\/[0-9]+'?$/;
+export const APTOS_BIP32_REGEX = /^m\/44'\/637'\/[0-9]+'\/[0-9]+\/[0-9]+$/;
```
alinush, line 17: This is a BIP-44 regex (see https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#path-levels)

src/core/crypto/hdKey.ts
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * A list of supported key types and associated seeds
+ */
+export enum KeyType {
```
alinush, line 22: Where is secp256k1? It should be here set to "Bitcoin seed", no?

src/core/crypto/hdKey.ts
```diff
@@ -0,0 +1,99 @@
...
+ */
+export function isValidBIP32Path(path: string): boolean {
+  return APTOS_BIP32_REGEX.test(path);
+}
+
+/**
+ * Aptos derive path is 637
+ *
+ * Parse and validate a path that is compliant to SLIP-0010 in form m/44'/637'/{account_index}'/{change_index}'/{address_index}'.
+ * See {@link https://github.com/satoshilabs/slips/blob/master/slip-0044.md}
```
alinush, line 46: Add link to BIP-44 also.

src/core/crypto/hdKey.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * Note that for ed25519, all components must be hardened.
```
alinush, line 48: Add an extra note here, that will be useful later: "This is because non-hardened [PK] derivation would not work due to the lack of key homomorphism. Specifically, you cannot derive the PK associated with derivation path `a/b/c` given the PK of `a/b` because there is no key homomorphism in Ed25519. This is because the PK in Ed25519 is, more or less, computed as $g^{H(sk)}$."

alinush, line 48: When referring to the signature scheme, Ed25519 should be capitalized!

src/core/crypto/hdKey.ts
```diff
@@ -0,0 +1,99 @@
...
+export enum KeyType {
+  ED25519 = "ed25519 seed",
+}
+
+export const HARDENED_OFFSET = 0x80000000;
+
+/**
+ * Aptos derive path is 637
+ *
+ * Parse and validate a path that is compliant to BIP-32 in form m/44'/637'/{account_index}'/{change_index}/{address_index}
```
alinush, line 31: BIP-44, not 32.

src/core/crypto/hdKey.ts
```diff
@@ -0,0 +1,99 @@
...
+ * for Secp256k1
+ *
+ * Note that for secp256k1, last two components must be non-hardened.
+ *
+ * @param path path string (e.g. `m/44'/637'/0'/0/0`).
+ */
+export function isValidBIP32Path(path: string): boolean {
```
alinush, line 38: Rename to isValidBip44Path

tests/unit/helper.ts
```diff
@@ -13,7 +14,15 @@ export const wallet = {
   publicKey: "0xea526ba1710343d953461ff68641f1b7df5f23b9042ffa2d2a798d3adb3f3d6c",
 };
 
-/* eslint-disable max-len */
+export const secp256k1WalletTestObject = {
```
alinush, line 17: Can you document where you got this test case from?

Also, what about the SLIP-0010 test vectors? Why not add one of those as well?

tests/unit/hdKey.test.ts
```diff
@@ -1,43 +1,64 @@
...
+    });
   });
 
-  it("Deriving a key is valid and different with different paths", () => {
-    const keys0 = derivePrivateKeyFromMnemonic(KeyType.ED25519, wallet.path, wallet.mnemonic);
-    const keys1 = derivePrivateKeyFromMnemonic(KeyType.ED25519, wallet.path.replace("0", "1"), wallet.mnemonic);
-    expect(keys0.key).toEqual(Hex.fromHexInput(wallet.privateKey).toUint8Array());
-    expect(keys0.key).not.toEqual(keys1.key);
-    expect(keys0.chainCode).not.toEqual(keys1.chainCode);
+  describe("BIP32 Path", () => {
```
alinush, line 35: BIP44

src/core/crypto/hdKey.ts
alinush: But actually, do you even use `KeyType` anywhere anymore?

alinush (changes requested):
Looks good, except for some minor editorial issues.

src/core/account.ts
```diff
@@ -169,19 +204,34 @@ export class Account {
...
+    let privateKey;
+    switch (scheme) {
+      case SigningSchemeInput.Secp256k1Ecdsa:
+        privateKey = new Secp256k1PrivateKey(Secp256k1PrivateKey.fromDerivationPath(path, mnemonic));
+        break;
+      case SigningSchemeInput.Ed25519:
+        privateKey = new Ed25519PrivateKey(Ed25519PrivateKey.fromDerivationPath(path, mnemonic));
+        break;
+      default:
+        throw new Error(`Unsupported private key ${privateKey}`);
```
0xjinn, line 232:
```suggestion
        throw new Error(`Unsupported scheme ${scheme}`);
```

should this be scheme instead of private key?

src/core/crypto/hdKey.ts
0xmaayan: no. we use `const { privateKey } = HDKey.fromMasterSeed(mnemonicToSeed(mnemonics)).derive(path)`
https://github.com/aptos-labs/aptos-ts-sdk/pull/146/files#diff-98942ba717d1ba7eafd0b4cfa834f6307d611d3ba525aa2ae68a93938b97e0b5R189

tests/unit/helper.ts
0xmaayan: it is a bit messy because of the regex validation check we do 
```
APTOS_BIP44_REGEX = /^m\/44'\/637'\/[0-9]+'\/[0-9]+\/[0-9]+$/;
```

alinush: Ah, I see.

0xmaayan: I was just using known mnemonic with this path to generate an account

src/core/crypto/hdKey.ts
alinush: So, then remove this enum?

tests/unit/helper.ts
alinush: Sorry, so how is the test case generated?

src/core/crypto/hdKey.ts
0xmaayan: cuz we need it for the Ed25519? and maybe for future implementation? tbh, it was implemented here before

tests/unit/helper.ts
0xmaayan: I used the `Account.fromDerivationPath()` with 
```
mnemonic: "shoot island position soft burden budget tooth cruel issue economy destroy above",
path: "m/44'/637'/0'/0/0",
```
and used the privateKey, publicKey and address it generated

src/core/crypto/ed25519.ts
0xmaayan: I'll add it on top of the static function https://github.com/aptos-labs/aptos-ts-sdk/pull/146/files#diff-819f43565d33e1eb1e2e53882e6bca77220613ee111ade4500b72862b102e2eeR193

src/core/crypto/hdKey.ts
alinush: I'm confused. I thought you just said it's not used anywhere...

tests/unit/helper.ts
alinush: I see. I worry that we don't have test cases generated with other libraries. I guess it's not as easy to add those because our code harcodes `637`.

src/core/crypto/hdKey.ts
0xmaayan: oh no, I meant we are not using it for secp256k1

tests/unit/helper.ts
0xmaayan: yes.

alinush: I would advise we find some generic BIP-44+32 library, generate a test case using it and then add it as a test here, to make sure we are compatible.

Otherwise, if we get this wrong, when hardware wallets try to support our (incorrect) BIP-44+32 implementation, we will have issues.

src/core/crypto/hdKey.ts
alinush: Hm... I see, it might be confusing to have this incomplete enum lying around...

tests/unit/helper.ts
0xmaayan: I guess we can test directly against https://github.com/aptos-labs/aptos-ts-sdk/pull/146/files#diff-98942ba717d1ba7eafd0b4cfa834f6307d611d3ba525aa2ae68a93938b97e0b5R189
```
const { privateKey } = HDKey.fromMasterSeed(mnemonicToSeed(mnemonics)).derive(path);
```
to bypass the validation check, this is what I did in my local tests. is this enough? 
if so, I can do it in a separate PR since we need this PR in to unblock work

alinush: That would defeat the point of testing, no? We are testing the code against itself, AFAICT.

We would need externally-computed private keys from some mnemonic + path.

But, sure, you can do it in another PR. 

After all, it looks like you are just directly using the scure/bip44 library, except for the `mnemonicToSeed` function, which cleans the input before calling scure/bip39. So, unless there's something wrong in our `mnemonicToSeed`, there shouldn't be any compatibility issues.

You could also just move this `HDKey.fromMasterSeed(mnemonicToSeed(mnemonics)).derive(path)` line into its own `fromDerivationPathInternal` function, and then test it against the SLIP-0010 vectors, avoiding the Aptos-specific path checks.

alinush (approved):
Looks good, except for:
1. The `KeyType`, which I think should be removed (or just put into the Ed25519-specific functions.
2. We need to generate some `secp256k1` test cases using external libraries/tools and make sure our code agrees with them. (It's likely that our code is fine, but best to have those tests to catch future bugs at least.)

PR #147, opened 2023-10-31
Strongly typed crypto proposal

hardsetting:
Discussed offline with @0xmaayan on how to possibly solve the following problem:

```typescript
// Here account is a "legacy" ed25519 account
signTransaction(args: { signer: Account; transaction: AnyRawTransaction }): AccountAuthenticatorEd25519

// Here we're using the new wrapped public key approach
signTransaction(args: { signer: Account; transaction: AnyRawTransaction }): AccountAuthenticatorSingleKey
```

I put a proposal under a different folder to keep everything contained, but if I get the green light I can do the porting with everything else.

Some implementation details:
- Renamed `AnyPublicKey` and `AnySignature` which are somewhat confusing, into `WrappedPublicKey` and `WrappedSignature`
- Renamed "Account" to "Signer" to be more more specific on the purpose of the class. We are passing a `Signer` instance to a function when we want to return a signature of some sort
- Making very clear what is legacy and what is not, and keeping them in separate classes when possible to avoid redundant switches
- Moved the public-key-specific authkey derivation logic under the public key implementation (that's where I would expect to find it)
- I defined a separate `Serializable` interface, as opposed to the current abstract class. I think that an interface is better as it can be added on top of other interfaces, and potentially another parent class. We now have a `toBcsBytes` pure function that takes a `Serializable` as input.
- Some keys and public keys implementations were definining a `load` method. With this implementation I removed them, as each signature and public key implementation is standalone, and it's the wrapper's responsibility to serialize/deserialize the variant integer prefix
- I didn't use object arguments in order to prototype faster, but will change it back if this looks good to y'all

Example of code completion:

Screenshot 2023-10-30 at 7 33 57 PM


src/core/playground/authenticationKey.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  // Extracted the common logic for authkey derivation and put it here.
+  // Don't love the name but we can find something better
+  static fromHashInput({ hashInput }: { hashInput: HexInput }): AuthenticationKey {
+    const hash = sha3Hash.create();
+    hash.update(Hex.fromHexInput(hashInput).toUint8Array());
+    const hashDigest = hash.digest();
+    return new AuthenticationKey({ data: hashDigest });
+  }
```
hardsetting, line 54: This is new, everything else is copied from the original class

src/core/playground/accountAuthenticator/legacyMultiEd25519.ts
```diff
@@ -0,0 +1,54 @@
...
+export class LegacyAccountAuthenticatorMultiEd25519 {
+  public readonly publicKey: LegacyMultiEd25519PublicKey;
+
+  public readonly signature: LegacyMultiEd25519Signature;
+
+  constructor(publicKey: LegacyMultiEd25519PublicKey, signature: LegacyMultiEd25519Signature);
+  constructor(publicKey: LegacyMultiEd25519PublicKey, authenticators: LegacyAccountAuthenticatorEd25519[]);
+  constructor(
+    publicKey: LegacyMultiEd25519PublicKey,
+    signatureOrAuthenticators: LegacyMultiEd25519Signature | LegacyAccountAuthenticatorEd25519[],
```
0xmaayan, line 17: why the function overload? and the 2 return types?

src/core/playground/authenticationKey.ts
```diff
@@ -0,0 +1,65 @@
...
+    return this.data.toString();
+  }
+
+  toUint8Array(): Uint8Array {
+    return this.data.toUint8Array();
+  }
+
+  // Extracted the common logic for authkey derivation and put it here.
+  // Don't love the name but we can find something better
+  static fromHashInput({ hashInput }: { hashInput: HexInput }): AuthenticationKey {
```
0xmaayan, line 49: I think we can use here one input param
```
static fromHashInput(authKeyBytes):
```

src/core/playground/bcs.ts
```diff
@@ -0,0 +1,23 @@
+import { Serializer } from "../../bcs";
+import { Hex } from "../hex";
+import type { Serializable } from "./interfaces";
+
+/**
+ * Serializes a `Serializable` value to its BCS representation.
+ * This function is the Typescript SDK equivalent of `bcs::to_bytes` in Move.
+ * @returns the BCS representation of the Serializable instance as a byte buffer
+ */
+export function bcsToBytes(serializable: Serializable): Uint8Array {
```
0xmaayan, line 10: we already have these functions right?

src/core/playground/interfaces.ts
```diff
@@ -0,0 +1,21 @@
...
+  verifySignature(message: HexInput, signature: TSignature): boolean;
+}
+
+export interface PrivateKey<TSignature, TPublicKey extends PublicKey<TSignature>> {
+  sign(message: HexInput): TSignature;
+
+  publicKey(): TPublicKey;
+}
+
+export interface BaseSigner<TAccountAuthenticator extends AccountAuthenticator = AccountAuthenticator> {
```
0xmaayan, line 19: is this to support legacy ed/multied and single signer?

src/core/playground/legacy/ed25519Signer.ts
```diff
@@ -0,0 +1,36 @@
+import { HexInput } from "../../../types";
+import { AccountAddress } from "../../accountAddress";
+import { LegacyAccountAuthenticatorEd25519 } from "../accountAuthenticator";
+import { Ed25519PrivateKey, Ed25519PublicKey } from "../ed25519";
+import type { BaseSigner } from "../interfaces";
+
+export class LegacyEd25519Signer implements BaseSigner<LegacyAccountAuthenticatorEd25519> {
```
0xmaayan, line 7: is this to replace Account?

src/core/playground/legacy/multiEd25519.ts
```diff
@@ -0,0 +1,293 @@
...
+    bytes[this.publicKeys.length * Ed25519PublicKey.LENGTH] = this.threshold;
+
+    return bytes;
+  }
+
+  toString(): string {
+    return Hex.fromHexInput(this.toUint8Array()).toString();
+  }
+
+  verifySignature({ message, signature }: { message: HexInput; signature: LegacyMultiEd25519Signature }): boolean {
```
0xmaayan, line 98: thanks!

src/core/playground/wrapped.ts
```diff
@@ -0,0 +1,78 @@
...
+import { AuthenticationKey } from "./authenticationKey";
+import { bcsToBytes } from "./bcs";
+import { Ed25519PublicKey, Ed25519Signature } from "./ed25519";
+import type { PublicKey } from "./interfaces";
+import { SignatureScheme } from "./scheme";
+import { Secp256k1PublicKey, Secp256k1Signature } from "./secp256k1";
+
+export type AllowedSignatures = Ed25519Signature | Secp256k1Signature;
+
+export class WrappedSignature<TSignature extends AllowedSignatures = AllowedSignatures> {
```
0xmaayan, line 12: tbh, I prefer `AnySignature` over `WrappedSignature`, I also like `AllowedSignature`

0xmaayan (commented):
Thanks! left some comments.

I think some of the structure can be a bit confusing, would love for @0xjinn @xbtmatt @gregnazario to take a look and provide their feedback around how understandable they find it.

src/core/playground/accountAuthenticator/legacyMultiEd25519.ts
hardsetting: One cool aspect of it is that the user doesn't have to worry about creating the bitmap :)

hardsetting: > and the 2 return types

which return types?

src/core/playground/interfaces.ts
hardsetting: This would be a common interface between a `Signer` and a `LegacyEd25519Signer`.
Having this defined is possibly not required. It can be useful to allow passing any signer to a method (e.g. `signTransaction`), but we'll see if/when integrating with the rest of the code if this should stay or go

src/core/playground/wrapped.ts
hardsetting: Fair, it would be good to be consistent with the rust SDK anyway.
Will revert

src/core/playground/legacy/ed25519Signer.ts
hardsetting: `LegacyEd25519Signer` and `Signer` will together replace `Account`

hardsetting: I added the `Legacy` prefix, but open to use different names if we don't plan to deprecate them

src/core/playground/bcs.ts
hardsetting: Yeah, just made them into pure functions instead of abstract class methods.
In general I personally prefer using interfaces and pure functions (it works better with duck-typed languages like python and javascript).

But talking offline with @xbtmatt I think there's a general preference on using a `Serializable` abstract class, with methods.

src/core/playground/accountAuthenticator/accountAuthenticator.ts
```diff
@@ -0,0 +1,21 @@
+import type { Deserializer } from "../../../bcs";
+import { LegacyAccountAuthenticatorEd25519 } from "./legacyEd25519";
+import { LegacyAccountAuthenticatorMultiEd25519 } from "./legacyMultiEd25519";
+import { AccountAuthenticatorSingleKey } from "./singleKey";
+import { AccountAuthenticatorVariant } from "./variant";
+
+export class AccountAuthenticator {
```
0xjinn, line 7: Hm, so this is not an abstract class anymore?

src/core/playground/scheme.ts
```diff
@@ -0,0 +1,4 @@
+export enum SignatureScheme {
```
0xjinn, line 1: I think we already have equivalent -> SigningScheme.

0xjinn:
-  I like the rename of the Account -> Signer. It make sense because we call them Signer in Move as well. Though I wonder if this will confuse the devlopers out there.
- Legacy - Agree that this is an issue. The concept of “legacy” is even confusing to me as well. Having separate class might be ideal. Curious on what other think

src/core/playground/accountAuthenticator/legacyMultiEd25519.ts
0xmaayan: sorry, input types 
`signatureOrAuthenticators: LegacyMultiEd25519Signature | LegacyAccountAuthenticatorEd25519[]`

src/core/playground/legacy/ed25519Signer.ts
0xmaayan: I think Legacy prefix is fine and good. We would need to support both forever

src/core/playground/scheme.ts
0xmaayan: note: there is SigningScheme and SigningSchemeInput.
SigningScheme represents all SigningSchemes (ed,multied,single signer, multi signer)
SigningSchemeInput represents what we accept (ed,secp)

src/core/playground/signer.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export interface GenerateSignerArgs<TSignatureScheme extends SignatureScheme = SignatureScheme> {
+  scheme: TSignatureScheme;
+}
+
+export class Signer<
+  TSignature extends AllowedSignatures = AllowedSignatures,
+  TPublicKey extends PublicKey<TSignature> = PublicKey<TSignature>,
+> implements BaseSigner<AccountAuthenticatorSingleKey<TSignature, TPublicKey>>
```
gregnazario, line 25: Signer could be good over an Account.  Especially if you could sign as an object in the future...

src/core/playground/interfaces.ts
gregnazario: BaseSigner could be confusing in the case of a multisigner, but what I would add is:

1. sign
2. verify
3. public key

Any signer should be able to sign and verify any blob, and show proof that they can verify it (especially since they need the public key anyways)

src/core/playground/authenticationKey.ts
gregnazario: This should... be instead fromSchemeAndBytes({scheme: Scheme, input: HexInput)

It should take the scheme, and the bytes required to generate the authentication key.

Right now, it's unclear what I'd put here since, if it's the public key, I still need to prepend (or append I forget), with the scheme to get the correct auth key.

The input, may be a key, it may be a set of keys etc.

src/core/playground/accountAuthenticator/legacyMultiEd25519.ts
gregnazario: This overload is kinda weird with `signaureOrAuthenticators`

Why not just:

(publicKey, signature)
(publicKey, authenticators)

And then do whatever type magic to handle one or the other?

.eslintrc.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
     "max-classes-per-file": ["error", 10],
     "import/prefer-default-export": "off",
     "object-curly-newline": "off",
+    "no-restricted-syntax": "off",
```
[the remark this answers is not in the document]
hardsetting: It's for a stupid airbnb rule preventing `for (const ... of ...) { ... }`

I can probably replace it with this custom config that I'm using in the wallet repo, to only whitelist for .. of 
```
   // Replacing airbnb rule with following, to re-enable "ForOfStatement"
    'no-restricted-syntax': [
      'error',
      'ForInStatement',
      'LabeledStatement',
      'WithStatement',
    ],
```

src/core/playground/signer.ts
hardsetting: what do you mean by "signing as an object"?

src/core/playground/interfaces.ts
hardsetting: I don't think it makes sense to have a multi-key signer.
The flow is fundamentally different and it would only makes sense if a dapp owns all the secret keys.

For multi-key flows (whether it's multi-sig or multi-agent) we have individual single-key signers signing a common message, and then we combine the signatures.
We should aim to facilitate that (hence the `authenticators` overload in `LegacyMultiEd25519AccountAuthenticator`)

src/core/playground/signer.ts
gregnazario: Today the only things that can sign are accounts, but imagine other things signing transactions

src/core/playground/interfaces.ts
gregnazario: All signers should be able to verify and get the public keys associated, single or multi (otherwise how do you verify).

Single signers should be able to sign (you have the key)

Sign shouldn't be applicable if there was a type called MultiKeySigner, since it doesn't have all the keys.

src/core/playground/authenticationKey.ts
hardsetting: > This should... be instead fromSchemeAndBytes({scheme: Scheme, input: HexInput)

I like it

.eslintrc.js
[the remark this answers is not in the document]
hardsetting: just updated

src/core/playground/accountAuthenticator/accountAuthenticator.ts
hardsetting: It's kinda hard to achieve this without cyclical dependency :(

hardsetting: only solution I could find is to have everything in the same file, which is ok but not really scalable

hardsetting: I thought "what's requiring other authenticators to extend from a base AccountAuthenticator?" and removed the "extends" for now

src/core/playground/accountAuthenticator/legacyMultiEd25519.ts
hardsetting: That's the actual implementation, that needs to take into account all overloads unfortunately (it's still JS at the end of the day)

The only possible overloads are 
```typescript
  constructor(publicKey: LegacyMultiEd25519PublicKey, signature: LegacyMultiEd25519Signature);
  constructor(publicKey: LegacyMultiEd25519PublicKey, authenticators: LegacyAccountAuthenticatorEd25519[]);
```

when definining overloads in TS, the main implementation is actually not available as overload, and would need to be specified explicitly.

src/core/playground/signer.ts
hardsetting: gotcha, interesting

src/core/playground/interfaces.ts
hardsetting: > otherwise how do you verify

@xbtmatt and I were talking about possibly always using authenticators to pass signatures around.

An authenticator is typically a signature + publicKey and has all it needs to verify a message already.
I added a `verify(message): boolean` method to authenticators.

https://github.com/aptos-labs/aptos-ts-sdk/blob/e5e717e90002d0a0cfed4e305bbe854ec68ae107/src/core/playground/accountAuthenticator/singleKey.ts#L16-L18

So the flows (for single key) would be:
```typescript

// Signer side
function handleSignatureRequest(message: Uint8Array) {
  const authenticator = signer.sign(message);
  return authenticator;
}

// Verifier side (NO access to signing capabilities)
const authenticator = requestSignature(message); 

// either trust the response (the fullnode would eventually catch the mismatch)
const isVerified = authenticator.verify(message);

// or manually verify (needs to know the public key locally)
const publicKey = getPublicKey();
if (!isSameSignatureScheme(publicKey, authenticator) {
  throw new Error('Signature scheme mismatch');
}
const isVerified = publicKey.verifySignature(authenticator.signature);
```

We can move away from the idea of always using authenticators, and instead always returning signatures.
In that case I would suggest having a different class / interface we could call account and which has address and public key.

```typescript
interface Account<TSignature, TPublicKey<TSignature>> {
  readonly address: string;
  readonly publicKey: TPublicKey;
  
  verifySignature(signature: TSignature): boolean;
}

interface Signer<TSignatrure, TPublicKey<TSignature>> extends Account<TSignature, TPublicKey> {
  sign(message: Uint8Array): TSignature;
}
```

This would "encourage" dapp developers to keep track of the public keys and use them to verify the signature.

hardsetting:
Just defined the concept of `Account` (publicKey + address)
https://github.com/aptos-labs/aptos-ts-sdk/blob/8642418bd1b6b1ea947afc54377183cef2ca808e/src/core/playground/account.ts#L14-L37

and `Signer` now extends from it, adding the secretKey

https://github.com/aptos-labs/aptos-ts-sdk/blob/8642418bd1b6b1ea947afc54377183cef2ca808e/src/core/playground/signer.ts#L23-L58

src/core/playground/interfaces.ts
hardsetting: https://github.com/aptos-labs/aptos-ts-sdk/pull/147#issuecomment-1789871454

src/core/playground/accountAuthenticator/singleKey.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { PublicKey } from "../interfaces";
+import { type AllowedSignatures, AnyPublicKey, AnySignature } from "../wrapped";
+import { AccountAuthenticatorVariant } from "./variant";
+
+export class AccountAuthenticatorSingleKey<
+  TSignature extends AllowedSignatures = AllowedSignatures,
+  TPublicKey extends PublicKey<TSignature> = PublicKey<TSignature>,
+> {
+  constructor(
+    public readonly publicKey: AnyPublicKey<TSignature, TPublicKey>,
```
0xmaayan, line 12: the format we use in sdk is define the class properties and then set it in the constructor
```
class AccountAuthenticatorSingleKey<>{
 public readonly publicKey: AnyPublicKey<TSignature, TPublicKey>;
 public readonly signature: AnySignature<TSignature>

 constructor(){
    ......
 }
}
```

src/core/playground/authenticationKey.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  static fromSchemeAndBytes({
+    scheme,
+    input,
+  }: {
+    scheme: AuthenticationKeyScheme;
+    input: HexInput;
+  }): AuthenticationKey {
+    const inputBytes = Hex.fromHexInput(input).toUint8Array();
```
0xmaayan, line 54: I couldn't find it in this PR, but we have auth key generation based on scheme logic
https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/core/authenticationKey.ts#L79

src/core/playground/bcs.ts
0xmaayan: yes, I like the doing `<>.bcsToBytes()` (where `<>` is anything extends Serializable

src/core/playground/interfaces.ts
0xmaayan: there is a benefit to return authenticator because it also holds the publicKey

src/core/playground/legacy/ed25519Account.ts
```diff
@@ -0,0 +1,30 @@
+import { HexInput } from "../../../types";
+import { AccountAddress } from "../../accountAddress";
+import { Ed25519PublicKey, Ed25519Signature } from "../ed25519";
+
+export interface LegacyEd25519AccountConstructorArgs {
+  publicKey: Ed25519PublicKey;
+  address?: AccountAddress | HexInput;
```
0xmaayan, line 7: just fyi https://github.com/aptos-labs/aptos-ts-sdk/pull/145/files#diff-90c0fc23e4a21dceb9625abcd3bec876671b9a682f9cc59a119d8f62ed0376daR24

src/core/playground/legacy/multiEd25519Account.ts
```diff
@@ -0,0 +1,30 @@
+import { HexInput } from "../../../types";
+import { AccountAddress } from "../../accountAddress";
+import type { LegacyMultiEd25519PublicKey, LegacyMultiEd25519Signature } from "./multiEd25519";
+
+export interface LegacyMultiEd25519AccountConstructorArgs {
+  publicKey: LegacyMultiEd25519PublicKey;
+  address?: AccountAddress | HexInput;
+}
+
+export class LegacyMultiEd25519Account {
```
0xmaayan, line 10: is there a reason that all legacy use `Account` ?

src/core/playground/wrapped.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+}
+
+export class AnyPublicKey<
+  TSignature extends AllowedSignatures = AllowedSignatures,
+  TPublicKey extends PublicKey<TSignature> = PublicKey<TSignature>,
+> implements PublicKey<AnySignature<TSignature>>
+{
+  constructor(public readonly publicKey: TPublicKey) {}
```
0xmaayan, line 44: can we please define `public readonly publicKey: TPublicKey` in the class and not in the  constructor signature

src/core/playground/secp256k1.ts
```diff
@@ -0,0 +1,72 @@
...
+    const signature = secp256k1.sign(messageHashBytes, this.key.toUint8Array());
+    return new Secp256k1Signature(signature.toCompactRawBytes());
+  }
+
+  publicKey(): Secp256k1PublicKey {
+    const bytes = secp256k1.getPublicKey(this.key.toUint8Array(), false);
+    return new Secp256k1PublicKey(bytes);
+  }
+
+  static generate(): Secp256k1PrivateKey {
```
0xmaayan, line 68: if we have `generate()` in the class level, should we also have `fromDerivationPath`, `fromPrivateKey` and `fromPrivateKeyAndAddress`?

src/core/playground/signer.ts
```diff
@@ -0,0 +1,58 @@
...
+    super({ publicKey, address });
+    this.privateKey = privateKey;
+  }
+
+  static generate(): Signer<Ed25519Signature, Ed25519PublicKey>;
+  static generate(args: GenerateSignerArgs<SignatureScheme.Ed25519>): Signer<Ed25519Signature, Ed25519PublicKey>;
+  static generate(args: GenerateSignerArgs<SignatureScheme.Secp256k1>): Signer<Secp256k1Signature, Secp256k1PublicKey>;
+  static generate(args?: GenerateSignerArgs) {
+    const scheme = args?.scheme ?? SignatureScheme.Ed25519;
+    if (scheme === SignatureScheme.Ed25519) {
```
0xmaayan, line 43: we should care for legacy vs non-legacy

0xmaayan, line 43: use `switch` instead?

src/core/playground/authenticationKey.ts
hardsetting: Yes, what I'm proposing here is to have each public key define how to derive its authentication key, as opposed to have a switch somewhere else that needs to handle all variants of public keys and schemas.

src/core/playground/secp256k1.ts
hardsetting: Yes absolutely.
Didn't port them over yet cause I wanted to make sure people liked this design

src/core/playground/authenticationKey.ts
0xmaayan: ohh is the public key handling on this PR? or plan to add it later?

hardsetting: It's in this PR already

for AnyPublicKey
https://github.com/aptos-labs/aptos-ts-sdk/blob/8642418bd1b6b1ea947afc54377183cef2ca808e/src/core/playground/wrapped.ts#L50-L55

for Ed25519PublicKey
https://github.com/aptos-labs/aptos-ts-sdk/blob/8642418bd1b6b1ea947afc54377183cef2ca808e/src/core/playground/ed25519.ts#L47-L52

for LegacyEd25519PublicKey
https://github.com/aptos-labs/aptos-ts-sdk/blob/8642418bd1b6b1ea947afc54377183cef2ca808e/src/core/playground/legacy/multiEd25519.ts#L128-L133

src/core/playground/signer.ts
hardsetting: Good point, a switch would look much better now! (there was some other stuff before)

src/core/playground/legacy/multiEd25519Account.ts
0xmaayan: I guess I am confused as why Legacy classes as Account, but secp256k1 doesnt.

src/core/playground/signer.ts
0xmaayan: check out my other comment https://github.com/aptos-labs/aptos-ts-sdk/pull/147/files#r1379326061

We needed it to distinguish between the input sdk accepts and the actual enum representation. not sure if with your design is still needed?

hardsetting: Yeah I'm fine with removing the re-definition. I was hoping to move away from the "input" sufffix but I see there's a reason behind it.
Removing

src/core/playground/legacy/multiEd25519Account.ts
hardsetting: We never supported accounts with Secp256k1 signature schemes before

hardsetting: in SDK v1 and prior to adding Secp256k1 and AnyPublicKey / AnyPrivateKey, the only available account was Ed25519 which used an authentication key derived from the Ed25519PublicKey bytes directly.

We still need to support this, and that would be LegacyEd25519Signer / Account.
For newer accounts (either Ed25519, Secp256k1 or anything that comes after) we want to use signers based on `AnyPublicKey`.

At least this is the understanding that I got chatting with people. I would not be opposed to hiding this details from the user and have all Ed25519 signers use the legacy scheme for simplicity.

Also as additional context to your question, I don't think it's even possible to have a Secp256k1 "legacy" account, I think the Rust SDK doesn't support that.

0xmaayan: ok thanks for explaining. I think the missing piece for me was that this PR is signers based.

xbtmatt:
I might be missing something obvious- but why do we separate the `AccountAuthenticator<T1, T2>` types if they should never be decoupled?

Like you can't do `AccountAuthenticator<Ed25519, Secp256k1>` right? If not, I think the generics should at least be coupled, because separate generics suggests otherwise. but maybe I'm misunderstanding

Also, I love the `signer` concept- and I really want to coalesce the concept of:

1. a local Account you can sign with
2. a wallet provider/some external signer you can sign with

I've tried coming up with it myself but it's confusing reconciling the local signing process with an `Account` and the callback nature of a wallet provider/ledger/whatever

I've played around with the idea of a `signer` that has a callback function: `sign: (signingFunction: async() => { accountAuthenticator: AccountAuthenticator, accountAddress: AccountAddress });`

or something like that...

src/internal/account.ts
```diff
@@ -509,50 +506,45 @@ export async function getAccountOwnedObjects(args: {
  * NOTE: There is a potential issue once unified single signer scheme will be adopted
  * by the community.
  *
- * Becuase on could create 2 accounts with the same private key with this new authenticator type,
+ * Because one could create 2 accounts with the same private key with this new authenticator type,
```
0xmaayan, line 509: =)

src/internal/transactionSubmission.ts
```diff
@@ -279,48 +289,51 @@ export async function publicPackageTransaction(args: {
...
   });
 
   // Sign the challenge
   const challengeHex = challenge.bcsToBytes();
-  const proofSignedByCurrentPrivateKey = fromAccount.sign(challengeHex);
-  const proofSignedByNewPrivateKey = newAccount.sign(challengeHex);
+  const proofSignedByCurrentPrivateKey = fromSigner.sign(challengeHex);
+  const proofSignedByNewPrivateKey = newSigner.sign(challengeHex);
+
+  const fromSigningScheme = fromSigner instanceof Signer ? SigningScheme.SingleKey : SigningScheme.Ed25519;
```
0xmaayan, line 315: can do something to not use this `instanceof` check

src/core/legacy/ed25519Account.ts
```diff
@@ -0,0 +1,37 @@
+import { HexInput } from "../../types";
+import { AccountAddress } from "../accountAddress";
+import { Ed25519PublicKey, Ed25519Signature } from "../crypto";
+
+export interface LegacyEd25519AccountConstructorArgs {
```
0xmaayan, line 5: lets add comment explaining this class

src/core/legacy/ed25519Signer.ts
```diff
@@ -0,0 +1,43 @@
...
+import { AccountAddress } from "../accountAddress";
+import { Ed25519PrivateKey, Ed25519Signature } from "../crypto";
+import { LegacyEd25519Account } from "./ed25519Account";
+
+export interface LegacyEd25519SignerConstructorArgs {
+  privateKey: Ed25519PrivateKey;
+  address?: AccountAddress | HexInput;
+}
+
+export class LegacyEd25519Signer extends LegacyEd25519Account {
```
0xmaayan, line 11: lets add comment explaining this class

0xmaayan (commented):
Overall LGTM! 
- Do we want to rename `AccountAuthenticatorEd25519` to `AccountAuthenticatorLegacyEd25519`
- Could we add more info on what the `Account` class is for?
- Also, lets make sure we have comments on every class/instance/method explaining what is it for

src/core/legacy/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./ed25519Account";
+export * from "./ed25519Signer";
```
0xmaayan, line 2: where did https://github.com/aptos-labs/aptos-ts-sdk/pull/147#discussion_r1379037032 go...?

src/core/crypto/ed25519.ts
```diff
@@ -2,11 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import nacl from "tweetnacl";
+import { AuthenticationKey } from "../authenticationKey";
 import { PublicKey, PrivateKey, Signature } from "./asymmetricCrypto";
 import { Deserializer } from "../../bcs/deserializer";
 import { Serializer } from "../../bcs/serializer";
 import { Hex } from "../hex";
-import { HexInput } from "../../types";
+import { HexInput, SigningScheme as AuthenticationKeyScheme } from "../../types";
```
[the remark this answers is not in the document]
hardsetting: It was previously conflicting with another name, but not anymore, removing

src/core/crypto/secp256k1.ts
```diff
@@ -62,12 +62,15 @@ export class Secp256k1PublicKey extends PublicKey {
...
    * @returns true if the signature is valid
    */
-  verifySignature(args: { message: HexInput; signature: Secp256k1Signature }): boolean {
+  verifySignature(args: { message: HexInput; signature: Signature }): boolean {
     const { message, signature } = args;
-    const msgHex = Hex.fromHexInput(message).toUint8Array();
-    const sha3Message = sha3_256(msgHex);
-    const rawSignature = signature.toUint8Array();
-    return secp256k1.verify(rawSignature, sha3Message, this.toUint8Array());
+    if (!(signature instanceof Secp256k1Signature)) {
```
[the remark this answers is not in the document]
hardsetting: It needs to match the `PublicKey` interface, which should be able to match any signature.
Unless we don't care about having a common interface haha

.eslintrc.js
```diff
@@ -27,6 +27,8 @@ module.exports = {
     "max-classes-per-file": ["error", 10],
     "import/prefer-default-export": "off",
     "object-curly-newline": "off",
+    // Replacing airbnb rule with following, to re-enable "ForOfStatement"
+    "no-restricted-syntax": ["error", "ForInStatement", "LabeledStatement", "WithStatement"],
```
0xmaayan, line 31: why do we do that?

0xmaayan:
lets put all the changes this PR introduces in a changelog? also need to resolve conflicts

src/core/crypto/ed25519.ts
```diff
@@ -232,26 +192,93 @@ export class Ed25519PrivateKey extends PrivateKey {
     return new Ed25519PrivateKey(privateKey);
   }
 
-  static isPrivateKey(privateKey: PrivateKey): privateKey is Ed25519PrivateKey {
-    return privateKey instanceof Ed25519PrivateKey;
+  // endregion
```
0xmaayan, line 195: delete?

src/core/crypto/index.ts
```diff
@@ -1,11 +1,16 @@
...
 export * from "./multiEd25519";
-export * from "./secp256k1";
 export * from "./multiKey";
-export * from "./hdKey";
-export * from "./anyPublicKey";
-export * from "./anySignature";
+export * from "./secp256k1";
+export * from "./singleKey";
+
+export type PublicKeyInput = Ed25519PublicKey | MultiEd25519PublicKey | AnyPublicKey | MultiKey;
```
0xmaayan, line 16: isnt AnyPublicKey contain MultiKey already?

src/core/crypto/multiEd25519.ts
```diff
@@ -148,24 +194,28 @@ export class MultiEd25519Signature extends Signature {
    * @param args.bitmap 4 bytes, at most 32 signatures are supported. If Nth bit value is `1`, the Nth
    * signature should be provided in `signatures`. Bits are read from left to right
    */
-  constructor(args: { signatures: Ed25519Signature[]; bitmap: Uint8Array }) {
+  constructor(args: { signatures: Ed25519Signature[]; bitmap: Uint8Array | number[] }) {
```
0xmaayan, line 197: lets add the new type `number[]` to the function doc

src/core/crypto/multiEd25519.ts
```diff
@@ -184,6 +234,8 @@ export class MultiEd25519Signature extends Signature {
     return Hex.fromHexInput(this.toUint8Array()).toString();
   }
 
+  // endregion
```
0xmaayan, line 237: whats up with all those `region` `endregion`? is it something new?

src/core/crypto/multiKey.ts
```diff
@@ -1,12 +1,31 @@
...
-import { Serializer } from "../../bcs/serializer";
-import { AnyPublicKey } from "./anyPublicKey";
-import { AnySignature } from "./anySignature";
-import { PublicKey } from "./asymmetricCrypto";
+import { Serializable, Serializer } from "../../bcs/serializer";
+import { Ed25519PublicKey, Ed25519Signature } from "./ed25519";
+import { Secp256k1PublicKey, Secp256k1Signature } from "./secp256k1";
+import { AnyPublicKey, AnySignature } from "./singleKey";
+
+type PublicKeyInput = Ed25519PublicKey | Secp256k1PublicKey | AnyPublicKey;
```
0xmaayan, line 10: I feel this type repeats itself in a couple of places?

src/core/crypto/singleKey.ts
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Represents any public key supported by Aptos.
+ *
+ * Since [AIP-55](https://github.com/aptos-foundation/AIPs/pull/263) Aptos supports
+ * `Legacy` and `Unified` authentication keys.
+ *
+ * Any unified authentication key is represented in the SDK as `AnyPublicKey`.
+ */
+export class AnyPublicKey extends Serializable {
```
0xmaayan, line 24: file name is `singleKey` but class name is still `AnyPublicKey` - could we go with one name?

src/core/crypto/multiKey.ts
```diff
@@ -1,12 +1,31 @@
...
+function bitCount(byte: number) {
+  let n = byte;
+  n -= (n >> 1) & 0x55555555;
+  n = (n & 0x33333333) + ((n >> 2) & 0x33333333);
+  return (((n + (n >> 4)) & 0xf0f0f0f) * 0x1010101) >> 24;
+}
+
+/* eslint-enable no-bitwise */
+
+export class MultiKey extends Serializable {
```
0xmaayan, line 23: can we add doc comment explaining what is this class for?

src/core/crypto/singleKey.ts
0xmaayan, line 24: might worth placing singleKey, multiKey and any public key relevant files under `crypto/publicKey/` folder

src/core/crypto/singleKey.ts
```diff
@@ -0,0 +1,199 @@
...
+  isSecp256k1(): this is Secp256k1PublicKey {
+    return this.variant === AnyPublicKeyVariant.Secp256k1;
+  }
+
+  toString() {
+    return this.publicKey.toString();
+  }
+}
+
+export class AnySignature extends Serializable {
```
0xmaayan, line 129: maybe even placing any signature relevant files/classes under `crypto/signature/` folder?

src/core/signer/ed25519.ts
```diff
@@ -0,0 +1,71 @@
...
+import { AccountAddressInput } from "../accountAddress";
+import { Ed25519PrivateKey } from "../crypto";
+import type { Signer } from "./signer";
+
+export interface Ed25519SignerConstructorArgs {
+  privateKey: Ed25519PrivateKey;
+  address?: AccountAddressInput;
+}
+
+export interface Ed25519SignerFromDerivationPathArgs {
```
0xmaayan, line 13: what do you think on having all crypto types in all those files under a one single `types.ts` file? same as we have under `/transactions/`

src/core/signer/ed25519.ts
```diff
@@ -0,0 +1,71 @@
...
+  path: string;
+  mnemonic: string;
+}
+
+export class Ed25519Signer extends Ed25519Account implements Signer {
```
0xmaayan, line 18: could we add doc comments explaining what the class is for?

src/internal/account.ts
```diff
@@ -515,50 +518,45 @@ export async function getAccountOwnedObjects(args: {
...
  * we’ll need to determine the order in which we lookup the accounts. First unified
  * scheme and then legacy scheme vs first legacy scheme and then unified scheme.
  *
  */
-export async function deriveAccountFromPrivateKey(args: {
+export async function deriveSignerFromPrivateKey(args: {
   aptosConfig: AptosConfig;
-  privateKey: PrivateKey;
-}): Promise<Account> {
+  privateKey: Ed25519PrivateKey | Secp256k1PrivateKey;
```
0xmaayan, line 528: maybe create a type for that?
```
type PrivateKey = Ed25519PrivateKey | Secp256k1PrivateKey;
```

src/internal/transactionSubmission.ts
```diff
@@ -134,10 +133,7 @@ export async function buildRawTransaction(
 ): Promise<AnyRawTransaction> {
   const { aptosConfig, sender, options } = args;
 
-  let feePayerAddress;
-  if (isFeePayerTransactionInput(args)) {
-    feePayerAddress = AccountAddress.ZERO.toString();
-  }
+  const feePayerAddress = args.withFeePayer ? AccountAddress.ZERO : undefined;
```
0xmaayan, line 136: ❤️

src/internal/transactionSubmission.ts
```diff
@@ -309,48 +301,46 @@ export async function publicPackageTransaction(args: {
  */
 export async function rotateAuthKey(args: {
   aptosConfig: AptosConfig;
-  fromAccount: Account;
-  toNewPrivateKey: PrivateKey;
+  fromSigner: Ed25519Signer | SingleKeySigner;
```
0xmaayan, line 304: union type?
```
type AnySigner = Ed25519Signer | SingleKeySigner;
```

tests/e2e/ans/ans.test.ts
```diff
@@ -146,8 +146,8 @@ describe("ANS", () => {
       domainName = randomString();
       subdomainName = randomString();
 
-      alice = Account.generate();
-      bob = Account.generate();
+      alice = Signer.generate();
+      bob = Signer.generate();
```
0xmaayan, line 150: how come we can do if bob's type is `Account`...?

tests/unit/account.test.ts
```diff
@@ -26,39 +27,35 @@ describe("Account", () => {
   describe("generate", () => {
     it("should create an instance of Account with a legacy ED25519 when nothing is specified", () => {
       // Account with Legacy Ed25519 scheme
-      const edAccount = Account.generate();
-      expect(edAccount).toBeInstanceOf(Account);
+      const edAccount = Signer.generate();
+      expect(edAccount).toBeInstanceOf(Ed25519Signer);
       expect(edAccount.publicKey).toBeInstanceOf(Ed25519PublicKey);
-      expect(edAccount.signingScheme).toEqual(SigningScheme.Ed25519);
+      // expect(edAccount.signingScheme).toEqual(SigningScheme.Ed25519);
```
0xmaayan, line 33: why comment?

src/core/crypto/ed25519.ts
hardsetting: the `// endregion` comment, togther with `// region Some descript` creates a custom foldable region in most recent editors

hardsetting: Screenshot 2024-02-12 at 3 06 35 PM

src/core/crypto/index.ts
hardsetting: Mm unless I got it wrong, `AnyPublicKey` and `AnySignature` are wrappers for single public keys and signatures.
`MultiKey` will take as an input N instances of single keys, and `MultiKeySignature` (introduced in this PR) will take as an input N instances of single keys

hardsetting: So the set of all possible "accepted" public keys are:
- Ed25519 (legacy)
- MultiEd25519 (legacy)
- AnyPublicKey aka SingleKey
- MultiKey

hardsetting: notably, you can't use a Secp256k1PublicKey directly

src/core/crypto/multiEd25519.ts
hardsetting: https://github.com/aptos-labs/aptos-ts-sdk/pull/147#discussion_r1486901932

src/core/crypto/multiKey.ts
hardsetting: This is slightly different. Namely you can't use a Multikey as input oif a MultiKey.
Or I guess you could but it's easier not to support it for now 😆

tests/e2e/ans/ans.test.ts
hardsetting: `Signer` is a specialization of `Account`. It still has an associatd public key and address.
You can't really "generate" an account,  since it does not have an associated private key,
buth we can still leverage `Signer.generate`

tests/unit/account.test.ts
hardsetting: good catch, just brought it back

hardsetting:
> lets put all the changes this PR introduces in a changelog?

What's the procedure for this?

src/core/crypto/singleKey.ts
hardsetting: We can either choose to "index" by type (e.g. public key, signature) or by scheme (e.g. ed25519, singleKey, multiKey), which are mostly equivalent.

Do you have an argument for switching to indexing by type?

src/core/crypto/singleKey.ts
hardsetting: Same as above

src/core/signer/ed25519.ts
hardsetting: I think it would be confusing. These are input args specifically for an Ed25519Signer for example.
I would put something under `types` if it's meant to be shared across multiple files.

0xmaayan (commented):
Overall LGTM.

Some missing things
1. Function and input args doc on each function
2. Doc/explanation for each type/interface/function
3. CHANGELOG updates with changes (types/classes/methods + indicating breaking changes)

I highly recommend looking for more people input on these changes

examples/typescript/swap.ts
```diff
@@ -206,7 +207,7 @@ const example = async () => {
   const aptos = new Aptos(aptosConfig);
   // Create three accounts
   const swapAddress = AccountAddress.from(process.argv[2]);
-  const admin = Account.fromPrivateKeyAndAddress({
+  const admin = Signer.fromPrivateKey({
```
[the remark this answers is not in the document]
hardsetting: The address is optional. This is true when constructing any account or signer.
By default, the address is derived by the public key, but if the account is rotated, the user needs to be able to pass the address.

It personally makes more sense to me to have a single `fromPrivateKey` which optionally takes the address, as opposed to having `fromPrivateKey` and `fromPrivateKeyAndAddress`

src/api/transactionSubmission/simulate.ts
```diff
@@ -51,10 +51,10 @@ export class Simulate {
    */
   @ValidateFeePayerDataOnSimulation
   async multiAgent(args: {
-    signerPublicKey: PublicKey;
+    signerPublicKey: PublicKeyInput;
     transaction: AnyRawTransaction;
-    secondarySignersPublicKeys: Array<PublicKey>;
-    feePayerPublicKey?: PublicKey;
+    secondarySignersPublicKeys: PublicKeyInput[];
```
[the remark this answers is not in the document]
hardsetting: What is the SDK syntax?
Happy to change this, but I'm curious about the reasons for using `Array` instead of `[]`

src/api/transactionSubmission/sign.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-import { Account } from "../../core";
+import { Signer } from "../../core";
```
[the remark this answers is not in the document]
hardsetting: removed

src/core/signer/signer.ts
```diff
@@ -0,0 +1,146 @@
+import type { AccountAuthenticator } from "../../transactions/authenticator/account";
+import { HexInput, SigningSchemeInput } from "../../types";
+import { Account } from "../account";
+import type { AccountAddressInput } from "../accountAddress";
+import { Ed25519PrivateKey } from "../crypto";
+import { Ed25519Signer } from "./ed25519";
+import { type PrivateKeyInput, SingleKeySigner } from "./singleKey";
+
+interface CreateEd25519SignerFromPrivateKeyArgs {
```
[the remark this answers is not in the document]
hardsetting: Exporting everything.
I think the names are pretty self-explanatory though.

src/api/transactionSubmission/simulate.ts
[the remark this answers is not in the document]
0xmaayan: dont think there was a specific reason, we just use `Array<>` instead of `[]` all over the SDK so might be better to conform with that

hardsetting:
Closing in favor of #294

PR #148, opened 2023-10-31
Update AuthenticationKey to extend from Serializable

0xjinn:
### Description

Update AuthKey to extend from Serializable.

### Test Plan

pnpm test

### DIFF
--- a/tests/unit/authenticationKey.test.ts
+++ b/tests/unit/authenticationKey.test.ts
@@ -91,4 +91,25 @@ describe("AuthenticationKey", () => {
       "No supported authentication scheme for public key",
     );
   });
+
+  it("should serialize correctly", () => {
+    const authKey = new AuthenticationKey({ data: ed25519.authKey });
+    const serializer = new Serializer();
+    authKey.serialize(serializer);
+    const expected = new Uint8Array([
+      151, 140, 33, 57, 144, 196, 131, 61, 247, 21, 72, 223, 124, 228, 157, 84, 199, 89, 214, 182, 217, 50, 222, 34,
+      178, 77, 86, 6, 11, 122, 242, 170,
+    ]);
+    expect(serializer.toUint8Array()).toEqual(expected);
+  });
+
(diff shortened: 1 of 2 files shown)


0xjinn:
* main
  * **PR #148** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-ts-sdk/148" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-ts-sdk/148?utm_source=stack-comment).

0xmaayan (approved):

tests/unit/authenticationKey.test.ts
```diff
@@ -91,4 +91,13 @@ describe("AuthenticationKey", () => {
...
   });
+
+  it("should serialize and deserialize correctly", () => {
+    const authKey = new AuthenticationKey({ data: ed25519.authKey });
+    const serializer = new Serializer();
+    authKey.serialize(serializer);
+    const deserializer = new Deserializer(serializer.toUint8Array());
+    const authKeyDeserialized = AuthenticationKey.deserialize(deserializer);
+    expect(authKeyDeserialized.data.toString()).toEqual(ed25519.authKey);
+  });
```
gregnazario, line 102: FYI for a test like this, you should ensure that the data matches what you expect in bytes if it's supposed to be consumed by anything else.

You can serialize and deserialize something wrong, as long as it's symmetric :)

xbtmatt (approved):
I think just add a more meaningful test (should be easy since it's a simple auth key) and it lgtm

tests/unit/authenticationKey.test.ts
0xjinn: Ah I kept forgetting about this. Will update soon

0xjinn:
Update: Split the test int Serialize and Deserialize separately

PR #149, opened 2023-10-31
[examples] Add examples to the CI, and ensure linting and formatting work correctly

gregnazario:
### Description
This sets up all examples to run in one command (except the two that are broken).
Additionally, linting now works properly on the examples as well (and has been cleaned up).

### Test Plan
All running examples worked correctly for me when running `pnpm test`, this PR tests the CI of running the examples.

### DIFF
--- a/src/utils/apiEndpoints.ts
+++ b/src/utils/apiEndpoints.ts
@@ -34,3 +34,11 @@ export const NetworkToChainId: Record<string, number> = {
   mainnet: 1,
   testnet: 2,
 };
+
+export const NetworkToNetworkName: Record<string, Network> = {
+  mainnet: Network.MAINNET,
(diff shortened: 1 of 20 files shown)


src/utils/apiEndpoints.ts
```diff
@@ -34,3 +34,20 @@ export const NetworkToChainId: Record<string, number> = {
   mainnet: 1,
   testnet: 2,
 };
+
+export function toNetwork(network: string): Network | undefined {
```
0xmaayan, line 38: is there a reason this needs to be a function and not enum?

gregnazario: It was just that I needed string -> network, let me fix that

0xmaayan (approved):

PR #152, opened 2023-10-31
[remoteAbi] Add transaction argument tests for simple args

gregnazario:
### Description
This mixes simple and non-simple args with the already existing tests to ensure that transactions are handled properly.

This also makes the string parsing for addresses less strict

### Test Plan
This adds tests specifically to test remote ABI inputs to ensure they work properly.

### DIFF
--- a/src/transactions/transactionBuilder/remoteAbi.ts
+++ b/src/transactions/transactionBuilder/remoteAbi.ts
@@ -150,7 +150,7 @@ function parseArg(
   // TODO: support uint8array?
   if (param.isAddress()) {
     if (isString(arg)) {
-      return AccountAddress.fromString(arg);
+      return AccountAddress.fromStringRelaxed(arg);
     }
     throwTypeMismatch("string", position);
   }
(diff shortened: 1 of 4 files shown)


src/transactions/transactionBuilder/remoteAbi.ts
```diff
@@ -150,7 +150,7 @@ function parseArg(
   // TODO: support uint8array?
   if (param.isAddress()) {
     if (isString(arg)) {
-      return AccountAddress.fromString(arg);
+      return AccountAddress.fromStringRelaxed(arg);
```
0xmaayan, line 153: why?

0xmaayan (approved):

src/transactions/transactionBuilder/remoteAbi.ts
gregnazario: Basically, if you pull something out of the API elsewhere and we don't type it for you, it could be missing the 0's.  If you're pulling it out of one system into the other, you want it to be more permissive.

In this case, that occurred a couple times to me when accounts were generated.

PR #155, opened 2023-10-31
[Documentation] Fixing typos in documentation

xbtmatt:
### Description
Adding factory methods for MoveVector for AccountAddress and MoveOption:

```typescript
MoveVector.AccountAddress
MoveOption.AccountAddress
```

This is from an old PR

### Test Plan
added unit tests
`pnpm jest`

### DIFF
--- a/src/bcs/serializable/moveStructs.ts
+++ b/src/bcs/serializable/moveStructs.ts
@@ -7,6 +7,7 @@ import { Deserializable, Deserializer } from "../deserializer";
 import { AnyNumber, HexInput, ScriptTransactionArgumentVariants } from "../../types";
 import { Hex } from "../../core/hex";
 import { EntryFunctionArgument, TransactionArgument } from "../../transactions/instances/transactionArgument";
+import { AccountAddress } from "../../core/accountAddress";
 
 /**
  * This class is the Aptos Typescript SDK representation of a Move `vector<T>`,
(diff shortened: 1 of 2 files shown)


src/bcs/serializable/moveStructs.ts
```diff
@@ -432,6 +447,32 @@ export class MoveOption<T extends Serializable & EntryFunctionArgument>
...
+   */
+  static AccountAddress(value?: HexInput | AccountAddress | null): MoveOption<AccountAddress> {
+    let accountAddress: AccountAddress | undefined;
+    if (value !== null && value !== undefined) {
+      if (value instanceof AccountAddress) {
+        accountAddress = value;
+      } else {
+        accountAddress = AccountAddress.fromHexInputRelaxed(value);
+      }
+    } else {
```
gregnazario, line 470: We should merge this in with this change https://github.com/aptos-labs/aptos-ts-sdk/pull/145 but it keeps bloating more and more for me every time I'm fixing one thing, so I'm trying to get it working soon.

src/bcs/serializable/moveStructs.ts
```diff
@@ -184,13 +180,31 @@ export class MoveVector<T extends Serializable & EntryFunctionArgument>
...
+   * @params values: an array of `HexInput | AccountAddress`
+   * @returns a `MoveVector<AccountAddress>`
+   */
+  static AccountAddress(values: Array<HexInput | AccountAddress>): MoveVector<AccountAddress> {
+    const accountAddresses = values.map((v) => {
+      if (v instanceof AccountAddress) {
+        return v;
+      }
+      return AccountAddress.fromHexInputRelaxed(v);
+    });
```
gregnazario, line 204: This seems like it could be put in a helper function.

gregnazario (approved):
Not really opposed to adding these extra factory functions, but it does really just help you if you're doing manual BCS encoding right?

xbtmatt:
@gregnazario Yea, tbh I was just adding because it was ready to be merged- but ultimately I haven't had a need for these functions so I don't think we need them.

This PR can just be an update for the docs, I don't want to make things needlessly more bloated/complex, so I removed the factory methods

PR #156, opened 2023-11-01
Add rotateAuthKey

0xjinn:
### Description

This PR creates an new api method rotateAuthKey

Note:
1. I place it under /api folder only and not /internal, because of dependency cycle
2. This only supports legacy ed25519, the move side does not support Single Signer or Secp.

### Test Plan

pnpm test

### DIFF
--- a/src/api/transactionSubmission.ts
+++ b/src/api/transactionSubmission.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { AptosConfig } from "./aptosConfig";
-import { Account } from "../core";
+import { Account, PrivateKey } from "../core";
 import { AccountAuthenticator } from "../transactions/authenticator/account";
 import {
   AnyRawTransaction,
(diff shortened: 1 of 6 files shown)


src/api/transactionSubmission.ts
```diff
@@ -189,4 +190,16 @@ export class TransactionSubmission {
   }): Promise<InputSingleSignerTransaction> {
     return publicPackageTransaction({ aptosConfig: this.config, ...args });
   }
+
+  /**
+   * Rotate an account's auth key. After rotation, only the new private key can be used to sign txns for
```
0xmaayan, line 195: lets add a comment on only ed25519 support ?

0xmaayan: also, a link to something explains what is key rotation would be nice

src/transactions/instances/rotationProofChallenge.ts
```diff
@@ -0,0 +1,51 @@
...
+import { AnyNumber } from "../../types";
+import { PublicKey } from "../../core/crypto/asymmetricCrypto";
+import { MoveString, MoveVector, U64, U8 } from "../../bcs";
+
+/**
+ * Representation of the challenge which is needed to sign by owner of the account
+ * to rotate the authentication key.
+ */
+export class RotationProofChallenge extends Serializable {
+  public readonly accountAddress: AccountAddress = AccountAddress.ONE;
```
0xmaayan, line 15: lets document each property

0xmaayan (approved):

PR #157, opened 2023-11-01
[Tests] Added script functions to argument types test suite

xbtmatt:
### Description
Updated the test suite to use scripts

The sender/authenticator type is already validated in transaction submission, so in order to keep this test focused, I just added scripts instead of adding all the different signer types.

## Test suite

TODO:
- [x] Add 1 - n signers for the functions
- [x] Verify the order of signers in the contract with assertions

### Argument types
- [x] basic argument types
- [x] vector argument types
- [x] option argument types
- [x] deeply nested complex types

### Number of signers
 - [x] One signer
 - [x] Multiple signers

### TODO:
- view function of all the argument types listed above
- [x] type tags, including various nested generics

NOTE: Also added a small fix to the script function vector `U8` serialize function. This is only needed for `U8` because script functions don't accept any other vector type.

### Test Plan
```typescript
pnpm jest transactionArguments.test.ts --verbose
```


tests/move/arguments/sources/multi_signer.move
```diff
@@ -0,0 +1,49 @@
+script {
+    use std::signer;
+    use std::object::{Object};
+    use std::string::{String};
+    use transaction_arguments::tx_args_module::{Self, EmptyResource};
+
+    fun main(
```
gregnazario, line 7: You can actually name this any name, and that'll be the name of the script bytecode file.

tests/move/arguments/sources/multi_signer.move
```diff
@@ -0,0 +1,49 @@
...
+        account_1: &signer,
+        account_2: &signer,
+        account_3: &signer,
+        account_4: &signer,
```
gregnazario, line 11: Can you mix `&signer` and `signer` here for good measure?

tests/move/arguments/sources/main.move
```diff
@@ -228,15 +234,19 @@ module transaction_arguments::tx_args_module {
...
     }
 
-    public entry fun public_arguments_one_signer(
+    public entry fun public_arguments_multiple_signers(
         account_1: &signer,
-        signer_address: address,
+        account_2: &signer,
+        account_3: &signer,
+        account_4: &signer,
+        account_5: &signer,
```
gregnazario, line 248: similarly would be good to mix &signer and signer

tests/move/arguments/sources/main.move
```diff
@@ -157,43 +162,44 @@ module transaction_arguments::tx_args_module {
         // deeply_nested_2: vector<Option<vector<vector<vector<Option<u256>>>>>>, // TODO
         // deeply_nested_3: vector<Option<vector<vector<vector<Option<Object<EmptyResource>>>>>>>, // TODO
```
gregnazario, line 163: It would be good to test vector<vector<value>> especially vector<vector<u8>> and vector<vector<u64>>

src/bcs/serializable/moveStructs.ts
```diff
@@ -68,9 +68,10 @@ export class MoveVector<T extends Serializable & EntryFunctionArgument>
    */
   serializeForScriptFunction(serializer: Serializer): void {
     // runtime check to ensure that you can't serialize anything other than vector<u8>
-    // TODO: consider adding support for MoveString later?
     const isU8 = this.values[0] instanceof U8;
-    if (!isU8) {
+    // if the inner array is length 0, we can't check the type because it has no instance, so we assume it's a u8
+    // it may not be, but we don't care because it's serialized as `0x0` regardless
+    if (!isU8 && this.values[0] !== undefined) {
```
gregnazario, line 74: Actually, sometimes 0 length is serialized as `0x` with no value in the API.  Unsure why

xbtmatt: ooh yes, you're actually right, I'll say `either way it's serialized to a single byte value: 0`

It's `0x` because they're indicating that it's a byte vector that's just being represented as a string. There's no values after it because it's a 0-length array (hence the serialized `0`)

0xmaayan (commented):
overall lgtm, makes me think we need to improve our testing framework and consolidate all helpers/testing objects/etc.

tests/move/arguments/sources/main.move
xbtmatt:
```rust
entry fun complex_arguments(
    deeply_nested_1: vector<vector<u8>>,
    deeply_nested_2: vector<vector<String>>,
    deeply_nested_3: vector<Option<vector<String>>>,
    deeply_nested_4: vector<vector<Option<vector<String>>>>,
) {
```

Done!

tests/e2e/transaction/helper.ts
```diff
@@ -39,11 +39,36 @@ export async function publishPackage(
...
+  await aptos.fundAccount({ accountAddress: firstAccount.accountAddress.toString(), amount: FUND_AMOUNT });
+  // Fund again because these txns can be expensive
+  await aptos.fundAccount({ accountAddress: firstAccount.accountAddress.toString(), amount: FUND_AMOUNT });
+  const addressesRemaining = accounts.slice(1).map((account) => account.accountAddress);
+  const amountToSend = Math.floor((FUND_AMOUNT * 2) / accounts.length);
+  // Send APT to the rest
+  const transaction = await aptos.generateTransaction({
+    sender: firstAccount.accountAddress.toString(),
+    data: {
+      function: `0x${1}::aptos_account::batch_transfer`,
```
gregnazario, line 53: Lol why like this?  If this is doing a pattern match that makes you do something like this, we need to fix it and remove the pattern match

tests/e2e/transaction/helper.ts
```diff
@@ -39,11 +39,36 @@ export async function publishPackage(
...
+      functionArguments: [
+        new MoveVector(addressesRemaining),
+        MoveVector.U64(addressesRemaining.map(() => amountToSend)),
```
gregnazario, line 56: Why map if it's exactly one?

tests/e2e/transaction/helper.ts
```diff
@@ -39,11 +39,36 @@ export async function publishPackage(
...
-  for (const account of accounts) {
-    // eslint-disable-next-line no-await-in-loop
-    await aptos.fundAccount({ accountAddress: account.accountAddress.toString(), amount: FUND_AMOUNT });
-  }
+  // Fund first account
+  const firstAccount = accounts[0];
+  await aptos.fundAccount({ accountAddress: firstAccount.accountAddress.toString(), amount: FUND_AMOUNT });
+  // Fund again because these txns can be expensive
+  await aptos.fundAccount({ accountAddress: firstAccount.accountAddress.toString(), amount: FUND_AMOUNT });
+  const addressesRemaining = accounts.slice(1).map((account) => account.accountAddress);
```
gregnazario, line 47: Why not just `accounts[1]`

gregnazario (approved):

tests/e2e/transaction/helper.ts
xbtmatt: I'll put a comment explaining it so it's easier to read, but it's sending the coins from `account[0]` out to `accounts[1..n]` not just `accounts[1]`

tests/e2e/transaction/helper.ts
xbtmatt: Lol I have no idea why I did this- it's fine with just the string, I updated it

tests/e2e/transaction/helper.ts
xbtmatt: if `end` in `Array.slice(start?: number, end?: number)` is undefined then it returns the slice from start to the end

PR #158, opened 2023-11-01
Replace all instances of `rawTransaction: Uint8Array` with an actual `RawTransaction`

xbtmatt:
### Description
After working with the wallet adapter more, I realized my initial motivation for suggesting @0xmaayan use a `Uint8Array` was misled and premature, because we can perform this conversion at the wallet adapter layer if necessary.

Until it's necessary, it is ideally more strongly typed as an actual RawTransaction instance.

1. This saves us an unnecessary round of serialization/deserialization in `deriveTransaction`
2. Increases clarity, because it's confusing knowing what the `Uint8Array` bytes really are
3. We can very easily just perform this conversion at the wallet adapter layer when it's necessary

After trying to build transactions/come up with a more intuitive flow, I realized how premature converting to bcs bytes was, and that's the motivation for this PR.

### Test Plan
```typescript
pnpm jest
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 - Support derive account from private key `Account.fromPrivateKey()`
 - Derive account from derivation path secp256k1 support
 - Default Account generation to Legacy Ed25519
+- Remove unnecessary pre-emptive serialization of the field `rawTransaction: Uint8Array` by replacing it with the unserialized `rawTransaction: RawTransaction` class
 
 ## 0.0.3 (2023-10-31)
 
--- a/src/api/transactionSubmission.ts
+++ b/src/api/transactionSubmission.ts
@@ -77,10 +77,10 @@ export class TransactionSubmission {
    * }
    * ```
    *
-   * @return A raw transaction type (note that it holds the raw transaction as a bcs serialized data)
+   * @return An instance of a RawTransaction, plus optional secondary/fee payer addresses
    * ```
    * {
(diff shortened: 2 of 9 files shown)


0xmaayan:
I remember another reason to use bcs and not json is when servers need/want to send the transaction to another server, json is pretty sucks. Is that not a concern anymore?

xbtmatt:
@0xmaayan correct, it's not a concern because you can always do the conversion right before transport across some network layer for yourself.

I suppose we could come up with a different type for it, but I still kind of think it was premature, since most cases where you're handling this, the wallet adapter can/will just do it for you

0xmaayan:
> @0xmaayan correct, it's not a concern because you can always do the conversion right before transport across some network layer for yourself.
> 
> I suppose we could come up with a different type for it, but I still kind of think it was premature, since most cases where you're handling this, the wallet adapter can/will just do it for you

so the idea it to rely on the adapter to do it for you? what if you use the sdk in a node project, and therefore dont use the adapter? what if the dapp doesnt use the adapter?

xbtmatt:
@0xmaayan hmm I suppose they could run into issues- but they can ultimately just serialize it themselves if they need to use a custom transmission function to a network. My initial thoughts for why we should do it were that the wallet adapter would run into issues if the user sends a `RawTransaction`, but we can always just serialize it before that. Tbh I'll put this in draft and we should discuss it/the flow again- sorry for prematurely putting this out there

0xmaayan (approved):

0xmaayan:
Please update CHANGELOG

gregnazario (approved):

PR #161, opened 2023-11-02
Optional Fee Payer

0xjinn:
### Description

1. Added "hasSponsor" field to all `InputGenerateTransactionData`. If set to `true`, feepayer address will be overrided to "0x0".
2. Fee payer will have to use "signTransactionAsPayer" to sign the transaction.

### Test Plan

```
Test Suites: 38 passed, 38 total
Tests:       1 skipped, 450 passed, 451 total
Snapshots:   0 total
Time:        24.384 s, estimated 25 s
```

### DIFF
--- a/src/api/transactionSubmission.ts
+++ b/src/api/transactionSubmission.ts
@@ -95,6 +95,7 @@ export class TransactionSubmission {
    *
    * @param args.signer The signer account to sign the transaction
    * @param args.transaction An instance of a RawTransaction, plus optional secondary/fee payer addresses
+   * @param args.asFeePayer optional. If the signer is also the fee payer
    * ```
    * {
    *  rawTransaction: RawTransaction,
@@ -106,30 +107,15 @@ export class TransactionSubmission {
    * @return The signer AccountAuthenticator
    */
   /* eslint-disable class-methods-use-this */
-  signTransaction(args: { signer: Account; transaction: AnyRawTransaction }): AccountAuthenticator {
-    return signTransaction({ ...args });
-  }
-
-  /**
-   * Sign a transaction that can later be submitted to chain, as a fee payer. The signer who signs the transaction
-   * will be the one who pays the gas
-   *
-   * @param args.signer The signer or fee payer
-   * @param args.transaction A raw transaction type (note that it holds the raw transaction as a bcs serialized data)
-   * ```
-   * {
-   *  rawTransaction: Uint8Array,
-   *  secondarySignerAddresses? : Array<AccountAddress>,
-   *  feePayerAddress?: AccountAddress
-   * }
-   * ```
-   *
-   * @return The signer AccountAuthenticator
-   */
-  /* eslint-disable class-methods-use-this */
-  signTransactionAsPayer(args: { signer: Account; transaction: AnyRawTransaction }): AccountAuthenticator {
-    const { signer, transaction } = args;
-    transaction.feePayerAddress = signer.accountAddress;
+  signTransaction(args: {
+    signer: Account;
+    transaction: AnyRawTransaction;
+    asFeePayer?: boolean;
+  }): AccountAuthenticator {
+    const { signer, transaction, asFeePayer } = args;
+    if (asFeePayer) {
+      transaction.feePayerAddress = signer.accountAddress;
+    }
     return signTransaction({
       signer,
       transaction,
--- a/src/api/transactionSubmission.ts
+++ b/src/api/transactionSubmission.ts
@@ -111,6 +111,32 @@ export class TransactionSubmission {
   }
 
   /**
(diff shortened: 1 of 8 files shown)


xbtmatt:
I love `hasSponsor` as a concept, but in the interest of keeping things consistent I think we should probably do `withFeePayer` or `hasFeePayer`

Although I'm partial to `withFeePayer` because `hasFeePayer` suggests it already has a fee payer, and it technically doesn't yet.

The clearest would be `willHaveFeePayer` but maybe that's too wordy

xbtmatt:
Also another consistency thing:

`signTransactionAsFeePayer` or even `signAsFeePayer` (since there's no other thing you'll sign as a fee payer)

0xjinn:
@xbtmatt 

> I love `hasSponsor` as a concept, but in the interest of keeping things consistent I think we should probably do `withFeePayer` or `hasFeePayer`
> 
> Although I'm partial to `withFeePayer` because `hasFeePayer` suggests it already has a fee payer, and it technically doesn't yet.
> 
> The clearest would be `willHaveFeePayer` but maybe that's too wordy
> 
> Personally I'm a fan of wordy names since it's an sdk and it generally makes the devex better as long as it isn't an eyesore to look at (camelCase kinda sucks lol)

Yeah good call. I used `hasSponsor` mainly because of the concept "Sponsored Transaction". TBH I wonder is the official name, is it "Fee Payer Transaction", or "Sponsor Transaction"? Maybe @davidiw would know :)

src/api/transactionSubmission.ts
```diff
@@ -110,6 +110,32 @@ export class TransactionSubmission {
...
+   *  rawTransaction: Uint8Array,
+   *  secondarySignerAddresses? : Array<AccountAddress>,
+   *  feePayerAddress?: AccountAddress
+   * }
+   * ```
+   *
+   * @return The signer AccountAuthenticator
+   */
+  /* eslint-disable class-methods-use-this */
+  signTransactionAsPayer(args: { signer: Account; transaction: AnyRawTransaction }): AccountAuthenticator {
```
0xmaayan, line 130: I feel this whole function can simply be a boolean arguments

```
signTransaction(args: { signer: Account; transaction: AnyRawTransaction, asSponsor?:true }): AccountAuthenticator {
if(asSponsor){
 transaction.feePayerAddress = signer.accountAddress;
}
    return signTransaction({ ...args });
  }
```

src/transactions/types.ts
```diff
@@ -267,7 +267,7 @@ export type InputSimulateTransactionOptions = {
  */
 export interface InputGenerateSingleSignerRawTransactionData {
   sender: AccountAddressInput;
-  feePayerAddress?: undefined;
+  hasSponsor?: undefined | false;
```
0xmaayan, line 270: not sure you need the `false`? the intention here is that `hasSponsor` is always undefined right? then I think we shouldn't expect/let them set it to false

0xmaayan (changes requested):
boolean argument instead of a new function

xbtmatt:
@0xjinn a bit ago we decided to go with fee payer in code and sponsored in non-code contexts, because changing the field names was backwards incompatibility already for some indexer stuff 😅

src/transactions/types.ts
0xjinn: Hm but if user decided to make it clear, and set the `hasSponsor` to false. I would expect that should also be a viable option.

src/api/transactionSubmission.ts
0xjinn: Yeah I was thinking about this vs a new function. I wonder what everyone's preference are. We can sync up real quick on monday standup.

src/transactions/types.ts
0xmaayan: why would we want user to specify it if it doesnt make sense in this context as it is always undefined? even if they specify false.

0xjinn: Okay I think I see what you mean. Since the Fee payer transaction data is marked as specifically, `hasSponsor: true`, we dont need to let user to specify the false, and its always undefine for other transaction data. 

Will update.

tests/e2e/transaction/helper.ts
```diff
@@ -116,15 +116,17 @@ export const rawTransactionMultiAgentHelper = async (
   secondarySignerAccounts: Array<Account>,
   feePayerAccount?: Account,
```
0xmaayan, line 117: should we change it to be a boolean flag?

0xmaayan (approved):
lgtm. lets also update CHANGELOG

tests/e2e/transaction/helper.ts
0xjinn: Nah lets not do that, we still need the feed payer account to actually sign the transaction.

PR #162, opened 2023-11-02
[Easy] Clean up typetag snake case to camelcase

0xjinn:
### Description

I think there are more. I think there is more, will fix it as I find them.

### DIFF
--- a/src/transactions/typeTag/index.ts
+++ b/src/transactions/typeTag/index.ts
@@ -291,11 +291,11 @@ export class TypeTagStruct extends TypeTag {
   toString(): `0x${string}::${string}::${string}` {
     // Collect type args and add it if there are any
     let typePredicate = "";
-    if (this.value.type_args.length > 0) {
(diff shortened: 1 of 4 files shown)


src/transactions/typeTag/index.ts
```diff
@@ -338,25 +338,25 @@ export class TypeTagStruct extends TypeTag {
 export class StructTag extends Serializable {
   public readonly address: AccountAddress;
 
-  public readonly module_name: Identifier;
+  public readonly moduleName: Identifier;
 
   public readonly name: Identifier;
 
-  public readonly type_args: Array<TypeTag>;
+  public readonly typeArgs: Array<TypeTag>;
```
gregnazario, line 345: depends if this is deserialized from JSON / it's used directly there.  If it's not then we can change it.

Do we use BCS for all of our communication to the node?

gregnazario (approved):

PR #165, opened 2023-11-02
Create input submit transaction interface type

0xmaayan:
### Description
- moving the input arguments into an interface
- flatten interface, removed unneeded `secondarySignerAuthenticators` layer

### DIFF
--- a/src/transactions/types.ts
+++ b/src/transactions/types.ts
@@ -17,6 +17,7 @@ import {
 } from "./instances";
 import { AnyNumber, HexInput, MoveFunctionGenericTypeParam, MoveStructType } from "../types";
 import { TypeTag } from "./typeTag";
+import { AccountAuthenticator } from "./authenticator/account";
 
 /**
  * Entry function arguments to be used when building a raw transaction using remote ABI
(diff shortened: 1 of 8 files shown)


src/transactions/types.ts
```diff
@@ -301,3 +302,15 @@ export type InputGenerateTransactionData =
...
+/**
+ * Interface that holds the user data input when submitting a transaction
+ */
+export interface InputSubmitTransactionData {
+  transaction: AnyRawTransaction;
+  senderAuthenticator: AccountAuthenticator;
+  secondarySignerAuthenticators?: {
+    feePayerAuthenticator?: AccountAuthenticator;
+    additionalSignersAuthenticators?: Array<AccountAuthenticator>;
+  };
```
gregnazario, line 315: This is a little weird now that i'm looking at it that, secondary signers, contains fee payer and additional signers.

We may want to flatten this to two top level optionals.  (Or rename the outside additional, and the internal one secondary which matches the API)

gregnazario (approved):

src/transactions/types.ts
0xmaayan: agreed. ill flatten it

gregnazario (approved):

PR #166, opened 2023-11-02
Add indexerTimeoutSecs Config

0xjinn:
### Description

Add a new config field to waitForTransaction `indexerTimeoutSecs`. This allow user to modify the timeout for indexer sync.

### DIFF
--- a/src/internal/transaction.ts
+++ b/src/internal/transaction.ts
@@ -19,7 +19,7 @@ import {
   type TransactionResponse,
   WaitForTransactionOptions,
 } from "../types";
-import { DEFAULT_TXN_TIMEOUT_SEC } from "../utils/const";
+import { DEFAULT_INDEXER_SYNC_TIMEOUT_SEC, DEFAULT_TXN_TIMEOUT_SEC } from "../utils/const";
 import { sleep } from "../utils/helpers";
 import { memoizeAsync } from "../utils/memoize";
 import { getIndexerLastSuccessVersion } from "./general";
@@ -105,6 +105,7 @@ export async function waitForTransaction(args: {
   const timeoutSecs = options?.timeoutSecs ?? DEFAULT_TXN_TIMEOUT_SEC;
   const checkSuccess = options?.checkSuccess ?? true;
   const indexerVersionCheck = options?.indexerVersionCheck ?? true;
+  const indexerTimeoutSecs = options?.indexerTimeoutSecs ?? DEFAULT_INDEXER_SYNC_TIMEOUT_SEC;
 
   let isPending = true;
   let timeElapsed = 0;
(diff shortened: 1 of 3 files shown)


src/internal/transaction.ts
```diff
@@ -105,6 +105,7 @@ export async function waitForTransaction(args: {
   const timeoutSecs = options?.timeoutSecs ?? DEFAULT_TXN_TIMEOUT_SEC;
   const checkSuccess = options?.checkSuccess ?? true;
   const indexerVersionCheck = options?.indexerVersionCheck ?? true;
```
0xjinn, line 107: I wonder if we should update this and default to `false`. I feel like, only dev that relies on indexer API will probably want. But if they are only sending request to full node api, they probably want to have this off.

Thought? cc: @0xmaayan @gregnazario

0xmaayan: hmm most of the sdk queries use indexer, so if a sdk developer uses the sdk to submit transactions and then fetch data, they/sdk should wait for indexer.

0xmaayan (approved):

gregnazario (commented):
I think that adding another timeout just gets into the level of too much configurability / probably keep it simple.

If you're a developer, you don't really care what you're waiting for, you just want it to be ready in X seconds.

Separating the two, requires extra thought and process to determine why one or the other, and plenty of room for misconfiguration.

src/internal/transaction.ts
0xjinn: > Only annoying thing is... if it's by default on and the indexer is delayed / broken, the dapps will completely not work.
> 
> It's a strange tradeoff based on what your dapp does.

Yeah, I am thinking about this trade off as well. Because I would hate to sdk to be broken b/c of indexer. Maybe we should just always return, but have error field saying that the indexer failed to sync?

For example:
const result = aptos.waitForTransaction();

and result looks like following:
{
  indexerSyncError: "timeout"
  ...
}


Or we just relies on the developer to wrap `waitForTransaction` in a try-catch

0xjinn:
> I think that adding another timeout just gets into the level of too much configurability / probably keep it simple.
> 
> If you're a developer, you don't really care what you're waiting for, you just want it to be ready in X seconds.
> 
> Separating the two, requires extra thought and process to determine why one or the other, and plenty of room for misconfiguration.

I think thats the reason we put it in a optional field `waitForTransactionOptions`. So the developer does not have to care, unless they really want to 😄

src/types/index.ts
```diff
@@ -996,6 +996,7 @@ export type WaitForTransactionOptions = {
   timeoutSecs?: number;
   checkSuccess?: boolean;
   indexerVersionCheck?: boolean;
+  indexerTimeoutSecs?: number;
```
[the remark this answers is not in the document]
xbtmatt: I agree- let's not rely too much on configuration options, because too many will be overwhelming and difficult for people to find. This is more true when the preferred config/option is going to be very consistent (i.e., 95%+ of developers will expect it by default)

0xjinn:
Closing this PR. 

Will work on a PR that fix this problem by implementing option 1 and 2 here -[wait for indexer](https://github.com/aptos-labs/aptos-ts-sdk/pull/161)

PR #167, opened 2023-11-03
[deps] Update dependencies to latest versions / fix examples

gregnazario:
### Description
1. I've fixed all the examples, and now they all run including the javascript ones.  We will likely want to at least have one ESM example again, just to show how a proper setup would look like for it.
2. All dependencies are updated to the latest including Typescript!  We may want to move some to peerDependencies and allow people to choose their own versions to not deal with conflicts.
3. `param?: undefined` is gone!  The newer versions of typescript are much more strict / didn't like this, so I've properly changed it to a type based overload.  It's additionally been tested in Javascript to work!
4. GraphQL had to be regenerated with the updated generation dependency

### Test Plan
pnpm test, for both examples and the unit tests, CI should also check them as well


0xjinn:
Looks like there is alot of timeouts with local indexer?

gregnazario:
> Looks like there is alot of timeouts with local indexer?

That's been happening to me a bunch in CI @banool

src/client/get.ts
```diff
@@ -47,7 +47,9 @@ export type GetAptosRequestOptions = Omit<GetRequestOptions, "type">;
  * @param options GetRequestOptions
  * @returns
  */
-export async function get<Req, Res>(options: GetRequestOptions): Promise<AptosResponse<Req, Res>> {
+export async function get<Req extends {}, Res extends {}>(
```
0xmaayan, line 50: why do you add extends everywhere..?

gregnazario: The generics have to be objects, rather than just any input. There's stronger typing in future versions of some of these

0xmaayan (commented):
lots of files. can we test it locally to make sure it works with nextjs/node/web app? can link a local build.

gregnazario:
> lots of files. can we test it locally to make sure it works with nextjs/node/web app? can link a local build.

For node, the examples are now automated in CI.

NextJS / webapp, I'll have to build one

0xmaayan:
> > lots of files. can we test it locally to make sure it works with nextjs/node/web app? can link a local build.
> 
> For node, the examples are now automated in CI.
> 
> NextJS / webapp, I'll have to build one

Great. I am also curios what would happen for project that use different/lower ts versions.

0xmaayan (approved):
I tested it doesnt break nextjs/node/web app with a very basic indexer/fullnode queries. i.e just run the app and make those queries.

Also tested with a local wallet adapter demo app with sdk v2 support.

Let's roll with it but keep an eye if issues are coming

PR #170, opened 2023-11-03
Adds additional ANS contract APIs

kaw2k:
### Description
This adds the remaining P0 contract APIs for ANS. There are a few indexer queries to pull in after this PR gets merged. The functions include:
* Registering a subdomain via `registerName`
* Setting and Getting primary name for an account `getPrimaryName` and `setPrimaryName`
* Setting and Getting target addresses for a name `getTargetAddress` and `setTargetAddress`
* Renewing a domain with `renewDomain`. **Note**, this is `Domain` and not `Name` as subdomains are not renewable.

**Note**: I am currently skipping one test for name renewal. I need to work with someone on the contract side to see what is going on. I think I might be missing one config variable when setting up the contract.

### Test Plan
E2E test suite

### Related Links
_na_

### DIFF
--- a/src/api/ans.ts
+++ b/src/api/ans.ts
@@ -1,9 +1,17 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-import { RegisterNameParameters, getExpiration, getOwnerAddress, registerName } from "../internal/ans";
-import { InputSingleSignerTransaction } from "../transactions/types";
-import { MoveAddressType } from "../types";
+import { Account } from "../core";
+import {
+  RegisterNameParameters,
+  getExpiration,
+  getOwnerAddress,
+  registerName,
+  getPrimaryName,
+  setPrimaryName,
+} from "../internal/ans";
+import { InputGenerateTransactionOptions, InputSingleSignerTransaction } from "../transactions/types";
+import { HexInput, MoveAddressType } from "../types";
 import { AptosConfig } from "./aptosConfig";
 
 /**
@@ -30,6 +38,35 @@ export class ANS {
    */
   async getOwnerAddress(args: { name: string }): Promise<MoveAddressType | undefined> {
     return getOwnerAddress({ aptosConfig: this.config, ...args });
+  }
+
+  /**
+   * Retrieve the primary name for an account address.
+   *
+   * @param args.address - A HexInput (address) of the account
+   *
+   * @returns null if no primary name is set
+   * and an object {domainName: string, subdomainName?: string} if a primary name is set
+   *
+   */
+  async getPrimaryName(args: { address: HexInput }): ReturnType<typeof getPrimaryName> {
+    return getPrimaryName({ aptosConfig: this.config, ...args });
+  }
+
+  /**
+   * Sets the primary name for the sender's account.
+   *
+   * @param args.sender - The sender account
+   * @param args.name - A string of the name: primary, primary.apt, secondary.primary, secondary.primary.apt, etc.
+   *
+   * @returns SingleSignerTransaction
+   */
+  async setPrimaryName(args: {
+    sender: Account;
+    name: string;
+    options?: InputGenerateTransactionOptions;
+  }): ReturnType<typeof setPrimaryName> {
+    return setPrimaryName({ aptosConfig: this.config, ...args });
   }
 
   /**
--- a/src/api/ans.ts
+++ b/src/api/ans.ts
@@ -46,8 +46,7 @@ export class ANS {
   /**
    * Retrieve the expiration time of a domain name or subdomain name.
    *
-   * @param args.domainName - A string of the domain name to retrieve
-   * @param args.subdomainName - A string of the subdomain name to retrieve
+   * @param args.name - A string of the name to retrieve
    *
    * @returns number as a unix timestamp in seconds.
    */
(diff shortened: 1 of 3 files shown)


src/api/ans.ts
```diff
@@ -33,17 +44,96 @@ export class ANS {
...
 
   /**
-   * Registers a new domain or subdomain name
+   * Retrieve the expiration time of a domain name or subdomain name.
+   *
+   * @param args.name - A string of the name to retrieve
+   *
+   * @returns number as a unix timestamp in seconds.
+   */
+  async getExpiration(args: { name: string }): ReturnType<typeof getExpiration> {
```
0xmaayan, line 53: curios what are the benefits for `ReturnType<typeof getExpiration>` - it is not a pattern we use in the sdk

src/api/ans.ts
```diff
@@ -33,17 +44,96 @@ export class ANS {
...
+  }
+
+  /**
+   * Retrieve the primary name for an account address.
+   *
+   * @param args.address - A HexInput (address) of the account
+   *
+   * @returns a string if the account has a primary name, undefined otherwise
+   */
+  async getPrimaryName(args: { address: HexInput }): ReturnType<typeof getPrimaryName> {
```
0xmaayan, line 91:
```suggestion
  async getPrimaryName(args: { address: AccountAddressInput }): ReturnType<typeof getPrimaryName> {
```

src/internal/ans.ts
```diff
@@ -171,5 +171,207 @@ export async function registerName(args: RegisterNameParameters): Promise<InputS
...
+    throw new Error("The subdomain expiration time cannot be greater than the domain expiration time");
+  }
+
+  const transaction = await generateTransaction({
+    aptosConfig,
+    sender: sender.accountAddress.toString(),
+    data: {
+      function: `${routerAddress}::router::register_subdomain`,
+      functionArguments: [
+        new MoveString(domainName),
```
0xmaayan, line 199: you should be able to use primitive values and not to instantiate with those classes

src/internal/ans.ts
```diff
@@ -171,5 +171,207 @@ export async function registerName(args: RegisterNameParameters): Promise<InputS
...
+  });
+
+  return res[0] as number;
+}
+
+export async function renewDomain(args: {
+  aptosConfig: AptosConfig;
+  sender: Account;
+  name: string;
+  years: 1;
```
0xmaayan, line 350: to we need it as argument is it is a fixed value?

kaw2k: We will eventually let users renew / register domains for longer than a year. We just need to figure out the pricing model. For now the contract is locked in at 1 year, but will eventually be relaxed. 

We can avoid a breaking change for the SDK if we start with a tighter value `1` and loosen it up to `number` later. I can make the `year` argument optional and default it, that way the user isn't exposed to this oddity, if that helps.

src/api/ans.ts
kaw2k: If your api function is just a light wrapper around the internal function, this makes it so you don't need to duplicate the types. These functions are simple enough were the inferance doesn't get us much, I can switch them over.

src/internal/ans.ts
kaw2k: I made it optional

src/api/ans.ts
```diff
@@ -33,17 +44,96 @@ export class ANS {
...
+   *
+   * @param args.name - A string of the name: primary, primary.apt, secondary.primary, secondary.primary.apt, etc.
+   * @param args.address - A AccountAddressInput of the address to set the domain or subdomain to
+   *
+   * @returns SingleSignerTransaction
+   */
+  async setTargetAddress(args: {
+    sender: Account;
+    name: string;
+    address: AccountAddressInput;
```
0xmaayan, line 79: do we also need `options?: InputGenerateTransactionOptions` here?

src/internal/ans.ts
```diff
@@ -147,11 +147,12 @@ export async function registerName(args: RegisterNameParameters): Promise<InputS
...
 
   if (expiration.policy === "domain") {
-    if (expiration.years !== 1) {
+    const years = expiration.years ?? 1;
+    if (years !== 1) {
       throw new Error("For now, names can only be registered for 1 year at a time");
     }
 
-    const registrationDuration = expiration.years * 31536000;
+    const registrationDuration = years * 31536000;
```
0xmaayan, line 155: lets add an explanation on what  `31536000` represents

src/internal/ans.ts
0xmaayan: have you looked at it ^ ?

0xmaayan (commented):
- Some places we use Move sub-classes as transaction arguments and some places we use values as is - let's use one way (think we should be bale to use values as is).
- Some api functions dont accept `options` for transaction, we should add those
- Add documentation around function/policies/etc

src/internal/ans.ts
```diff
@@ -119,17 +119,17 @@ export interface RegisterNameParameters {
   sender: Account;
   name: string;
   expiration:
-    | { policy: "domain"; years: 1 }
+    | { policy: "domain"; years?: 1 }
```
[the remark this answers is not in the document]
kaw2k: I [document it in the API file](https://github.com/aptos-labs/aptos-ts-sdk/blob/d1a146761dac0b4442664e128ebadf2f8a6beeaa/src/api/ans.ts#L112-L145), is that sufficient?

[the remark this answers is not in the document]
0xmaayan: oh yep that works

src/internal/ans.ts
```diff
@@ -171,5 +172,207 @@ export async function registerName(args: RegisterNameParameters): Promise<InputS
...
+}): Promise<MoveAddressType | undefined> {
+  const { aptosConfig, name } = args;
+  const routerAddress = getRouterAddress(aptosConfig);
+  const { domainName, subdomainName } = isValidANSName(name);
+
+  const res = await view({
+    aptosConfig,
+    payload: {
+      function: `${routerAddress}::router::get_target_addr`,
+      functionArguments: [domainName, Option(subdomainName)],
```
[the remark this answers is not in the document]
kaw2k: This one is a bit confusing. The `view` is returning an empty vector. I created [helper functions for this file](https://github.com/aptos-labs/aptos-ts-sdk/blob/d1a146761dac0b4442664e128ebadf2f8a6beeaa/src/internal/ans.ts#L84-L88) to do the unwrapping. Would it help if I changed the naming?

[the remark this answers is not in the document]
0xmaayan: ohh I see what you mean. @gregnazario have we introduced something to help with this? [helper functions for this file](https://github.com/aptos-labs/aptos-ts-sdk/blob/d1a146761dac0b4442664e128ebadf2f8a6beeaa/src/internal/ans.ts#L84-L88)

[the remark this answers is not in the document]
gregnazario: View functions... no.  I can add some helpers / a remote ABI, but keeping in mind that it will need an entirely different input path than the current API (which is a real pain tbh).  I'm trying to get a different BCS update in for the next release on view functions on the API side.

Let's go forward what's here, and I can do a remote ABI refactor for this.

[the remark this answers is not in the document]
0xmaayan: should we at least have the SDK holds these functions and the view function to use it?

tests/e2e/ans/ans.test.ts
```diff
@@ -52,6 +61,45 @@ describe("ANS", () => {
...
         }),
       );
+
+      changeExpirationDate = async (
+        tokenMode: 0 | 1,
+        expirationDate: number,
+        domainName: string,
+        subdomainName?: string,
+      ) => {
+        const domain = new MoveString(domainName);
```
0xmaayan, line 71: dont need those classes...?

tests/e2e/ans/ans.test.ts
```diff
@@ -170,5 +228,290 @@ describe("ANS", () => {
...
+  });
+
+  describe("setPrimaryName and getPrimaryName", () => {
+    let alice: Account;
+    let bob: Account;
+    let domainName: string;
+    let subdomainName: string;
+
+    beforeEach(async () => {
+      alice = Account.generate();
```
0xmaayan, line 384: why do we need to generate a new account before each test? can't we use the same account?

tests/e2e/ans/ans.test.ts
```diff
@@ -52,6 +61,45 @@ describe("ANS", () => {
           },
         }),
       );
+
+      changeExpirationDate = async (
+        tokenMode: 0 | 1,
```
0xmaayan, line 66: what's tokenMode?

src/internal/ans.ts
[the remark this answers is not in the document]
0xmaayan: for sure.

0xmaayan (approved):

tests/e2e/ans/ans.test.ts
kaw2k: Added comments to the code

tests/e2e/ans/ans.test.ts
kaw2k: To isolate the tests. Having shared accounts between tests introduces race conditions with primary names.

PR #172, opened 2023-11-06
Add and use `indexed_type` field

0xmaayan:
### Description

Add new `indexed_type` field for `events` query

https://aptos-org.slack.com/archives/C03MN5F7WUV/p1697815925868219

### Test Plan
`getAccountEventsByEventType` test is failing because localnet needs to be regenerate with the new field, @banool is working on it
https://aptos-org.slack.com/archives/C03MN5F7WUV/p1699286279415299?thread_ts=1697815925.868219&cid=C03MN5F7WUV

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 
 # Unreleased
 
+- `getAccountEventsByEventType` query uses new `indexed_type` indexed field to avoid rate limit
+
 ## 0.0.4 (2023-11-03)
 
 - [`Breaking`] Changed all instances of `AccountAddress.fromHexInput` to `AccountAddress.from` to accept AccountAddress as well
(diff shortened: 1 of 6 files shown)


0xjinn (approved):
Looking at the table from hasura, is this the index id generated by "event type"? In this case, the user has to know what `id` the event they are querying for?

0xmaayan:
> Looking at the table from hasura, is this the index id generated by "event type"? In this case, the user has to know what `id` the event they are querying for?

no, this is the indexed event type

gregnazario (approved):

gregnazario:
Looks like the local testnet doesn't have this field, it needs to be updated in a CLI release?

0xmaayan:
> ### Test Plan
> `getAccountEventsByEventType` test is failing because localnet needs to be regenerate with the new field, @banool is working on it https://aptos-org.slack.com/archives/C03MN5F7WUV/p1699286279415299?thread_ts=1697815925.868219&cid=C03MN5F7WUV

yes
```
Test Plan
getAccountEventsByEventType test is failing because localnet needs to be regenerate with the new field, @banool is working on it
https://aptos-org.slack.com/archives/C03MN5F7WUV/p1699286279415299?thread_ts=1697815925.868219&cid=C03MN5F7WUV
```

PR #177, opened 2023-11-08
Some fixes across the SDK

0xmaayan:
### Description
Some fixes to the SDK while working on the wallet adapter

1. Type `params` in remote abi fetch flow with `TypeTag[]`
2. Remove unneeded function overload
3. Rename TOKEN => AUTH_TOKEN 
4. Use AUTH_TOKEN only for faucet requests - need to find a nicer way here
5. Add API_KEY for future use

### DIFF
--- a/src/client/core.ts
+++ b/src/client/core.ts
@@ -32,8 +32,11 @@ export async function request<Req, Res>(options: ClientRequest<Req>, client: Cli
     "content-type": contentType ?? MimeType.JSON,
   };
 
-  if (overrides?.TOKEN) {
-    headers.Authorization = `Bearer ${overrides?.TOKEN}`;
+  // TODO - auth token is being used only for faucet, it breaks full node requests.
+  // Find a more sophisticated way than that but without the need to add the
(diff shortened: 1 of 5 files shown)


src/client/core.ts
```diff
@@ -32,8 +32,11 @@ export async function request<Req, Res>(options: ClientRequest<Req>, client: Cli
     "content-type": contentType ?? MimeType.JSON,
   };
 
-  if (overrides?.TOKEN) {
-    headers.Authorization = `Bearer ${overrides?.TOKEN}`;
+  // TODO - auth token is being used only for faucet, it breaks full node requests.
+  // Find a more sophisticated way than that but without the need to add the
+  // auth_token on every `aptos.fundAccount()` call
+  if (overrides?.AUTH_TOKEN && url.includes("faucet")) {
```
gregnazario, line 38: We'll need to make a more complex auth token strategy for that then.

It's possible an node provider uses a token like this for authorization of requests

gregnazario (approved):

PR #181, opened 2023-11-10
Refactor transaction interface - more functions less arguments

0xmaayan:
### Description
We've been getting feedback that too much arguments is.... well, too much.

This introduces a new transaction flow interface where each "step" is under a dedicated namespace so we wont kill the `aptos` namespace and developers wont get lost using it.

In addition, we have different functions for different operations instead of killing a function with bunch of arguments available. 

We keep a `withFeePayer` boolean arguments that can be passed whenever we build a new transaction, since fee payer can technically be any transaction, we dont really need to reference it as a new transaction type for the user and it should only be an option for any single/multiAgent transaction.

This makes the dev exp more clear as developers dont need to remember or find their way into what arguments to pass.

**Single Signer Flow**
No Fee Payer
```
const transaction = await aptos.build.transaction({
      sender: "0x123",
      data: {
        function: `0x1::transfer::transfer`,
        functionArguments: [1, "0x456],
     },
});

const senderAuthenticator = aptos.sign.transaction({signer:"0x123",transaction})
const response = await aptos.submit.transaction({transaction,senderAuthenticator})
```
With Fee Payer
```
const transaction = await aptos.build.transaction({
      sender: "0x123",
      data: {
        function: `0x1::transfer::transfer`,
        functionArguments: [1, "0x456],
     },
     withFeePayer:true
});

const senderAuthenticator = aptos.sign.transaction({signer:"0x123",transaction})
const feePayerAuthenticator = aptos.sign.transactionAsFeePayer({signer:"0x789",transaction})
const response = await aptos.submit.transaction({transaction,senderAuthenticator,feePayerAuthenticator})
```

**Multi Agent Flow**
No Fee Payer
```
const transaction = await aptos.build.multiAgentTransaction({
      sender: "0x123",
      secondarySignerAddresses:["0x789"],
      data: {
        function: `0x1::transfer::transfer`,
        functionArguments: [1, "0x456],
     },
});

const senderAuthenticator = aptos.sign.transaction({signer:"0x123",transaction})
const secondarySignerAuthenticator = aptos.sign.transaction({signer:"0x789",transaction})

const response = await aptos.submit.transaction({transaction,senderAuthenticator,additionalSignersAuthenticators: [secondarySignerAuthenticator]})
```
With Fee Payer
```
const transaction = await aptos.build.multiAgentTransaction({
      sender: "0x123",
      secondarySignerAddresses:["0x789"],
      data: {
        function: `0x1::transfer::transfer`,
        functionArguments: [1, "0x456],
     },
     withFeePayer:true
});

const senderAuthenticator = aptos.sign.transaction({signer:"0x123",transaction})
const secondarySignerAuthenticator = aptos.sign.transaction({signer:"0x789",transaction})
const feePayerAuthenticator = aptos.sign.transactionAsFeePayer({signer:"0x1011",transaction})
const response = await aptos.submit.transaction({transaction,senderAuthenticator,additionalSignersAuthenticators: [secondarySignerAuthenticator],feePayerAuthenticator})
```

Attached a screen video demonstrating the dev exp with this flow. (note this uses `generate` instead of `build` namespace and some functions from last commit but other than that it should pretty similar)

https://github.com/aptos-labs/aptos-ts-sdk/assets/29798064/0e02da66-5279-4bc2-ae31-6b341b86fd94


xbtmatt:
Wow, you did this fast, and it's incredible! So much nicer to read and look at, and a lot less overwhelming. This would also go *perfectly* with codegen for entry functions/even script functions, since they would both just go very nicely in the `data` section.

src/api/aptos.ts
```diff
@@ -46,6 +50,14 @@ export class Aptos {
...
   readonly transactionSubmission: TransactionSubmission;
 
+  readonly generate: Generate;
+
+  readonly sign: Sign;
+
+  readonly simulate: Simulate;
+
+  readonly submit: Submit;
+
```
xbtmatt, line 60: The best part about the namespacing here is that it intuitively encourages developers to understand the different parts of building a signed transaction without being overwhelming or convoluted.

Before we half tried to do it for them, half expected them to understand- so it was confusing/unintuitive. 

I think the main thing left is to see what the multi-agent flow is like and see if other people appreciate this flow better.

0xmaayan: multi-agent flow

https://github.com/aptos-labs/aptos-ts-sdk/assets/29798064/bfa294d2-c8f2-4e5c-8a7d-32850f44997e

src/api/transactionSubmission/submit.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  multiAgentTransaction(args: {
+    transaction: AnyRawTransaction;
+    senderAuthenticator: AccountAuthenticator;
+    additionalSignersAuthenticators: Array<AccountAuthenticator>;
+  }): Promise<PendingTransactionResponse> {
+    return submitTransaction({ aptosConfig: this.config, ...args });
+  }
+
+  multiAgentWithfeePayerTransaction(args: {
```
gregnazario, line 43: Looks like most here are missing `feePayer` -> `FeePayer`

src/api/transactionSubmission/generate.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  async multiAgentFeePayerTransaction(args: {
+    sender: AccountAddressInput;
+    data: InputGenerateTransactionPayloadData;
+    secondarySignerAddresses: AccountAddressInput[];
+    options?: InputGenerateTransactionOptions;
+  }) {
+    return generateTransaction({ aptosConfig: this.config, ...args, hasFeePayer: true });
+  }
```
gregnazario, line 51: Let's just combine the fee payers, and I think we're good to go

src/transactions/types.ts
```diff
@@ -261,38 +239,29 @@ export type InputSimulateTransactionOptions = {
...
- * Interface that holds the user data input when generating a fee payer transaction
- */
-export interface InputGenerateFeePayerRawTransactionData {
-  sender: AccountAddressInput;
-  hasFeePayer: true;
-  secondarySignerAddresses?: AccountAddressInput[];
   options?: InputGenerateTransactionOptions;
-  data: InputGenerateTransactionPayloadData;
+  withFeePayer?: boolean;
+  secondarySignerAddresses?: undefined;
```
0xmaayan, line 245: typescript doesnt know how to solve the return type without having the same type props in all function overload. 

@gregnazario not sure how it worked for you before. I think it didnt really infer the return type because we also use the general type in the function overload https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/api/transactionSubmission.ts#L47

gregnazario: Oh, I see without the overload you'd have to do splitting of the output type.

Does it matter though?

src/api/ans.ts
```diff
@@ -141,9 +141,9 @@ export class ANS {
    * @param args.toAddress optional - The address to send the domain name to. If not provided,
    * the transaction will be sent to the router.
    *
-   * @returns InputSingleSignerTransaction
+   * @returns SingleSignerTransaction
    */
-  async registerName(args: Omit<RegisterNameParameters, "aptosConfig">): Promise<InputSingleSignerTransaction> {
+  async registerName(args: Omit<RegisterNameParameters, "aptosConfig">): Promise<SingleSignerTransaction> {
```
gregnazario, line 146: Yay! remove `Input` lol

src/api/transactionSubmission/submit.ts
```diff
@@ -0,0 +1,45 @@
...
+    additionalSignersAuthenticators: Array<AccountAuthenticator>;
+    feePayerAuthenticator?: AccountAuthenticator;
+  }): Promise<PendingTransactionResponse> {
+    if (args.transaction instanceof FeePayerRawTransaction) {
+      if (!args.feePayerAuthenticator) {
+        throw new Error("You are submitting a Fee Payer transaction but missing the feePayerAuthenticator");
+      }
+    }
+    return submitTransaction({ aptosConfig: this.config, ...args });
+  }
```
gregnazario, line 44: If it's a fee payer, we should also ensure that the FeePayerAddress is not 0x0 (this will always fail).  It should be the address of the actual fee payer on submission.

src/api/transactionSubmission/simulate.ts
```diff
@@ -0,0 +1,48 @@
...
+  transaction(args: {
+    signerPublicKey: PublicKey;
+    transaction: AnyRawTransaction;
+    feePayerPublicKey?: PublicKey;
+    options?: InputSimulateTransactionOptions;
+  }): Promise<Array<UserTransactionResponse>> {
+    if (args.transaction instanceof FeePayerRawTransaction) {
+      if (!args.feePayerPublicKey) {
+        throw new Error("You are simulating a Fee Payer transaction but missing the feePayerPublicKey");
+      }
```
gregnazario, line 29: Similarly unsure if you need the address, but I'm not worried about this one.

src/transactions/types.ts
0xmaayan: it matters since we want to return to the user the correct type so they wont need to
```
if (transaction typeof SingleSignerTransaction){....}
```

src/api/transactionSubmission/sign.ts
```diff
@@ -0,0 +1,44 @@
...
+    });
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  transactionAsFeePayer(args: { signer: Account; transaction: AnyRawTransaction }): AccountAuthenticator {
+    const { signer, transaction } = args;
+
+    // if "feePayerAddress" is not in "transaction" it means
+    // this is not a fee payer transaction
+    if (!("feePayerAddress" in transaction)) {
```
gregnazario, line 32: Someday we'll have to create typeguard functions all around to make these nice looking.

gregnazario (approved):
So these are now mostly just a 1:1 mapping of the old to new.

Could you make a quick summary of what changed / how to migrate in the CHANGELOG?

src/api/transactionSubmission/submit.ts
0xmaayan: we expect to have a `feePayerAuthenticator` that is the authenticator of the account who signed with `sign.transactionAsFeePayer` where we assign the signer address to the `transaction.feePayerAddress`

src/api/transactionSubmission/simulate.ts
0xmaayan: as for the wallet, I believe @hardsetting mentioned they only need the fee payer account public key

src/api/transactionSubmission/sign.ts
0xmaayan: ohh we can actually have a better way to handle it now - let me change this yack thing

PR #182, opened 2023-11-10
Respect max gas amount value

0xmaayan:
### Description
Although we give the option to pass in `maxGasAmount` value when generating transaction, we dont respect it.

### DIFF
--- a/src/transactions/transactionBuilder/transactionBuilder.ts
+++ b/src/transactions/transactionBuilder/transactionBuilder.ts
@@ -232,7 +232,7 @@ export async function generateRawTransaction(args: {
   );
 
   const { maxGasAmount, gasUnitPrice, expireTimestamp } = {
-    maxGasAmount: BigInt(DEFAULT_MAX_GAS_AMOUNT),
+    maxGasAmount: options?.maxGasAmount ? BigInt(options.maxGasAmount) : BigInt(DEFAULT_MAX_GAS_AMOUNT),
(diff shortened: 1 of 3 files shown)


xbtmatt (approved):

src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -234,7 +234,7 @@ export async function generateRawTransaction(args: {
   );
 
   const { maxGasAmount, gasUnitPrice, expireTimestamp } = {
-    maxGasAmount: BigInt(DEFAULT_MAX_GAS_AMOUNT),
+    maxGasAmount: options?.maxGasAmount ? BigInt(options.maxGasAmount) : BigInt(DEFAULT_MAX_GAS_AMOUNT),
     gasUnitPrice: BigInt(gasEstimate),
```
gregnazario, line 238: Do we also care about amounts from gas unit price?

gregnazario (approved):

src/transactions/transactionBuilder/transactionBuilder.ts
0xmaayan: we do https://github.com/aptos-labs/aptos-ts-sdk/pull/182/files#diff-97f79e1249bb5dfd1fc52b16da86fbcd1d517e8897c7cc0d70c79e73c61ecf3cR230

PR #183, opened 2023-11-11
[Rename] Changing ViewRequestData => InputViewRequestData

xbtmatt:
### Description
Renaming `ViewRequestData` to `InputViewRequestData` to adhere to our naming conventions

I mostly did this because I came across `ViewRequest` which seems to be the JSON payload response type from the node. Am I correct in my understanding @0xmaayan ?

### Test Plan
`pnpm jest`

Updated `CHANGELOG.md` as well to note that it's a breaking change


gregnazario (approved):
This is direct from the node and only supports string and number json inputs

PR #184, opened 2023-11-11
[Error message] Adding a clearer error message for when the typeTagParser encounters a generic but generics aren't allowed

xbtmatt:
### Description
While using the `typeTagParser(...)` function, I was getting this error:

```shell
TypeTagParserError: Failed to parse typeTag 'T0', unknown type.
```

I added a clearer error message if the string matches a generic type but generics are disallowed:

```shell
TypeTagParserError: Failed to parse typeTag 'T0', unexpected generic type.
```

Updated the tests

```shell
pnpm jest typeTagParser.test.ts
```

### DIFF
--- a/src/transactions/typeTag/parser.ts
+++ b/src/transactions/typeTag/parser.ts
@@ -54,7 +54,7 @@ type TypeTagState = {
 
 export enum TypeTagParserErrorType {
   InvalidTypeTag = "unknown type",
-  InvalidButPossiblyGeneric = "unknown type. Did you mean to enable generics?",
+  InvalidButPossiblyGeneric = "unknown type. Did you mean to allow generics?",
   UnexpectedTypeArgumentClose = "unexpected '>'",
   UnexpectedWhitespaceCharacter = "unexpected whitespace character",
   UnexpectedComma = "unexpected ','",
--- a/src/transactions/typeTag/parser.ts
+++ b/src/transactions/typeTag/parser.ts
@@ -267,15 +267,15 @@ function parseTypeTagInner(str: string, types: Array<TypeTag>, allowGenerics: bo
       }
       return new TypeTagVector(types[0]);
     default:
-      if (allowGenerics && isGeneric(str)) {
-        return new TypeTagGeneric(Number(str.split("T")[1]));
+      if (isGeneric(str)) {
+        if (allowGenerics) {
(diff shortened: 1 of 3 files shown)


src/transactions/typeTag/parser.ts
```diff
@@ -262,12 +267,15 @@ function parseTypeTagInner(str: string, types: Array<TypeTag>, allowGenerics: bo
       }
       return new TypeTagVector(types[0]);
     default:
-      if (allowGenerics && str.match(/^T[0-9]+$/)) {
+      if (allowGenerics && isGeneric(str)) {
         return new TypeTagGeneric(Number(str.split("T")[1]));
       }
```
gregnazario, line 272: I would change this logic to:
```
if (isGeneric(str) {
  if (allowGenerics){
   return...
   } else {
    throw
  }
}
```

src/transactions/typeTag/parser.ts
```diff
@@ -50,6 +54,7 @@ type TypeTagState = {
 
 export enum TypeTagParserErrorType {
   InvalidTypeTag = "unknown type",
+  InvalidButPossiblyGeneric = "unknown type. Did you mean to allow generics?",
```
gregnazario, line 57: Full types must start with 0x..., so we can just say it's an unexpected generic.

```suggestion
  InvalidButPossiblyGeneric = "unexpected generic.  Did you mean to allow generics?",
```

gregnazario (approved):

src/transactions/typeTag/parser.ts
```diff
@@ -50,6 +54,7 @@ type TypeTagState = {
 
 export enum TypeTagParserErrorType {
   InvalidTypeTag = "unknown type",
+  UnexpectedGenericType = "unexpected generic type",
```
0xmaayan, line 57: do we need to document the `allowGenerics` option better?

0xmaayan (approved):

src/transactions/typeTag/parser.ts
xbtmatt: TBH I think it's fine with this, although maybe in the function doc comment we can note the option clearer.

The main thing for me that was confusing was the error didn't let me know it was a generic, so I didn't realize it had an understanding of generics. Once I realized that I just looked at the options args and it was clear how to enable them

PR #187, opened 2023-11-12
[MoveValue] Cleanup types to match the proper types

gregnazario:
### Description
Struct types were not able to be used as MoveValue which breaks some use cases.  Additionally, the MoveStructType has been renamed to MoveStructId as there was some strange merging of the two.

### Test Plan
CI, but need to follow up with a full suite of tests on view function outputs.

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -252,7 +252,7 @@ export type GasEstimation = {
 };
 
 export type MoveResource = {
-  type: MoveStructType;
+  type: MoveStructId;
   data: {};
 };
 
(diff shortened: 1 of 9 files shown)


src/types/index.ts
```diff
@@ -686,7 +686,13 @@ export type MoveOptionType = MoveType | null | undefined;
 /**
  * This is the format for a fully qualified struct, resource, or entry function in Move.
  */
-export type MoveStructType = `${string}::${string}::${string}`;
+export type MoveStructId = `${string}::${string}::${string}`;
+// These are the same, unfortunately, it reads really strangely to take a StructId for a Function and there wasn't a
+// good middle ground name.
+export type MoveFunctionId = MoveStructId;
```
xbtmatt, line 692: 💯

xbtmatt (approved):

PR #188, opened 2023-11-13
[view] Add generic for typing view function outputs

gregnazario:
### Description
Similar to getResource, but typed accordingly with MoveValue[] as the default. Only types that extend that can be used.

### Test Plan
This adds a few view function input and output type tests.  Full tests will need a module properly published for it to work.

### DIFF
--- a/tests/e2e/api/general.test.ts
+++ b/tests/e2e/api/general.test.ts
@@ -34,19 +34,6 @@ describe("general api", () => {
     expect(blockData.block_height).toBe(blockVersion.toString());
   });
 
-  test("it fetches view function data", async () => {
-    const config = new AptosConfig({ network: Network.LOCAL });
-    const aptos = new Aptos(config);
-
-    const payload: InputViewRequestData = {
-      function: "0x1::chain_id::get",
-    };
-
-    const chainId = await aptos.view({ payload });
-
-    expect(chainId[0]).toBe(4);
-  });
-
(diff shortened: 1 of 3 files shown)


tests/e2e/api/general.test.ts
```diff
@@ -108,4 +95,120 @@ describe("general api", () => {
...
+      const isRegistered = (await aptos.view<[boolean]>({ payload: payload2 }))[0];
+      expect(isRegistered).toEqual(false);
+
+      const payload3: InputViewRequestData = {
+        function: "0x1::coin::supply",
+        typeArguments: ["0x1::aptos_coin::AptosCoin"],
+        functionArguments: [],
+      };
+
+      const supply = (await aptos.view<[{ vec: [string] }]>({ payload: payload3 }))[0].vec[0];
```
xbtmatt, line 198: so painful :P

I also just realized this morning while writing tests for view function outputs is that the returned results are serialized in some cases. It seems like the inputs *and* the outputs are a frankenstein of a JSON mapping and BCS-serialized hex strings...

xbtmatt (approved):
This is great though, makes things easier for sure

tests/e2e/api/general.test.ts
gregnazario: I suspect with improvements on the actual API, a lot of this can be improved upon / cleaned up.

PR #193, opened 2023-11-15
Use primitive values in entry function transactions in tests and exam…

0xmaayan:
…ples


examples/typescript/simple_transfer.ts
```diff
@@ -88,8 +79,8 @@ const example = async () => {
     sender: alice.accountAddress,
     data: {
       function: "0x1::coin::transfer",
-      typeArguments: [parseTypeTag(APTOS_COIN)],
-      functionArguments: [AccountAddress.from(bob.accountAddress), new U64(TRANSFER_AMOUNT)],
+      typeArguments: [APTOS_COIN],
+      functionArguments: [bob.accountAddress, TRANSFER_AMOUNT],
     },
```
gregnazario, line 84: Yay simple mode!

Two examples I want to add are: How to just use a script directly, and also how to use prebuilt ABI so it doesn't have as much overhead.

gregnazario (approved):
We should probably make 1 example of how to mix and match or use encoded ones.  But, simple should be simple :)

PR #197, opened 2023-11-17
Unifies ANS dates

kaw2k:
### Description
ANS had 3 different implementations for times. This converts all uses of time to be milliseconds since the epoch. 

### Test Plan
Updated and passing test cases

### Related Links
https://aptos-org.slack.com/archives/C05NLAKJM9U/p1700245777573069

### DIFF
--- a/src/internal/ans.ts
+++ b/src/internal/ans.ts
@@ -123,7 +123,7 @@ export interface RegisterNameParameters {
   expiration:
     | { policy: "domain"; years?: 1 }
     | { policy: "subdomain:follow-domain" }
-    | { policy: "subdomain:independent"; expirationDate: Date };
+    | { policy: "subdomain:independent"; expirationDate: number };
   transferable?: boolean;
   toAddress?: AccountAddressInput;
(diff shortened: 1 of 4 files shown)


src/internal/ans.ts
```diff
@@ -465,7 +470,7 @@ export async function getAccountSubdomains(
     },
   });
 
-  return data.current_aptos_names;
+  return data.current_aptos_names.map(sanitizeANSName);
```
0xmaayan, line 473: do we need to validate `current_aptos_names` is not empty?

0xmaayan (approved):

src/internal/ans.ts
kaw2k: Nope! The indexer always gives back an array, empty or not. `map` here will be a no-op if we get an empty array back. 

image

gregnazario:
@kaw2k please update the CHANGELOG, and we can merge it

kaw2k:
Ok, updated the chnagelog!

PR #204, opened 2023-11-21
Some clean ups

0xmaayan:
### Description
1. remove `toString()` method for `accountAddress` when building a transaction
2. add `.env` file to .gitignore
3. Use simple values when building a transaction
4. Missing doc comments
5. Remove `throw Error` from queries when no actual error


tests/e2e/ans/ans.test.ts
```diff
@@ -449,13 +449,13 @@ describe("ANS", () => {
...
       alice = Account.generate();
       await aptos.fundAccount({
-        accountAddress: alice.accountAddress.toString(),
+        accountAddress: alice.accountAddress,
         amount: 500_000_000,
       });
 
       bob = Account.generate();
       await aptos.fundAccount({
-        accountAddress: bob.accountAddress.toString(),
```
gregnazario, line 458: It's so much nicer without these

tests/e2e/api/staking.test.ts
```diff
@@ -20,11 +20,12 @@ describe("staking api", () => {
     expect(numDelegators).toEqual(numDelegatorsData[0].num_active_delegator);
   });
 
-  test("it throws if the poolAddress does not exist", async () => {
+  test.only("it throws if the poolAddress does not exist", async () => {
```
gregnazario, line 23: what does `only` do?

gregnazario (approved):

tests/e2e/ans/ans.test.ts
0xmaayan: oh YES!

tests/e2e/api/staking.test.ts
0xmaayan: ohh it should be here. `only` makes it that only this test will run when you run this test suite

examples/typescript/facoin/facoin.json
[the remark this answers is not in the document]
0xmaayan: added, but probably need to delete it, commit and then add it again. we can do it in a new PR - or better makes sense in this example

PR #207, opened 2023-11-27
[refactor] Flatten options to be consistent across all, remove casts, fix imports and exports

gregnazario:
### Description
This flattens all input arguments to be a single level, which should simplify the workflow without having to decide which one is nested and which one is not.

It also removes all unnecessary casts from tests and everywhere else, to ensure that there aren't any time bomb bugs due to casts.

Cleans up imports, found some more things to export

### Test Plan
All the tests pass locally!

### DIFF
--- a/src/internal/ans.ts
+++ b/src/internal/ans.ts
@@ -84,8 +84,8 @@ function getRouterAddress(aptosConfig: AptosConfig): string {
   return address;
 }
 
-const Some = <T>(value: T): MoveValue => ({ vec: [value] }) as any;
-const None = (): MoveValue => ({ vec: [] }) as any;
+const Some = <T>(value: T): MoveValue => ({ vec: [value] });
+const None = (): MoveValue => ({ vec: [] });
 // != here is intentional, we want to check for null and undefined
 // eslint-disable-next-line eqeqeq
 const Option = <T>(value: T | undefined | null): MoveValue => (value != undefined ? Some(value) : None());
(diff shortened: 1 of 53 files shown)


src/internal/ans.ts
```diff
@@ -514,15 +514,14 @@ async function getANSExpirationDate(args: { aptosConfig: AptosConfig }): Promise
   const { aptosConfig } = args;
   const routerAddress = getRouterAddress(aptosConfig);
 
-  const res = await view({
+  const [gracePeriodInSeconds] = await view<[number]>({
```
0xmaayan, line 517: for consistency
```suggestion
  const [gracePeriodInSeconds] = await view<Array<number>>({
```

tests/unit/account.test.ts
```diff
@@ -53,10 +53,14 @@ describe("Account", () => {
       const accountAddress = AccountAddress.from(address);
       const newAccount = Account.fromPrivateKeyAndAddress({ privateKey, address: accountAddress, legacy: true });
       expect(newAccount).toBeInstanceOf(Account);
-      expect(newAccount.publicKey).toBeInstanceOf(Ed25519PublicKey);
-      expect(newAccount.privateKey).toBeInstanceOf(Ed25519PrivateKey);
-      expect((newAccount.privateKey as Ed25519PrivateKey).toString()).toEqual(privateKey.toString());
-      expect((newAccount.publicKey as Ed25519PublicKey).toString()).toEqual(new Ed25519PublicKey(publicKey).toString());
+      if (!Ed25519PublicKey.isPublicKey(newAccount.publicKey)) {
```
0xmaayan, line 56: it feels a different testing format - where we throw instead of expect, not sure why we would prefer to throw?

0xmaayan (approved):
looks good! left a comment on throw vs expect in tests

also, any CHANGELOG updates?

PR #210, opened 2023-11-28
fix zero account sequence number

maykonmichel:
### Description

This pull request addresses an issue in the `generateRawTransaction` function where the `getInfo` function throws an error when the account is not on the chain, and manually setting the account sequence number to 0 was ineffective due to the falsy nature of this value. The proposed solution involves a minor adjustment to ensure more robust handling of the account sequence number. This modification replaces the falsy check with a comparison to `undefined`, providing a more accurate and reliable method for handling the account sequence number.

### Test Plan

A new test case has been added to verify the behavior when generating a raw transaction with an account not on the chain and the account sequence number set to 0. Additionally, an error scenario has been tested where an attempt is made to generate a raw transaction without specifying the account sequence number, expecting an error to be thrown.

### DIFF
--- a/src/transactions/transactionBuilder/transactionBuilder.ts
+++ b/src/transactions/transactionBuilder/transactionBuilder.ts
@@ -209,9 +209,10 @@ export async function generateRawTransaction(args: {
 }): Promise<RawTransaction> {
   const { aptosConfig, sender, payload, options } = args;
 
-  const getSequenceNumber = options?.accountSequenceNumber
-    ? Promise.resolve({ sequence_number: options.accountSequenceNumber })
-    : getInfo({ aptosConfig, accountAddress: sender });
+  const getSequenceNumber =
+    options?.accountSequenceNumber !== undefined
+      ? Promise.resolve({ sequence_number: options.accountSequenceNumber })
+      : getInfo({ aptosConfig, accountAddress: sender });
 
   const getChainId = NetworkToChainId[aptosConfig.network]
     ? Promise.resolve({ chain_id: NetworkToChainId[aptosConfig.network] })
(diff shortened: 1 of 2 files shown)


src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -209,9 +209,10 @@ export async function generateRawTransaction(args: {
 }): Promise<RawTransaction> {
   const { aptosConfig, sender, payload, options } = args;
 
-  const getSequenceNumber = options?.accountSequenceNumber
-    ? Promise.resolve({ sequence_number: options.accountSequenceNumber })
-    : getInfo({ aptosConfig, accountAddress: sender });
+  const getSequenceNumber =
+    options?.accountSequenceNumber !== undefined
+      ? Promise.resolve({ sequence_number: options.accountSequenceNumber })
+      : getInfo({ aptosConfig, accountAddress: sender });
```
gregnazario, line 215: We will probably want to verify anywhere else we use ternary operators and other places where we're trying to take advantage of falsiness then

gregnazario (commented):
Interesting... I mean the transaction will fail to submit if the account doesn't exist on chain right?

gregnazario (approved):
Oh, I see, if the sequence number is 0 is not a good check.

src/transactions/transactionBuilder/transactionBuilder.ts
maykonmichel: Certainly, but the appropriateness of using ternary operators and leveraging falsiness depends on the specific use case.

gregnazario: I'd prefer to avoid using falsiness / truthiness as much as possible to prevent mistakes like this

maykonmichel:
> Interesting... I mean the transaction will fail to submit if the account doesn't exist on chain right?

[AIP-52](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-52.md) allows for the submission of sponsored transactions by an account that is not on chain, thereby the account's sequence number is 0.

PR #214, opened 2023-11-29
[accountAddress] Make strict mode optional, and relaxed default

gregnazario:
### Description
More often than not the SDK receives addresses that are shortened.  This is very possible to occur, and it should be allowed as an input as default.  Strictness should only be applied if wanted, and until there is a different address standard, it should accept any output from the API.  Otherwise, strings won't be easily converted.

We can disable this once we add full type bindings to do on the fly struct conversion rather than use JSON intermediaries.

### Test Plan
I simply flipped relaxed to normal, and added a strict.  The CI should run enough tests to cover it.


0xmaayan (approved):
lgtm. just need to fix tests

banool (approved):

PR #219, opened 2023-12-01
[remoteAbi] Add argument parsing unit tests

gregnazario:
### Description
This provides nearly full coverage of all the parsing logic, as well as fixed a couple bugs in the parsing logic.

### Test Plan
This tests all inputs and outputs, along with mixed inputs, and ensures that they pass correctly.

I purposely did not spend time to check all the error messages explicitly to save time, but coverage shows they are covered.

### DIFF
--- a/src/transactions/transactionBuilder/remoteAbi.ts
+++ b/src/transactions/transactionBuilder/remoteAbi.ts
@@ -149,7 +149,7 @@ function parseArg(
     if (isString(arg)) {
       return AccountAddress.fromString(arg);
     }
-    throwTypeMismatch("string", position);
+    throwTypeMismatch("string | AccountAddress", position);
   }
   if (param.isU8()) {
     if (isNumber(arg)) {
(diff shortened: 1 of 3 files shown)


src/transactions/transactionBuilder/remoteAbi.ts
```diff
@@ -335,7 +340,10 @@ function checkType(param: TypeTag, arg: EntryFunctionArgumentTypes, position: nu
     }
     if (param.isOption()) {
       if (arg instanceof MoveOption) {
-        // TODO: more introspection for the type
+        // If there's a value, we can check the inner type (otherwise it doesn't really matter)
+        if (arg.value !== undefined) {
```
0xmaayan, line 344: need to also check for `null`?

tests/unit/remoteAbi.test.ts
```diff
@@ -0,0 +1,234 @@
...
+  TypeTagVector,
+  U128,
+  U16,
+  U256,
+  U32,
+  U64,
+  U8,
+} from "../../src";
+
+const MAX_U8 = 255;
```
0xmaayan, line 20: sdk has these consts https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/bcs/consts.ts

0xmaayan (approved):
noice!

PR #220, opened 2023-12-01
[string] Add toString for most classes

gregnazario:
### Description
Adds a toString implementation for many classes

### Test Plan
None

### DIFF
--- a/src/bcs/serializable/moveStructs.ts
+++ b/src/bcs/serializable/moveStructs.ts
@@ -220,6 +220,10 @@ export class MoveVector<T extends Serializable & EntryFunctionArgument>
     }
     return new MoveVector(values);
   }
+
+  toString(): string {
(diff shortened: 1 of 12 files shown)


src/bcs/serializable/moveStructs.ts
```diff
@@ -279,8 +287,10 @@ export class MoveOption<T extends Serializable & EntryFunctionArgument>
   /**
    * Retrieves the inner value of the MoveOption.
    *
-   * This method is inspired by Rust's `Option<T>.unwrap()`.
-   * In Rust, attempting to unwrap a `None` value results in a panic.
+   * This method is inspired by Rust's `;
+   Option<T>.unwrap()`.
+   * In Rust, attempting to unwrap a `;
+   None` value results in a panic.
```
0xmaayan, line 293: format here is a bit weird

0xmaayan (approved):
very cool. could we add tests?

PR #221, opened 2023-12-01
[refactor] Cleanup minor nits

gregnazario:
### Description
This cleans up grammar, spellchecks, and style that was picked up automatically by the IDE.  Additionally, fix the docs page, and remove duplicate constants.

### Test Plan
`pnpm test`  This actually pointed out that some test were improperly awaiting on something that wasn't supposed to be async (but it was).

Additionally, found some other issues in tests that caused failures due to not waiting on the indexer to catch up.


gregnazario:
Closing will, come back with another pass

PR #225, opened 2023-12-06
Rename custom metric header

0xmaayan:
### Description
Rename custom metric header to "aptos-new-ts-sdk" to distinguish between the legacy and the new sdk

### DIFF
--- a/src/client/core.ts
+++ b/src/client/core.ts
@@ -28,7 +28,7 @@ export async function request<Req, Res>(options: ClientRequest<Req>, client: Cli
   const { url, method, body, contentType, params, overrides } = options;
   const headers: Record<string, string | AnyNumber | boolean | undefined> = {
     ...overrides?.HEADERS,
-    "x-aptos-client": `aptos-ts-sdk/${VERSION}`,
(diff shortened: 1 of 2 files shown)


src/client/core.ts
```diff
@@ -28,7 +28,7 @@ export async function request<Req, Res>(options: ClientRequest<Req>, client: Cli
   const { url, method, body, contentType, params, overrides } = options;
   const headers: Record<string, string | AnyNumber | boolean | undefined> = {
     ...overrides?.HEADERS,
-    "x-aptos-client": `aptos-ts-sdk/${VERSION}`,
+    "x-aptos-client": `aptos-new-ts-sdk/${VERSION}`,
```
gregnazario, line 31: How about putting `@aptos-labs/ts-sdk`?  That'll keep it named the same as the npm package, and forgo needing the `new` which will someday be old.

gregnazario (approved):

src/client/core.ts
gregnazario: Oh, yeah that's a good point... maybe something simpler that differentiates, but won't become too old:
`aptos-typescript-sdk`

PR #226, opened 2023-12-07
Change transaction submission flow namespace

0xmaayan:
### Description
We got some user feedbacks that the current transaction builder flow is a bit confusing
```
aptos.build.transaction({})
aptos.sign.transaction({})
aptos.submit.transaction({})
```

This PR changes all the builder options to be under a one and meaningful `transaction` namespace
```
aptos.transaction.build.simple();
aptos.transaction.simulate.simple();
aptos.transaction.submit.simple();

aptos.transaction.sign()
aptos.transaction.signAsFeePayer()

aptos.transaction.build.multiAgent();
aptos.transaction.simulate.multiAgent();
aptos.transaction.submit.multiAgent();
```

### Test Plan
unit tests are passing
example tests are passing

### Related Links
https://aptos-org.slack.com/archives/C05NLAKJM9U/p1701765572858389


README.md
```diff
@@ -120,8 +120,8 @@ const transaction = await aptos.build.transaction({
 });
 
 // using sign and submit separately
-const senderAuthenticator = aptos.sign.transaction({ signer: alice, transaction });
-const committedTransaction = await aptos.submit.transaction({ transaction, senderAuthenticator });
+const senderAuthenticator = aptos.transaction.sign({ signer: alice, transaction });
+const committedTransaction = await aptos.transaction.submit.simple({ transaction, senderAuthenticator });
```
gregnazario, line 124: simple, much better than natural.  Though, I now see why had had it the other way.  `simple` is for a single signer?

src/api/transactionSubmission/build.ts
```diff
@@ -21,7 +21,17 @@ export class Build {
...
+   * Build a simple transaction
+   *
+   * @param args.sender The sender account address
+   * @param args.data The transaction data
+   * @param args.options optional. Optional transaction configurations
+   * @param args.withFeePayer optional. Whether there is a fee payer for the transaction
+   *
+   * @returns SingleSignerTransaction
+   */
+  async simple(args: {
```
gregnazario, line 34: Oh, I see... simple is fee payer also, which gets confusing if it's singleSigner, though it does return a SingleSignerTransaction.

gregnazario (approved):
LGTM, I'm trying to brainstorm a better name than `simple`, but I can't seem to think of one.

`single_sender` seems confusing, there's only ever one sender

`single_signer` can be confusing, because it could be a multisig wallet

`single_agent` this is also confusing because it isn't used anywhere as a term

`single` ... Is this a single transaction 👎🏻 

`simple` makes me think multiAgent must be complex

`basic` :/

I think.... `simple` is probably fine to go with.  At least it's not the old `EntryFunction.natural`

src/api/transactionSubmission/build.ts
0xmaayan: yea... i'll probably rename that also in a second PR to try and not add 10 more files to this PR

README.md
0xmaayan: simple is for... a simple transaction. Since everything can be fee payer there is no real only single signer

PR #227, opened 2023-12-07
[config] Ensure custom indexer URLs errors work properly / document how to use custom URLs

gregnazario:
### Description
There's two things here:
1. We were parsing Indexer URLs to tell how to handle errors, though a user could have a custom URL.  Since, this custom URL basically has to work the same as the Aptos labs indexer, I removed the "check it it's an indexer by URL", and moved it to use the API type instead.
2. Documentation around how to provide a config directly was straightforward, but needed to be more explicit around `/v1` paths, since the old SDK handled it slightly differently.

### Test Plan
`pnpm test` still passes

### DIFF
--- a/src/api/aptosConfig.ts
+++ b/src/api/aptosConfig.ts
@@ -8,13 +8,29 @@ import { AptosApiType, DEFAULT_NETWORK } from "../utils/const";
 
 /**
  * This class holds the config information for the SDK client instance.
+ *
+ * @example To use Aptos Labs endpoints, simply set the `network` property to the desired network.
+ * ```typescript
+ *   const config = new AptosConfig({ network: Network.DEVNET });
+ * ```
+ *
+ * @example To use custom endpoints, set the `network` property to the desired network, and the appropriate URLs. The URL
+ * must contain the full path and will not have a `/v1` added.
+ * ```typescript
+ *   const config = new AptosConfig({
+ *     network: Network.TESTNET,
+ *     fullnode: "https://my-testnet-fullnode.com/v1",
+ *   });
+ * ```
  */
 export class AptosConfig {
-  /** The Network that this SDK is associated with. Defaults to DEVNET */
+  /**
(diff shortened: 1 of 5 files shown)


src/client/core.ts
```diff
@@ -83,7 +86,7 @@ export async function aptosRequest<Req extends {}, Res extends {}>(
 
   // to support both fullnode and indexer responses,
   // check if it is an indexer query, and adjust response.data
-  if (aptosConfig.isIndexerRequest(url)) {
+  if (type === AptosApiType.INDEXER) {
```
0xmaayan, line 89: hmm not sure we can assume that a custom url/service acts the same as our own service so it makes sense to check if it is an aptos indexer response and then do the conversation - we might need to do a different logic for custom services but not sure how custom indexer service is common

gregnazario: I think at this point we assume it works the same, given that we assume all the tables that exist and all.

Any custom indexing at this point basically would need its own client, separate of the SDK

src/api/aptosConfig.ts
```diff
@@ -8,13 +8,29 @@ import { AptosApiType, DEFAULT_NETWORK } from "../utils/const";
...
+ * ```typescript
+ *   const config = new AptosConfig({ network: Network.DEVNET });
+ * ```
+ *
+ * @example To use custom endpoints, set the `network` property to the desired network, and the appropriate URLs. The URL
+ * must contain the full path and will not have a `/v1` added.
+ * ```typescript
+ *   const config = new AptosConfig({
+ *     network: Network.TESTNET,
+ *     fullnode: "https://my-testnet-fullnode.com/v1",
```
0xmaayan, line 22: this needs to be better documented. We also use `Network.CUSTOM` where the sdk expects to fullnode,faucet and indexer custom endpoints.
https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/api/aptosConfig.ts#L57

PR #234, opened 2023-12-12
[indexer/example] Fix waitOnIndexer to wait on appropriate tables, and fix examples to not sleep

gregnazario:
### Description
The examples kept failing locally for me, because the indexer wasn't being waited on and instead sleeping.  Then, when I switched it to wait on the indexer it didn't work because the processor it waited on was wrong.

This fixes the SDK so it waits on the appropriate processor, and it removes the double (for when looking up APT), and there was one indexer wait, even though it was using a table API lookup.

### Test Plan
All examples that were repeatedly failing (because I had too much stuff on my machine) passed.

`pnpm test` also passed


0xjinn (commented):
Nice found!

btw, looks like multi-agent_transfer example is failing on CI?

gregnazario:
> Nice found!
> 
> btw, looks like multi-agent_transfer example is failing on CI?

It didn't before but, now it is.

But, of course if I put something to observe it, then it doesn't fail.

gregnazario:
> Nice found!
> 
> btw, looks like multi-agent_transfer example is failing on CI?

Fixed.  Was waiting on the wrong transaction, so it was always too fast

0xmaayan (approved):

PR #236, opened 2023-12-12
[remoteAbi] Allow MoveString for vector<u8>

gregnazario:
### Description
This is a feature that kind of worked in the last SDK, but it's debatable if it's desirable.  This behavior definitely causes overhead for all BCS vector checks, that it didn't have to do before.

### Test Plan
Wrote a test, plus the existing tests helped check.

### DIFF
--- a/src/transactions/transactionBuilder/remoteAbi.ts
+++ b/src/transactions/transactionBuilder/remoteAbi.ts
@@ -117,8 +117,7 @@ export function checkOrConvertArgument(
   // If the argument is bcs encoded, we can just use it directly
   if (isEncodedEntryFunctionArgument(arg)) {
     // Ensure the type matches the ABI
-    checkType(param, arg, position);
-    return arg;
+    return checkType(param, arg, position);
   }
 
   // If it is not BCS encoded, we will need to convert it with the ABI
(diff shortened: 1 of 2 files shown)


src/transactions/transactionBuilder/remoteAbi.ts
```diff
@@ -262,89 +261,91 @@ function parseArg(
...
-      if (arg.values.length > 0) {
-        checkType(param.value, arg.values[0], position);
-      }
+      const newValues = arg.values.map((item) => checkType(param.value, item, position));
+      return new MoveVector(newValues);
+    }
 
-      return;
+    // Special case string as vector<u8>
+    if (arg instanceof MoveString && param.value.isU8()) {
```
0xmaayan, line 321: it still doesnt solve the user simple input `["hello"]`. I think the problem is that we try to convert it into Hex if the type is `vector<u8>` and the arg is `string`. We might want to try do it, and if it doesnt succeed just parse it as a string
https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/transactions/transactionBuilder/remoteAbi.ts#L207

gregnazario: That's super tricky in itself.  What if you want to put the string `0x12345`, and not the bytes `0x12345`?

This variable behavior can be really confusing for a user, and I'd prefer not to have something that chooses this behavior arbitrarily

Similarly, if you try to pass in an address string, say the address "0x1", and you want to store it as a string, and not as 1 byte

0xmaayan: I feel we mostly expect the users the understand what is right or wrong here. There are some components involved like the right way for a move function to ask for argument types (instead of `vector<vector<u8>>` use `vector<vector<String>>` if expected a string) unless you are intently want to accept  any `u8` value and expect any type to be serialized before).

gregnazario:
Closing in favor of the easy "string" approach with vector<u8>

PR #237, opened 2023-12-13
[Aviraj] Added pre-test condition

avirajkhare00:
### Description
I have added a pre test condition in Jest. If Aptos local node is not running with indexer. Tests will not start

### Test Plan
Run `pnpm test` with and without localnode to test it out.


0xmaayan (commented):
PR description says that if Indexer is not up, test wont run - but the code actually check if the node is not up.

Also, not sure `ps -ax | grep aptos` is compatible with windows - might be only in Unix-like OS.

I think a better check would be to actually ping or check whether `localhost:8080` is up and if, for example, `localhost:8080/v1` returns a response.

jest.config.js
```diff
@@ -1,26 +1,29 @@
...
-    global: {
-      branches: 50, // 90,
-      functions: 50, // 95,
-      lines: 50, // 95,
-      statements: 50, // 95,
-    },
-  },
-  // To help avoid exhausting all the available fds.
-  maxWorkers: 4,
+	preset: "ts-jest",
```
CapCap, line 3: please use space indent instead of tabs, and please align your linter with existing file, thanks!

avirajkhare00:
Hey @CapCap. Please check 🙏

0xmaayan:
Closing since running local testnet with indexer process while running tests was implemented in the sdk

PR #239, opened 2023-12-14
Fix readme file

0xmaayan:

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,11 +83,38 @@ const account = Account.generate({ scheme: SingingSchemeInput.Ed25519, legacy: f
 #### Derive from private key
 
 ```ts
-const aptos = new Aptos();
-// This functions resolves the provided private key type and derives the public key from it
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -83,11 +83,29 @@ const account = Account.generate({ scheme: SingingSchemeInput.Ed25519, legacy: f
 #### Derive from private key
 
 ```ts
-const aptos = new Aptos();
-// This functions resolves the provided private key type and derives the public key from it
+// Create a private key instance based on the scheme
+const privateKey = new Ed25519PrivateKey("myprivatekeystring") | new Secp256k1PrivateKey("myprivatekeystring");
```
gregnazario, line 87: This may be a little confusing, since this code would really compile.  Maybe... we comment out one, or we say one is privateKey1 and one is privateKey2.

gregnazario (approved):

PR #244, opened 2023-12-19
Add support to Array in digital asset property map

0xmaayan:
### Description
Following https://github.com/aptos-labs/aptos-ts-sdk/pull/238, we can now support an array value in digital asset property map
To pass in an array value, use a string type - i.e `"[hello, 1, true, world]"`

### Test Plan
pnpm run test

### DIFF
--- a/tests/e2e/api/digitalAsset.test.ts
+++ b/tests/e2e/api/digitalAsset.test.ts
@@ -1,7 +1,7 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-import { Account, Aptos, AptosConfig, Network } from "../../../src";
+import { Account, Aptos, AptosConfig, Bool, MoveString, MoveVector, Network, U8 } from "../../../src";
 import { FUND_AMOUNT } from "../../unit/helper";
 
 const config = new AptosConfig({ network: Network.LOCAL });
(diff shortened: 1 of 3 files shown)


gregnazario (approved):

tests/e2e/api/digitalAsset.test.ts
```diff
@@ -252,9 +255,10 @@ describe("DigitalAsset", () => {
         "U256 key",
         "ADDRESS key",
         "STRING key",
+        "ARRAY KEY",
       ],
-      propertyTypes: ["BOOLEAN", "U8", "U16", "U32", "U64", "U128", "U256", "ADDRESS", "STRING"],
-      propertyValues: [true, 1, 1, 1, 1, 1, 1, bob.accountAddress, "hi"],
+      propertyTypes: ["BOOLEAN", "U8", "U16", "U32", "U64", "U128", "U256", "ADDRESS", "STRING", "ARRAY"],
+      propertyValues: [true, 1, 1, 1, 1, 1, 1, bob.accountAddress, "hi", "[hello, nice, world]"],
```
gregnazario, line 261: So, can you only provide vector<u8> as a string then?  Since byte arrays aren't allowed?

0xmaayan: this is what we allow in the remoteABI parser for `vector<u8>`. To support an array with different value types (i.e `"[hello, 1, true, world]"` we need to pass it as a string.
```
if (param.value.isU8()) {
      // We don't allow vector<u8>, but we convert strings to UTF8 uint8array
      // This is legacy behavior from the original SDK
      if (isString(arg)) {
        return MoveVector.U8(TEXT_ENCODER.encode(arg));
      }
      if (arg instanceof Uint8Array) {
        return MoveVector.U8(arg);
      }
      if (arg instanceof ArrayBuffer) {
        return MoveVector.U8(new Uint8Array(arg));
      }
    }
```

gregnazario: It's a little bit stranger than that though, it's not really an array of values, but instead a string `[hello, nice, world]`.  It doesn't parse the brackets and all into an array.

This means I can't put arbitrary bytes, but only a string in `vector<u8>`

gregnazario (approved):

PR #245, opened 2023-12-22
Update transaction option `maxGasAmount, gasUnitPrice, expireTimestamp` properties to use `number` type

0xmaayan:
### Description
`maxGasAmount, gasUnitPrice, expireTimestamp` properties should be represented as `number` type.
First,`bigint` causes serialization issues for dapps and wallets and second these property values fall within the JS number type range. 
https://aptos-org.slack.com/archives/C06BFEMLAKT/p1703193020312399

Next - once this PR lands, wallet adapter should be updated to use the new sdk version and types.

### DIFF
--- a/src/transactions/types.ts
+++ b/src/transactions/types.ts
@@ -77,9 +77,9 @@ export type AnyRawTransactionInstance = RawTransaction | MultiAgentRawTransactio
  * Optional options to set when generating a transaction
  */
 export type InputGenerateTransactionOptions = {
-  maxGasAmount?: AnyNumber;
-  gasUnitPrice?: AnyNumber;
-  expireTimestamp?: AnyNumber;
+  maxGasAmount?: number;
+  gasUnitPrice?: number;
+  expireTimestamp?: number;
   accountSequenceNumber?: AnyNumber;
 };
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 
 - [`Breaking`] Capitalize `TransactionPayloadMultiSig` type
 - Add support to Array value in digital asset property map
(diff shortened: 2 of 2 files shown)


src/transactions/types.ts
```diff
@@ -77,9 +77,9 @@ export type AnyRawTransactionInstance = RawTransaction | MultiAgentRawTransactio
  * Optional options to set when generating a transaction
  */
 export type InputGenerateTransactionOptions = {
-  maxGasAmount?: AnyNumber;
-  gasUnitPrice?: AnyNumber;
-  expireTimestamp?: AnyNumber;
+  maxGasAmount?: number;
+  gasUnitPrice?: number;
+  expireTimestamp?: number;
```
gregnazario, line 82: I guess the only issue is setting "no expiry", as well as anytime after January 19th 2038, it may fail if it's not represented as a bigint or a string...

Would a string work?

bv-aptos: JavaScript uses a 64-bit floating-point number to represent dates. Jan 19, 2038 shouldn't be affected with use of number. 

Yes a string _would_ work, but would have to be converted to number elsewhere. If your only issue was the 32-bit overflow, would be good to keep as number.

bv-aptos (approved):
Looks good to me! Thanks for making the changes!

src/transactions/types.ts
gregnazario: Okay, let's make it a number, I'm convinced since the max safe integer in microseconds is 2255.

Can we document that it must be a `SafeInteger`?  Because I guess the bounds are still the same problem anyways, where the number could be put in here negative, and that would also fail.

For a future update let's add a function that we can use to verify `isUnsignedInt` which would verify the bounds `0-> Number.MAX_SAFE_NUMBER`.  Then we won't get weird float values, and we can add it more places to provide better help with it.

gregnazario (approved):
Let's follow up with some bounds checks for parts like this.

PR #246, opened 2024-01-02
Add Aptos CLI

0xmaayan:
### Description
- Install `@aptos-labs/aptos-cli` as a dev dependency 
- Implement 'LocalNode' module to easily run a local testnet within the SDK. 
- Pre running SDK tests, we check if aptos local testnet is up - if it isn't we start local testnet node process and wait for it to run before running the SDK tests, after tests are finish we kill the node process.

Next step would be to run unit and integration tests  separately so unit tests wont need to wait for a local testnet.

In future work, we can add support to common tasks like compile and publish amove module; theoretically - we can support any cli commands we think are needed.

### Test Plan
pnpm test


src/cli/localNode.ts
```diff
@@ -0,0 +1,98 @@
...
+    let last = start;
+
+    while (!operational && start + this.MAXIMUM_WAIT_TIME_SEC > last) {
+      // eslint-disable-next-line no-await-in-loop
+      await sleep(1000);
+      // eslint-disable-next-line no-await-in-loop
+      operational = await this.checkIfProcessIsUp();
+      last = Date.now() / 1000;
+    }
+    return true;
```
gregnazario, line 78: Should we return false, or error if it doesn't reach `operational`?  Or just return `operational` at the end.

.github/actions/run-tests/action.yaml
```diff
@@ -22,21 +22,21 @@ runs:
...
+    # - run: pnpm install -g @aptos-labs/aptos-cli
+    #   shell: bash
+
+    # # Run a local testnet in the background.
+    # - run: aptos node run-local-testnet --force-restart --assume-yes --with-indexer-api --log-to-stdout >& ${{ runner.temp }}/local-testnet-logs.txt &
+    #   shell: bash
+
+    # # Wait for the local testnet to be ready by hitting the readiness endpoint.
+    # # We give it a while because the CLI will have to download some images before
+    # # actually running the local testnet, which can take a while.
```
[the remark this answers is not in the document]
0xmaayan: need to delete it, left it as a comment to make sure it passes CI

README.md
```diff
@@ -179,9 +179,10 @@ const pendingTransaction = await aptos.signAndSubmitTransaction({ signer: alice,
 
 To run the SDK tests, simply run from the root of this repository:
 
-> Note: make sure aptos local node is up and running. Take a look at the [local development network guide](https://aptos.dev/guides/local-development-network/) for more details.
+> Note: for a better experience, make sure there is no aptos local node process up and running.
```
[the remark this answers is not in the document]
0xmaayan: done

src/cli/localNode.ts
0xmaayan: we can't return cause it would "solve" the promise - so we throw

src/cli/localNode.ts
```diff
@@ -0,0 +1,105 @@
...
+      console.log(str);
+    });
+  }
+
+  /**
+   * Waits for the local testnet process to be up
+   *
+   * @returns Promise<boolean>
+   */
+  async waitUntilProcessIsUp(): Promise<boolean> {
```
gregnazario, line 66: Can probably return void promise then

gregnazario (approved):

PR #247, opened 2024-01-03
[consts] Make upper bound numbers hardcoded

gregnazario:
### Description
Doing powers of bigints doesn't work on older versions of node, and it seems to be hard to reproduce, but still occurs.  Hardcoding these should have the same behavior, without these quirks.

Fixes https://github.com/aptos-labs/aptos-wallet-adapter/issues/223

### Test Plan
TBD, making a test now with an older version of node to check

### Related Links
Related https://github.com/aptos-labs/aptos-core/pull/11540

### DIFF
--- a/src/bcs/consts.ts
+++ b/src/bcs/consts.ts
@@ -4,9 +4,11 @@
 import { Uint8, Uint16, Uint32, Uint64, Uint128, Uint256 } from "../types";
 
 // Upper bound values for uint8, uint16, uint64 and uint128
-export const MAX_U8_NUMBER: Uint8 = 2 ** 8 - 1;
-export const MAX_U16_NUMBER: Uint16 = 2 ** 16 - 1;
-export const MAX_U32_NUMBER: Uint32 = 2 ** 32 - 1;
-export const MAX_U64_BIG_INT: Uint64 = BigInt(2) ** BigInt(64) - BigInt(1);
-export const MAX_U128_BIG_INT: Uint128 = BigInt(2) ** BigInt(128) - BigInt(1);
-export const MAX_U256_BIG_INT: Uint256 = BigInt(2) ** BigInt(256) - BigInt(1);
+export const MAX_U8_NUMBER: Uint8 = 255;
+export const MAX_U16_NUMBER: Uint16 = 65535;
+export const MAX_U32_NUMBER: Uint32 = 4294967295;
+export const MAX_U64_BIG_INT: Uint64 = BigInt("18446744073709551615");
+export const MAX_U128_BIG_INT: Uint128 = BigInt("340282366920938463463374607431768211455");
+export const MAX_U256_BIG_INT: Uint256 = BigInt(
+  "115792089237316195423570985008687907853269984665640564039457584007913129639935",
+);
--- a/src/bcs/consts.ts
+++ b/src/bcs/consts.ts
@@ -3,12 +3,12 @@
 
 import { Uint8, Uint16, Uint32, Uint64, Uint128, Uint256 } from "../types";
 
-// Upper bound values for uint8, uint16, uint64 and uint128
+// Upper bound values for uint8, uint16, uint64 etc.  These are all derived as
+// 2^N - 1, where N is the number of bits in the type.
 export const MAX_U8_NUMBER: Uint8 = 255;
 export const MAX_U16_NUMBER: Uint16 = 65535;
 export const MAX_U32_NUMBER: Uint32 = 4294967295;
-export const MAX_U64_BIG_INT: Uint64 = BigInt("18446744073709551615");
-export const MAX_U128_BIG_INT: Uint128 = BigInt("340282366920938463463374607431768211455");
-export const MAX_U256_BIG_INT: Uint256 = BigInt(
-  "115792089237316195423570985008687907853269984665640564039457584007913129639935",
-);
+export const MAX_U64_BIG_INT: Uint64 = 18446744073709551615n;
+export const MAX_U128_BIG_INT: Uint128 = 340282366920938463463374607431768211455n;
+export const MAX_U256_BIG_INT: Uint256 =
+  115792089237316195423570985008687907853269984665640564039457584007913129639935n;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 
 # Unreleased
 
+- Update BigInt constants to be hardcoded rather than use Math.pow
 - [`Breaking`] Capitalize `TransactionPayloadMultiSig` type
 - Add support to Array value in digital asset property map
 - [`Breaking`] Change `maxGasAmount, gasUnitPrice and expireTimestamp` properties in `InputGenerateTransactionOptions` type to `number` type
--- a/tests/unit/deserializer.test.ts
+++ b/tests/unit/deserializer.test.ts
@@ -74,20 +74,20 @@ describe("BCS Deserializer", () => {
 
   it("deserializes a uint64", () => {
     let deserializer = new Deserializer(new Uint8Array([0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff]));
-    expect(deserializer.deserializeU64()).toEqual(BigInt("18446744073709551615"));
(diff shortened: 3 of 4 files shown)


src/bcs/serializable/moveStructs.ts
```diff
@@ -52,7 +54,7 @@ export class MoveVector<T extends Serializable & EntryFunctionArgument>
 {
   public values: Array<T>;
 
-  constructor(values: Array<T>) {
+  constructor(values: ToArray<T>) {
```
gregnazario, line 57: This isn't supposed to be here.

0xmaayan:
are we sure this is the fix? Looks like adapter v1 works fine, and that version works with the legacy sdk that actually uses the same consts
https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/bcs/consts.ts

gregnazario:
> are we sure this is the fix? Looks like adapter v1 works fine, and that version works with the legacy sdk that actually uses the same consts https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/bcs/consts.ts

Ah... it is possible this happens on the old SDK as well.  In the case of this particular individual no version of the wallet adapter worked for them.  So, this would explain that.

0xmaayan:
> > are we sure this is the fix? Looks like adapter v1 works fine, and that version works with the legacy sdk that actually uses the same consts https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/bcs/consts.ts
> 
> Ah... it is possible this happens on the old SDK as well. In the case of this particular individual no version of the wallet adapter worked for them. So, this would explain that.

that's weird... I mean it works for others. It might be an issue with a specific prod env..?

gregnazario:
> > > are we sure this is the fix? Looks like adapter v1 works fine, and that version works with the legacy sdk that actually uses the same consts https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/bcs/consts.ts
> > 
> > 
> > Ah... it is possible this happens on the old SDK as well. In the case of this particular individual no version of the wallet adapter worked for them. So, this would explain that.
> 
> that's weird... I mean it works for others. It might be an issue with a specific prod env..?

That is possible.  I'm working with him on it, but this also happened to me once, I don't remember how (then I used create-aptos-dapp and it was all fine).

I'm leaning towards this because it's a simple fix that could prevent some users from even running into this at all, since there are so many different platforms to support

0xmaayan:
> > > > are we sure this is the fix? Looks like adapter v1 works fine, and that version works with the legacy sdk that actually uses the same consts https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/bcs/consts.ts
> > > 
> > > 
> > > Ah... it is possible this happens on the old SDK as well. In the case of this particular individual no version of the wallet adapter worked for them. So, this would explain that.
> > 
> > 
> > that's weird... I mean it works for others. It might be an issue with a specific prod env..?
> 
> That is possible. I'm working with him on it, but this also happened to me once, I don't remember how (then I used create-aptos-dapp and it was all fine).
> 
> I'm leaning towards this because it's a simple fix that could prevent some users from even running into this at all, since there are so many different platforms to support.
> 
> I think... I was using netlify and moved to vercel, then it just worked.

yes I think your issue was with Netlify (tho you probably could have configured a specific node version to use) - we use BigInt a lot in the sdk, would it make more sense to simply say that in order to use the SDK is required to use Node version X and above? where x is the minimum version that supports BigInt

gregnazario:
> > > > > are we sure this is the fix? Looks like adapter v1 works fine, and that version works with the legacy sdk that actually uses the same consts https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/bcs/consts.ts
> > > > 
> > > > 
> > > > Ah... it is possible this happens on the old SDK as well. In the case of this particular individual no version of the wallet adapter worked for them. So, this would explain that.
> > > 
> > > 
> > > that's weird... I mean it works for others. It might be an issue with a specific prod env..?
> > 
> > 
> > That is possible. I'm working with him on it, but this also happened to me once, I don't remember how (then I used create-aptos-dapp and it was all fine).
> > I'm leaning towards this because it's a simple fix that could prevent some users from even running into this at all, since there are so many different platforms to support.
> > I think... I was using netlify and moved to vercel, then it just worked.
> 
> yes I think your issue was with Netlify (tho you probably could have configured a specific node version to use) - we use BigInt a lot in the sdk, would it make more sense to simply say that in order to use the SDK is required to use Node version X and above? where x is the minimum version that supports BigInt

Well that's where it's weird...

It supports BigInt, it just doesn't support the power `Bigint(5) ** Bigint(2)`.  Node 10+ supports BigInt, and a similar issue was here https://github.com/rollup/rollup/issues/2626

And it looks like it needs to be ES7+ to support native `**` exponent over pow https://www.designcise.com/web/tutorial/what-is-the-math-pow-alternative-for-bigint-values-in-javascript

It's one of those things though, Netlify didn't really let you configure some pieces, and that causes a headache depending on your deployment platform, that isn't something you necessarily have control over

0xmaayan:
> > > > > > are we sure this is the fix? Looks like adapter v1 works fine, and that version works with the legacy sdk that actually uses the same consts https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/bcs/consts.ts
> > > > > 
> > > > > 
> > > > > Ah... it is possible this happens on the old SDK as well. In the case of this particular individual no version of the wallet adapter worked for them. So, this would explain that.
> > > > 
> > > > 
> > > > that's weird... I mean it works for others. It might be an issue with a specific prod env..?
> > > 
> > > 
> > > That is possible. I'm working with him on it, but this also happened to me once, I don't remember how (then I used create-aptos-dapp and it was all fine).
> > > I'm leaning towards this because it's a simple fix that could prevent some users from even running into this at all, since there are so many different platforms to support.
> > > I think... I was using netlify and moved to vercel, then it just worked.
> > 
> > 
> > yes I think your issue was with Netlify (tho you probably could have configured a specific node version to use) - we use BigInt a lot in the sdk, would it make more sense to simply say that in order to use the SDK is required to use Node version X and above? where x is the minimum version that supports BigInt
> 
> Well that's where it's weird...
> 
> It supports BigInt, it just doesn't support the power `Bigint(5) ** Bigint(2)`. Node 10+ supports BigInt, and a similar issue was here [rollup/rollup#2626](https://github.com/rollup/rollup/issues/2626)
> 
> And it looks like it needs to be ES7+ to support native `**` exponent over pow https://www.designcise.com/web/tutorial/what-is-the-math-pow-alternative-for-bigint-values-in-javascript
> 
> It's one of those things though, Netlify didn't really let you configure some pieces, and that causes a headache depending on your deployment platform, that isn't something you necessarily have control over

yea... looks like you need ES10+ to use pow with BigInt, and ES7+ to use it with number primitives. tho ES10 was out in 2019 so most if not all should be using it. 
Anyway, I am fine with this change but could we add some explanation on how we got to these numbers and this pow issue

src/bcs/consts.ts
```diff
@@ -4,9 +4,11 @@
...
-export const MAX_U8_NUMBER: Uint8 = 2 ** 8 - 1;
-export const MAX_U16_NUMBER: Uint16 = 2 ** 16 - 1;
-export const MAX_U32_NUMBER: Uint32 = 2 ** 32 - 1;
-export const MAX_U64_BIG_INT: Uint64 = BigInt(2) ** BigInt(64) - BigInt(1);
-export const MAX_U128_BIG_INT: Uint128 = BigInt(2) ** BigInt(128) - BigInt(1);
-export const MAX_U256_BIG_INT: Uint256 = BigInt(2) ** BigInt(256) - BigInt(1);
+export const MAX_U8_NUMBER: Uint8 = 255;
+export const MAX_U16_NUMBER: Uint16 = 65535;
+export const MAX_U32_NUMBER: Uint32 = 4294967295;
+export const MAX_U64_BIG_INT: Uint64 = BigInt("18446744073709551615");
```
0xmaayan, line 10: nit
```suggestion
export const MAX_U64_BIG_INT: Uint64 = 18446744073709551615n;
```

0xmaayan (approved):

PR #250, opened 2024-01-05
[coin] transferCoinTransaction sender type change to AccountAddressInput

hariria:
### Description
Follow up on [slack thread](https://aptos-org.slack.com/archives/C05NLAKJM9U/p1704496831185589) sender does not need to be an `Account` type, it can just be an `AccountAddressInput`. It's helpful to change it to an address in the event you don't have access to the user's private key.

### Test Plan
```bash
pnpm lint && pnpm unit-test && pnpm fmt && pnpm build
```


gregnazario (approved):

0xmaayan (approved):
LGTM

PR #255, opened 2024-01-17
`TransactionWorker` to fire events a dapp can listen to

0xmaayan:
### Description
The transaction management is a send-and-forget process in a way that once transactions are sent to worker, the dapp is not aware of what is going on unless a potential error happens.

This PR adds logic for the `TransactionWorker` module to fire events during a worker task and dapp can listen to these events.

```
aptos.transaction.batch.on(TransactionWorkerEvents.ExecutionFinish, async (data) => {
   console.log("execution finish", data);
});
```

Note:
Currently, dapp uses `aptos.batchTransactionsForSingleAccount()` to start the worker. We now introduce a dedicated namespace under `aptos.transaction.batch` where one can use `aptos.transaction.batch.forSingleAccount()` same as before. This is for potential future enhancement of the worker and to conform with the sdk design.

### Test Plan
`pnpm test`
`batch_mint` and `batch_funds` for examples

### DIFF
--- a/examples/typescript/batch_funds.ts
+++ b/examples/typescript/batch_funds.ts
@@ -26,7 +26,7 @@ import {
   InputGenerateTransactionPayloadData,
   Network,
   NetworkToNetworkName,
-  TransactionWorkerEvents,
+  TransactionWorkerEventsEnum,
   UserTransactionResponse,
 } from "@aptos-labs/ts-sdk";
 
@@ -100,9 +100,14 @@ async function main() {
   // emit batch transactions
   senders.map((sender) => aptos.transaction.batch.forSingleAccount({ sender, data: payloads }));
 
-  aptos.transaction.batch.on(TransactionWorkerEvents.ExecutionFinish, async (data: any) => {
+  aptos.transaction.batch.on(TransactionWorkerEventsEnum.TransactionSent, async (data) => {
+    console.log("message:", data.message);
+    console.log("transaction hash:", data.transactionHash);
+  });
+
+  aptos.transaction.batch.on(TransactionWorkerEventsEnum.ExecutionFinish, async (data) => {
     // log event output
-    console.log(data);
+    console.log(data.message);
 
     // verify accounts sequence number
     const accounts = senders.map((sender) => aptos.getAccountInfo({ accountAddress: sender.accountAddress }));
--- a/examples/typescript/batch_funds.ts
+++ b/examples/typescript/batch_funds.ts
@@ -26,6 +26,7 @@ import {
   InputGenerateTransactionPayloadData,
   Network,
   NetworkToNetworkName,
+  TransactionWorkerEvents,
   UserTransactionResponse,
 } from "@aptos-labs/ts-sdk";
 
@@ -35,7 +36,7 @@ const config = new AptosConfig({ network: APTOS_NETWORK });
 const aptos = new Aptos(config);
 
 async function main() {
-  const accountsCount = 1;
+  const accountsCount = 2;
   const transactionsCount = 10;
   const totalTransactions = accountsCount * transactionsCount;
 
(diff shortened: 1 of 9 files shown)


0xaptosj (approved):

src/api/transactionSubmission/management.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    for (const d of data) {
+      this.transactionWorker.push(d, options);
+    }
+  }
+
+  /**
+   * Internal function to start listening to transaction worker events
+   *
+   * TODO - should we ask events to listen to as an input?
```
banool, line 56: Unless it's really expense to emit events it might not be worth the complexity.

src/transactions/management/transactionWorker.ts
```diff
@@ -22,9 +23,16 @@ import { InputGenerateTransactionOptions, InputGenerateTransactionPayloadData, S
...
-const promiseFulfilledStatus = "fulfilled";
+export const promiseFulfilledStatus = "fulfilled";
 
-export class TransactionWorker {
+export enum TransactionWorkerEvents {
+  TransactionSent = "transactionSent",
+  TransactionSendFailed = "transactionsendFailed",
+  TransactionExecuted = "transactionExecuted",
+  TransactionExecutionFailed = "transactionexecutionFailed",
+  ExecutionFinish = "executionFinish",
```
banool, line 33:
```suggestion
  TransactionSent = "transactionSent",
  TransactionSendFailed = "transactionSendFailed",
  TransactionExecuted = "transactionExecuted",
  TransactionExecutionFailed = "transactionExecutionFailed",
  ExecutionFinish = "executionFinish",
```

src/transactions/management/transactionWorker.ts
```diff
@@ -157,12 +165,21 @@ export class TransactionWorker {
             // transaction sent to chain
             this.sentTransactions.push([sentTransaction.value.hash, sequenceNumber, null]);
             // check sent transaction execution
+            this.emit(
+              TransactionWorkerEvents.TransactionSent,
+              `transaction hash ${sentTransaction.value.hash} has been commited to chain`,
```
banool, line 170:
```suggestion
              `transaction hash ${sentTransaction.value.hash} has been committed to chain`,
```

banool, line 170: Is this the data being emitted? If so, shouldn't we emit something structured instead? That way the callback can actually do something with the data. So it emits some kind of `TransactionSentEventData` that has a `txnHash` field. The user can use that in the callback too rather than `data: any`.

banool (changes requested):

src/transactions/management/transactionWorker.ts
0xmaayan: Good call! I have updated the worker/management to have a types returned data

src/transactions/management/transactionWorker.ts
```diff
@@ -22,9 +23,38 @@ import { InputGenerateTransactionOptions, InputGenerateTransactionPayloadData, S
...
+  ExecutionFinish = "executionFinish",
+}
+
+export interface TransactionWorkerEvents {
+  transactionSent: (data: SuccessEventData) => void;
+  transactionSendFailed: (data: FailureEventData) => void;
+  transactionExecuted: (data: SuccessEventData) => void;
+  transactionExecutionFailed: (data: FailureEventData) => void;
+  executionFinish: (data: ExecutionFinishEventData) => void;
+}
```
kaw2k, line 42: Ok, just double checking I understand each event:
- `transactionSent` is fired after a transaction gets sent to the chain
- `transactionSendFailed` is fired if there is an error sending the transaction to the chain
- `transactionExecuted` is fired when a single transaction has run successfully
- `transactionExecutionFailed` is run if a single transaction fails in execution
- `executionFinish` is run when the queue has been emptied

So if someone wanted to make an async function to batch requests, it would look like this:


```ts
// Pseudo code, returns txn_hash array
type TxnHash = string;
function batchTransactionExample(sender: Account, payload: InputGenerateTransactionPayloadData[]): Promise<{
	succeeded_txns: TxnHash[],
	failed_txns: TxnHash[]
}> {
	return new Promise(async (resolve, reject) => {
		let succeeded_txns = [];
		let failed_txns = [];
		
		aptos.transaction.batch.forSingleAccount({sender, data: payload})

		// Handle transaction successes, we don't care about sent, just executed
		aptos.transaction.batch.on(TransactionWorkerEvents.TransactionExecuted, async (data: any) => {
			succeeded_txns.push(data.txn_hash);
		});

		// Handle transaction failures (execution or send)
		aptos.transaction.batch.on(TransactionWorkerEvents.TransactionExecutionFailed, async (data: any) => {
			failed_txns.push(data.txn_hash);
		});
		aptos.transaction.batch.on(TransactionWorkerEvents.TransactionSendFailed, async (data: any) => {
			failed_txns.push(data.txn_hash);
		});

		// Handle resolving the promise
		aptos.transaction.batch.on(TransactionWorkerEvents.ExecutionFinish, async (data: any) => {
			resolve({ succeeded_txns, failed_txns })
		});
	})
}
```
I'm guessing the caller will have to unsubscribe from the emitter? That is something we would want to highlight in the two examples potentially.

0xmaayan: yep! I'll add some doc comments - good catch!

examples/typescript/batch_funds.ts
```diff
@@ -97,8 +98,31 @@ async function main() {
...
+    const accountsData = await Promise.all(accounts);
+    accountsData.forEach((accountData) => {
+      console.log(
+        `account sequence number is ${(totalTransactions * senders.length) / 2}: ${
+          accountData.sequence_number === "20"
+        }`,
+      );
+    });
+    // worker finished execution, we can now unsubscribe from event listeners
+    aptos.transaction.batch.removeAllListeners();
```
kaw2k, line 123: ❤️

kaw2k (approved):
Thanks for getting this in!!!

banool (approved):

PR #257, opened 2024-01-18
Add transaction submission through servers communication example

0xmaayan:
### Description
This PR adds examples for
- how one can generate and sign a transaction on one server (for example, frontend), have it serialized and send it to another server (for example, backend) to deserialize and submit
server A -> server B -> chain

- how one can generate a transaction on one server (for example, frontend), have it serialized and send it to another server (for example, backend) to deserialize and sign it and then send it back to the other server (the frontend server) to submit it
server A -> server B -> server A -> chain


davidiw (commented):
if I want to send something to the server to sign, what is it?

0xmaayan:
> if I want to send something to the server to sign, what is it?

server as the fee payer? you would need to send the raw transaction as bcs serialized, have the server to deserialize it and sign it. Then the server either submits the transaction or sends the serialized transaction and the serialized fee payer account address to the other server to submit. 

That is a different flow, I had an example for that flow also, can put it on a another PR

PR #262, opened 2024-01-24
[faucet] Remove waitForIndexer for faucet fundAccount

hariria:
### Description
Unneeded waitForIndexer call for faucet.

### Test Plan
`pnpm lint && pnpm unit-test`
Got the same number of errors locally (before and after any changes) running `pnpm api-test` 🤷

### DIFF
--- a/src/api/faucet.ts
+++ b/src/api/faucet.ts
@@ -5,7 +5,6 @@ import { fundAccount } from "../internal/faucet";
 import { UserTransactionResponse, WaitForTransactionOptions } from "../types";
 import { AccountAddressInput } from "../core";
 import { AptosConfig } from "./aptosConfig";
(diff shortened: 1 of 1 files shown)


0xmaayan:
`pnpm test` includes both `pnpm unit-test` and `pnpm api-test`, tho for this change `pnpm api-test` is enough as this piece of code is only tested in api-tests.

Local tests should not fail locally. If it does it might be because localnet (probably a indexer processor) breaks and need to restart it.

I am not fully convinced `waitForIndexer` is not needed. If indexer doesnt catch up it might be something wrong with indexer. Also, you should be able to disable this default behavior with  `options: { waitForIndexer: false }`

hariria:
Ok fair. ill close

PR #272, opened 2024-02-06
Fix paginateWithCursor function

### LINKED ISSUE #254: [Bug] Account Transactions doesn't listen to pagination limits
# 🐛 Bug Description

Pagination inputs aren't applied for `getAccountTransactions`

## How to reproduce

**Code snippet to reproduce**
```typescript
import {AccountAddress, Aptos} from "@aptos-labs/ts-sdk"
const aptos = new Aptos();
const accountAddress = AccountAddress.from("0x1");
const txns = await aptos.getAccountTransactions({accountAddress, options: {offset: 5, limit: 1});

// txns will just give the last 10 transactions no matter what.
```

**Stack trace / error message**
```
// Paste the output here
```

## System information

**System details:**
- Typescript SDK Version
- Platform (e.g. Node, browser, etc.)

etkinpinar:
### Description
Since assignment of requestParams.start to the cursor variable occurs at the beginning of the while loop, it overwrites the start value given as a parameter, I fixed it by moving the assignment to the end.

Fixes #254


0xmaayan:
@etkinpinar Thanks for the fix! Could you also add tests for it? Can add a new test case [here](https://github.com/aptos-labs/aptos-ts-sdk/blob/main/tests/e2e/api/paginateQuery.test.ts), something like `test("it should paginate correctly on fullnode queries")`

0xmaayan (approved):

PR #280, opened 2024-02-08
Support automated account creation for sponsored transactions

0xmaayan:
### Description
[AIP-52](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-52.md) introduced automated account creation for sponsored transactions. This PR supports this change by if the transaction is a sponsored transaction, check the main signer chain info. If it is not found, assign the sequence number to 0.

### Test Plan
`pnpm test`

### DIFF
--- a/src/transactions/transactionBuilder/transactionBuilder.ts
+++ b/src/transactions/transactionBuilder/transactionBuilder.ts
@@ -74,7 +74,7 @@ import {
 } from "../types";
 import { convertArgument, fetchEntryFunctionAbi, standardizeTypeTags } from "./remoteAbi";
 import { memoizeAsync } from "../../utils/memoize";
-import { SigningScheme } from "../../types";
+import { AnyNumber, SigningScheme } from "../../types";
 import { getFunctionParts, isScriptDataInput } from "./helpers";
 
 /**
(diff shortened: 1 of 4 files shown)


src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -217,9 +213,33 @@ export async function generateRawTransaction(args: {
...
+      ? Promise.resolve({ sequence_number: options.accountSequenceNumber })
+      : getInfo({ aptosConfig, accountAddress: sender });
+
+  let sequenceNumber: string | AnyNumber;
+
+  /**
+   * Check if is sponsored transaction to honor AIP-52
+   * {@link https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-52.md}
+   */
+  if (feePayerAddress === AccountAddress.ZERO.toString()) {
```
gregnazario, line 229: I think fee payer address doesn't necessarily have to be a string?

I think you need to resolve the fee payer address prior to the comparison.

e.g.
```
AccountAddress.from(feePayerAddress) === AccountAddress.ZERO
```

gregnazario (approved):

src/transactions/transactionBuilder/transactionBuilder.ts
0xmaayan: updated

PR #284, opened 2024-02-08
Bump version to `1.6.0`

0xmaayan:


gregnazario (approved):

0xmaayan:
> didn't look much at added files, modifications seem reasonable.
> 
> side thought is, would it be possible to only include these `*.html` files locally as opposed to checking them in? Asking because these look like generated files

good question. I am not completely sure how it works if it is possible since eventually those files should be live here https://aptos-labs.github.io/aptos-ts-sdk/

gregnazario:
> didn't look much at added files, modifications seem reasonable.
> 
> side thought is, would it be possible to only include these `*.html` files locally as opposed to checking them in? Asking because these look like generated files

Github pages is hosting these statically.  For now, it's easiest to do it this way.

We can probably also just run the doc generation on the fly, but that might require more work / not keep the old versions static to older docs

PR #293, opened 2024-02-15
[docs] Ensure there is a latest page

gregnazario:
### Description
This I think should add a latest page to the SDK following https://github.com/jekyll/jekyll-redirect-from#redirect-to

https://stackoverflow.com/questions/39657226/jekyll-redirect-link

It's a little unclear how best to test, so we'll just try it out directly.


0xaptosj (approved):
let's test in prod

PR #294, opened 2024-02-16
Strongly typed crypto proposal

hardsetting:
Simplified and cleaned up #147

Notably, I rolled back renaming the `Account` class into `Signer` to avoid a breaking change and following some naming discussions.

Will add a better description in the AM.


tests/e2e/api/account.test.ts
```diff
@@ -266,15 +257,15 @@ describe("account api", () => {
       await aptos.fundAccount({ accountAddress: account.accountAddress, amount: 1_000_000_000 });
 
       // account that holds the new key
-      const rotateToPrivateKey = Ed25519PrivateKey.generate();
+      const rotateToSigner = Account.generate({ scheme: SigningSchemeInput.Ed25519, legacy: true });
```
0xmaayan, line 260: it is not really needed to specify `{ scheme: SigningSchemeInput.Ed25519, legacy: true }` right? `Account.generate()` defaults to it

tests/e2e/transaction/helper.ts
```diff
@@ -189,14 +190,14 @@ export const rawTransactionMultiAgentHelper = async (
       transaction: generatedTransaction,
     });
     transactionResponse = await aptos.transaction.submit.multiAgent({
-      transaction: generatedTransaction,
+      transaction: generatedTransaction as MultiAgentTransaction,
```
0xmaayan, line 193: why casting?

src/core/account/account.ts
```diff
@@ -0,0 +1,139 @@
...
+import { Ed25519Account } from "./ed25519";
+import { type PrivateKeyInput, SingleKeyAccount } from "./singleKey";
+
+export interface CreateEd25519AccountFromPrivateKeyArgs {
+  privateKey: Ed25519PrivateKey;
+  address?: AccountAddressInput;
+  legacy?: true;
+}
+
+export interface CreateEd25519SingleKeyAccountFromPrivateKeyArgs {
```
0xmaayan, line 14: I think it is worth to put up some explanation here on each type. As Single Key is not well known devs might be confusing on the difference between `CreateEd25519Account` , `CreateEd25519SingleKeyAccount`, `CreateSingleKeyAccount` etc

src/core/account/account.ts
```diff
@@ -0,0 +1,139 @@
...
+  legacy: false;
+}
+
+export interface CreateSingleKeyAccountFromPrivateKeyArgs {
+  privateKey: Exclude<PrivateKeyInput, Ed25519PrivateKey>;
+  address?: AccountAddressInput;
+  legacy?: false;
+}
+
+export interface CreateAccountFromPrivateKeyArgs {
```
0xmaayan, line 26: This type overload can be confusing especially for devs that are less TS advanced - let's add a brief explanation who each type eventually resolved to `CreateAccountFromPrivateKeyArgs`

src/core/account/singleKey.ts
```diff
@@ -0,0 +1,118 @@
...
+        privateKey = Secp256k1PrivateKey.generate();
+        break;
+      default:
+        throw new Error(`Unsupported signature scheme ${scheme}`);
+    }
+    return new SingleKeyAccount({ privateKey });
+  }
+
+  /**
+   * Derives an account with bip44 path and mnemonics,
```
0xmaayan, line 79: `Default to using an Ed25519 signature scheme`

src/core/crypto/multiEd25519.ts
```diff
@@ -89,9 +93,46 @@ export class MultiEd25519PublicKey extends PublicKey {
     return Hex.fromHexInput(this.toUint8Array()).toString();
   }
 
-  // eslint-disable-next-line class-methods-use-this, @typescript-eslint/no-unused-vars
   verifySignature(args: { message: HexInput; signature: MultiEd25519Signature }): boolean {
-    throw new Error("TODO - Method not implemented.");
+    const { message, signature } = args;
```
0xmaayan, line 97: could we add a test for this new function implementation?

src/core/crypto/multiEd25519.ts
```diff
@@ -146,26 +192,33 @@ export class MultiEd25519Signature extends Signature {
    *
    * @param args.signatures A list of signatures
    * @param args.bitmap 4 bytes, at most 32 signatures are supported. If Nth bit value is `1`, the Nth
-   * signature should be provided in `signatures`. Bits are read from left to right
+   * signature should be provided in `signatures`. Bits are read from left to right.
+   * Alternatively, you can specify an array of bitmap positions.
+   * Valid position should range between 0 and 31.
```
0xmaayan, line 197: lets add this check in the constructor?

src/core/crypto/multiKey.ts
```diff
@@ -1,12 +1,39 @@
...
+/* eslint-disable no-bitwise */
+
+function bitCount(byte: number) {
+  let n = byte;
+  n -= (n >> 1) & 0x55555555;
+  n = (n & 0x33333333) + ((n >> 2) & 0x33333333);
+  return (((n + (n >> 4)) & 0xf0f0f0f) * 0x1010101) >> 24;
+}
+
+/* eslint-enable no-bitwise */
```
0xmaayan, line 22: remove?

src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -321,90 +326,81 @@ export async function buildTransaction(args: InputGenerateRawTransactionArgs): P
...
-    );
+  return new SignedTransaction(transaction.rawTransaction, transactionAuthenticator).bcsToBytes();
+}
 
-    const transactionAuthenticator = new TransactionAuthenticatorMultiAgent(
-      accountAuthenticator,
-      transaction.secondarySignerAddresses,
-      secondaryAccountAuthenticators,
-    );
+export function getEmptySignature(publicKey: AnyPublicKey) {
```
0xmaayan, line 368: add the function signature return type.

src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -321,90 +326,81 @@ export async function buildTransaction(args: InputGenerateRawTransactionArgs): P
...
+    return AnySignature.fromSignature(new Ed25519Signature(new Uint8Array(64)));
+  }
+  if (publicKey.publicKey instanceof Secp256k1PublicKey) {
+    return AnySignature.fromSignature(new Secp256k1Signature(new Uint8Array(64)));
+  }
+  throw new Error("Unexpected public key type");
+}
 
-    return new SignedTransaction(transactionToSign.raw_txn, transactionAuthenticator).bcsToBytes();
+export function getAuthenticatorForSimulation(publicKey: PublicKeyInput) {
```
0xmaayan, line 378: same as above

src/transactions/types.ts
```diff
@@ -226,15 +225,15 @@ export type InputSimulateTransactionData = {
   /**
    * For a single signer transaction
    */
-  signerPublicKey: PublicKey;
+  signerPublicKey: PublicKeyInput;
   /**
    * For a fee payer or multi-agent transaction that requires additional signers in
    */
-  secondarySignersPublicKeys?: Array<PublicKey>;
+  secondarySignersPublicKeys?: PublicKeyInput[];
```
0xmaayan, line 232: as mentioned in the previous PR - there is no firm reason this syntax was chosen, but since we already have it throughout the SDK we might as well keep going with it

```suggestion
  secondarySignersPublicKeys?: Array<PublicKeyInput>;
```

0xmaayan (commented):
Very nice! 

CHANGELOG updates is missing

tests/e2e/api/account.test.ts
hardsetting: Yeah that's correct. I actually rolled back that part as well

tests/e2e/transaction/helper.ts
hardsetting: Because `submit.multiAgent` requires a `MultiAgentTransaction` that includes secondary signers.
If it's not, then you should use `submit.simple`.

In this case `generatedTransaction` can be either type

src/transactions/types.ts
hardsetting: ok sure

src/core/crypto/multiKey.ts
hardsetting: That's needed.
There's a global rule for not using bitwise operations, but we need to use them in that function.
Alternatively, we can just globally remove the rule?

examples/typescript/batch_mint.ts
```diff
@@ -50,7 +50,7 @@ async function main() {
   // First need to create a collection on chain
   console.log("creating collection...");
   const transaction = await aptos.createCollectionTransaction({
-    creator: sender,
+    creator: sender.accountAddress,
```
heliuchuan, line 53: So I would suggest instead of this, add Account to the AccountAddressInput type.

Then update the from function in accountAddress.ts

static from(input: AccountAddressInput): AccountAddress {
    if (input instanceof Account) {
      return input.accountAddress;
    }
    if (input instanceof AccountAddress) {
      return input;
    }
    if (input instanceof Uint8Array) {
      return new AccountAddress(input);
    }
    return AccountAddress.fromString(input);
  }

examples/typescript/external_signing.ts
```diff
@@ -47,7 +48,7 @@ const balance = async (aptos: Aptos, account: Account, name: string): Promise<nu
  * Provides a mock "Cold wallet" that's signed externally from the SDK
  */
 class ExternalSigner {
-  private account: Account;
+  private account: Ed25519Account;
```
heliuchuan, line 51: why does this need to change to Ed25519Account?

examples/typescript/batch_mint.ts
hardsetting: I like it

hardsetting: Using `instanceof` requires importing the full type, and causes a cylical dependency :(

hardsetting: I'm doing this instead, should still work:
```typescript
  static from(input: AccountAddressInput): AccountAddress {
    if (input instanceof AccountAddress) {
      return input;
    }
    if (input instanceof Uint8Array) {
      return new AccountAddress(input);
    }
    if (typeof input === "string") {
      return AccountAddress.fromString(input);
    }
    return input.accountAddress;
  }
```

hardsetting: Adding the `Account` requires changing the code in a lot of places specifically before generating an entry function payload where we're passing an `AccountAddressInput`.

Both `AccountAddress` and `string` are valid entry function payload argument inputs, but if we add `Account` it starts complaining.
To solve that we would either have to:
- Use `AccountAddress.from` every time we're passing an `AccountAddressInput` to the function arguments
  I don't think we should go down this path as it adds more complexity especially for the end user.

- Add support for `Account` in the layer that takes care of generating / submitting an entry function payload.
  This could be a good option, but I'm not familiar with that part of the code and worry about the time needed. Happy to give it a shot later.

examples/typescript/external_signing.ts
hardsetting: Looking at the code below (not in the changeset) `AccountAuthenticatorEd25519` is used specifically.
There's also a hard cast for the public key to `Ed25519PublicKey`.
I thought it was unnecessary to keep the property as the generic `Account`.

src/core/crypto/multiEd25519.ts
hardsetting: The constructor calls `createBitmap` which already performs the check

examples/typescript/external_signing.ts
heliuchuan: I guess in my mind, an ExternalSigner interface need not commit to any particular signing scheme.

I think similar to Account, we should provide a generic constructor for AccountAuthenticator and update the example to be generic to the scheme.

examples/typescript/batch_mint.ts
heliuchuan: Ok yeah, kinda annoying.  Don't worry about it within the scope of this PR.

Do note, this does still constitute a breaking change, but its trivial.

src/core/crypto/singleKey.ts
```diff
@@ -0,0 +1,194 @@
...
+    return AuthenticationKey.fromSchemeAndBytes({
+      scheme: this.authenticationScheme,
+      input: this.bcsToBytes(),
+    });
+  }
+
+  toUint8Array() {
+    // TODO (gabi): should this be the BCS-serialized key instead?
+    return this.publicKey.toUint8Array();
+  }
```
hariria, line 79: I think we should use BCS-serialization for all of the `toUint8Array()` functions to be consistent. And then we can add a `toPublicKeyBytes()` as well. Wdy think?

hariria (commented):
I think the naming of some of the types are a bit confusing. Some may have been there from before this PR, but it would be great to clarify them if possible.

src/core/account/account.ts
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * Note: Generating an account instance does not create the account on-chain.
+ */
+export abstract class Account {
+  /**
+   * Private key associated with the account.
+   * Note: this will be removed in the next major release,
+   *  as not all accounts have a private key.
+   */
+  abstract readonly privateKey: PrivateKey;
```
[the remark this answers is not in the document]
hardsetting: Mmm good point, I also thought about it.
Although it would be `CryptoPrivateKey` as it would not be associated to a specific authentication scheme.
The only possible private keys are `Ed25519PrivateKey` and `Secp256k1PrivateKey`

src/core/account/account.ts
```diff
@@ -0,0 +1,252 @@
...
+   * @return the AccountAuthenticator containing the signature, together with the account's public key
+   */
+  abstract signWithAuthenticator(message: HexInput): AccountAuthenticator;
+
+  /**
+   * Sign the given message with the private key.
+   * @param message in HexInput format
+   * @returns AccountSignature
+   */
+  abstract sign(message: HexInput): AccountSignature;
```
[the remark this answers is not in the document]
hardsetting: I like this.
Initially the plan was to just use authenticators in general (even for signing messages).
And to eventually just have a single `sign` that returns the authenticator

But tbh that kind of makes little sense.

src/core/crypto/publicKey.ts
```diff
@@ -0,0 +1,108 @@
...
+ * generated by an account and associated to a specific authentication scheme.
+ */
+export abstract class AccountPublicKey extends Serializable {
+  /**
+   * The authentication scheme associated with the public key.
+   *
+   * This property can be used at runtime to match
+   * the authentication scheme of a given AccountSignature.
+   */
+  abstract readonly authenticationScheme: AuthenticationKeyScheme;
```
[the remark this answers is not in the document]
hardsetting: As discussed offline `AccountAuthenticator` wouldn't work imo, since you'd expect an `AccountAuthenticator` instance.

I agree all the schemes out there are really confusing and we should clean that up at some point.
Specifically SigninScheme vs SigningSchemeInput are conceptually very different but sound very similar.

`AuthenticationScheme` makes sense to me, as it's the scheme used to derive an authentication key from the public key bytes.

```
- AuthenticationScheme.Ed25519
Ed25519Account
Ed25519PublicKey
Ed25519Signature
AccountAuthenticatorEd25519

- AuthenticationScheme.MulitEd25519
MulitEd25519PublicKey
MulitEd25519Signature
AccountAuthenticatorMulitEd25519

- AuthenticationScheme.SingleKey
SingleKeyAccount
AnyPublicKey
AnySignature
AccountAuthenticatorSingleKey

- AuthenticationScheme.MultiKey
MultiKey
MultiKeySignature
AccountAuthenticatorMultiKey
```

But tying this concept to authenticators also would work.
We could use `AccountAuthenticatorVariant` which is the enum used for `AccountAuthenticator`.

src/core/crypto/signature.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * This is the product of signing a message through an account,
+ * and can be verified against an AccountPublicKey.
+ */
+export abstract class AccountSignature extends Serializable {
+  abstract readonly authenticationScheme: AuthenticationKeyScheme;
+
+  /**
+   * Get the raw signature bytes
+   * TODO (gabi): do we need this?
```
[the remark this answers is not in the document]
hardsetting: I'm keeping the function around to avoid breaking changes,
but also I don't want to make a decision whether to remove it or not in this PR to avoid slowing the process.

src/core/crypto/signature.ts
```diff
@@ -0,0 +1,53 @@
...
+   */
+  abstract toUint8Array(): Uint8Array;
+
+  /**
+   * Get the signature as a hex string with a 0x prefix e.g. 0x123456...
+   * TODO (gabi): do we need this?
```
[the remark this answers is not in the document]
hardsetting: Same as above

src/core/crypto/singleKey.ts
hardsetting: we already have `bcsToBytes`, so if `toUint8Array` and `bcsToBytes` are equivalent I would just drop `toUint8Array`

If you're suggesting a renaming of `toUint8Array` in something more explanatory, that could work.
Can be addressed in another PR for renaming

hardsetting: I think in general `toUint8Array` and `bcsToBytes` give different results, since the bcs serialization is prefixed with the key length.

The question is.. what is the "bytes" representation of `AnyPublicKey`? Is it the same as the underlying crypto public key? Or should it be different?

I think it should be the former, which also plays well with authentication key derivation.
All instances of `AccountPublicKey` can then derive the associated authentication key with `AuthenticationKey.fromSchemeAndBytes(this.authenticationScheme, this.toUint8Array())`.

src/core/account/account.ts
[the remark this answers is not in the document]
hardsetting: Mmm I tried implementing this, but it's a dependency cycle mess ugh
Will maybe keep it this way for now and build on top of this PR?

src/core/account/account.ts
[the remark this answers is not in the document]
hariria: Passkeys add `Secp256r1` -- its just not included yet, there's a draft sdk PR for it though -- and ZkID may add other schemes. I personally would like it to be consistent just so its easier to find things

src/core/crypto/publicKey.ts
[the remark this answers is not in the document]
hariria: Something with `AccountAuthenticator` so it's consistent with the Rust codebase would be good. Will leave it up to you on what specifically you want to make it. For context, I had no idea what an `authenticationScheme` meant. I thought it was a `SignatureScheme` or like an HTTP authentication scheme which didn't make much sense to me. If you google "Authentication Scheme" you'll see what I mean

src/core/crypto/singleKey.ts
hariria: Either works for me, just wanted it to be consistent. I see we also have a `serialize` function on some classes which may or may not do `bcsToBytes()`? I'm not too sure. Anyways I trust you :)

src/core/crypto/singleKey.ts
```diff
@@ -0,0 +1,194 @@
+import { Deserializer, Serializer } from "../../bcs";
```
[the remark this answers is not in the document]
0xmaayan: left a comment on the `account/singleKey.ts` file

src/core/crypto/singleKey.ts
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  // endregion
+
+  // region AccountPublicKey
+
+  /**
+   * Authentication scheme
+   */
+  public readonly authenticationScheme = AuthenticationKeyScheme.SingleKey;
```
0xmaayan, line 54: can remove to the top of the class

src/core/crypto/singleKey.ts
0xmaayan: fwiw, i think `toUint8Array` is more familiar and understandable in our dev community.

I think the bytes representation of  `AnyPublicKey` and `CryptoPublicKey` is different in the rust implementation level - but i might be wrong.

src/types/index.ts
```diff
@@ -664,36 +664,36 @@ export type TransactionMultiEd25519Signature = {
 
 export type TransactionMultiAgentSignature = {
   type: "multi_agent_signature";
-  sender: AccountSignature;
+  sender: TransactionAccountSignature;
```
0xmaayan, line 667: should be marked as [`Breaking`] in changelog

tests/e2e/transaction/transactionSubmission.test.ts
```diff
@@ -606,7 +605,7 @@ describe("transaction submission", () => {
         signaturesRequired: 2,
       });
 
-      const authKey = AuthenticationKey.fromPublicKey({ publicKey: multiKey });
```
0xmaayan, line 609: is this function removed? if so, need to mention it on changelog

src/core/account/account.ts
```diff
@@ -0,0 +1,252 @@
...
+ * Interface for a generic Aptos account.
+ *
+ * The interface is defined as abstract class to provide a single entrypoint for account generation,
+ * either through `Account.generate()` or `Account.fromDerivationPath`.
+ * Despite this being an abstract class, it should be treated as an interface and enforced using
+ * the `implements` keyword.
+ *
+ * Note: Generating an account instance does not create the account on-chain.
+ */
+export abstract class Account {
```
0xmaayan, line 114: As part of the new wallet standard, we had the AccountInfo interface to hold an account of type Account
```
export interface AccountInfo = { account: Account, ansName?: string }
```
this was mostly done so it would be easier for us to resolve the account type (single signer/ multi signer) using the `variant` field - was this changed with this PR?

0xmaayan: Also, where did `AptosAccountVariant` enum go?

src/core/account/account.ts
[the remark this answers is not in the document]
hardsetting: Yes that makes sense

src/core/crypto/signature.ts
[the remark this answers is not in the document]
hardsetting: There's already `bcsToBytes()` for that purpose.
Transmitting the raw bytes of the signature without any indication of which signature scheme it's based off could also be confusing.

src/core/crypto/singleKey.ts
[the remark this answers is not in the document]
hardsetting: Confusing in which way? What would you suggest?

src/core/crypto/singleKey.ts
hardsetting: You mean move?
`AnyPublicKey` extends `AccountPublicKey` and it is required to implement some methods and getters.
I put all such members within the `AccountPublicKey` region so it's nicely organized.
Similarly there's a `Serializable` region

src/core/crypto/singleKey.ts
hardsetting: >The question is.. what is the "bytes" representation of AnyPublicKey? Is it the same as the underlying crypto public key? Or should it be different?

@0xmaayan do you have an opinion?
Should I check in `aptos-core` to see what they do?

src/types/index.ts
hardsetting: I'm positive nobody is using this type 😆 
but can add it for good measure

tests/e2e/transaction/transactionSubmission.test.ts
hardsetting: It's still there, just marked as deprecated.
Should I still mention it in the current changelog, or only when we decide to fully remove it?

.eslintrc.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
     "max-classes-per-file": ["error", 10],
     "import/prefer-default-export": "off",
     "object-curly-newline": "off",
+    // Replacing airbnb rule with following, to re-enable "ForOfStatement"
```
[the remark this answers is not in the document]
hardsetting: Oh I think you added the line below in a different PR (but with no comment).
So after rebasing on `main`, only the comment is left.
I think a comment explaining what the rule override does is nice, but I don't feel strongly about keeping / removing

src/core/account/account.ts
hardsetting: I got some feedback regarding having both `Signer` and `Account` from @heliuchuan .
The argument is that it would be a breaking change, and the `Signer` name wouldn't be as intuitive.

So in this PR we only have `Account` and it has signing capabilities.
I think that for `AccountInfo` we can have the following struct, which would be functionally equivalent and similar to what we had before:

```
export interface AccountInfo = { address: AccountAddress, publicKey: AccountPublicKey, ansName?: string }
```
would also be nice to make it BCS serializable.

0xmaayan: Let's update the standard AIP then. The AIP currently relies on that change we planned to have

src/core/crypto/multiEd25519.ts
```diff
@@ -146,30 +181,41 @@ export class MultiEd25519Signature extends Signature {
...
+  // region AccountSignature
+
+  readonly authenticationScheme = AuthenticationKeyScheme.MultiEd25519;
+
   /**
    * Converts a MultiSignature into Uint8Array (bytes) with `bytes = s1_bytes | ... | sn_bytes | bitmap`
    */
   toUint8Array(): Uint8Array {
+    // TODO (gabi): this serialization is different from the BCS serialization.
+    //  Is this acceptable? Is this even used?
```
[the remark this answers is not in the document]
0xmaayan: yea I think it was copy past from the legacy sdk. It is meant to convert it into bytes and just return it
https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/aptos_types/multi_ed25519.ts#L91

src/core/crypto/singleKey.ts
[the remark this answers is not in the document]
0xmaayan: confusing in when we want to search in the sdk, I left a comment on the `account/singleKey.ts` file to potentially rename it to `AccountSingleKey`

src/core/crypto/singleKey.ts
0xmaayan: sorry, I meant move to the top of the class with all other class variables definitions.

I understand the benefits of having those region comments, but I feel we implement it in some specific places and end up with the SDK not being consistent. I would argue we should have/implement the region comment helper in a new and fresh PR where we implement it throughout the SDK and not only in some random files we happen to work on in a PR

src/core/crypto/singleKey.ts
0xmaayan: as mentioned in my previous comment

I think the bytes representation of `AnyPublicKey` and `CryptoPublicKey` is different in the rust (aptos-core) implementation level - but i might be wrong, need to varify

src/types/index.ts
0xmaayan: sdk users do... wild things

src/core/crypto/singleKey.ts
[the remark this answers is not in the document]
hardsetting: yeah I'm down

src/core/crypto/signature.ts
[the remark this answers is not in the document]
hardsetting: Just checked the Rust code. `AnySignature` does not have a "to_bytes" function, although `AnyPublicKey" does.
I can just keep it around and consistent with `AnyPublicKey`, for now

src/core/authenticationKey.ts
```diff
@@ -65,69 +61,43 @@ export class AuthenticationKey extends Serializable {
...
   /**
+   * @deprecated Use `AccountPublicKey.authKey()` instead
    * Converts a PublicKey(s) to an AuthenticationKey, using the derivation scheme inferred from the
    * instance of the PublicKey type passed in.
    *
    * @param args.publicKey
    * @returns AuthenticationKey
    */
-  static fromPublicKey(args: { publicKey: PublicKey }): AuthenticationKey {
+  static fromPublicKey(args: { publicKey: AccountPublicKey }): AuthenticationKey {
```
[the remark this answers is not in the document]
hardsetting: This should not be breaking unless I'm missing something.
You should be able to pass the same input as before the change

src/core/account/ed25519.ts
```diff
@@ -0,0 +1,88 @@
...
+  signature: Ed25519Signature;
+}
+
+/**
+ * Signer implementation for the Ed25519 authentication scheme.
+ * This extends an {@link Ed25519Account} by adding signing capabilities through an {@link Ed25519PrivateKey}.
+ *
+ * Note: Generating a signer instance does not create the account on-chain.
+ */
+export class Ed25519Account implements Account {
```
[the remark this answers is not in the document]
0xmaayan: I guess this is an implementation decision question and @davidiw can help with it.

I assume we eventually would want to support by default single key scheme, for future new authentication  schemes also. 

Also, if aptos-core supports both schemes, I think the sdk should probably support it also. We already need to support single key scheme for secp and probably some in the future so why not support all single key schemes?

src/core/crypto/singleKey.ts
hardsetting: What I meant is.. which would you **personally** choose given your understanding.

hardsetting: I went ahead and looked at the Rust code.
The `to_bytes` function for `AnyPublicKey` returns the bcs serialized value.
There's no `to_bytes` function for `AnySignature`.

So I'm updating the PR accordingly

src/core/crypto/multiEd25519.ts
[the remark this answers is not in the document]
hardsetting: Just keeping as-is then, just in case.
But realistically this will probably never be used 🤞🏼

src/core/crypto/publicKey.ts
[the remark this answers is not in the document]
0xmaayan: +1 to `AccountAuthenticatorType`

0xmaayan:
@hardsetting once it lands, we would need to update the docs on [aptos.dev ](https://aptos.dev/sdks/new-ts-sdk/)

hardsetting:
> once it lands, we might need to update the docs on [aptos.dev ](https://aptos.dev/sdks/new-ts-sdk/)if there is anything specific to call out

there's no breaking change at this point

hardsetting:
Gathered all the feedback, and just rolled back more changes from the original PR.
The only notable change in this PR is now just splitting the monolithic `Account` class into abstract class + implementations

src/core/crypto/publicKey.ts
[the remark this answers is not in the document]
hardsetting: after some discussion, we opted to just drop the field

hariria (approved):
LGTM pending Oliver's feedback

0xmaayan (approved):
LGTM!

heliuchuan (approved):

src/core/account/Account.ts
```diff
@@ -0,0 +1,245 @@
...
+}
+
+/**
+ * Arguments for generating an `SingleKeyAccount` with any supported private key
+ * that is not an `Ed25519PrivateKey`.
+ * The `legacy` argument defaults to false and cannot be explicitly set to true.
+ */
+export interface GenerateSingleKeyAccountArgs {
+  scheme: Exclude<SigningSchemeInput, SigningSchemeInput.Ed25519>;
+  legacy?: false;
```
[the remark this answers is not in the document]
hardsetting: The ? is so that both of the following are accepted:
```typescript
Account.generate({ SigninSchemeInput.Secp256k1 }): SingleKeyAccount;
Account.generate({ SigninSchemeInput.Secp256k1, legacy: false }): SingleKeyAccount;
```
but this isn't
```typescript
Account.generate({ SigninSchemeInput.Secp256k1, legacy: true }): never;
```

src/core/account/Account.ts
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * Note: Generating an account instance does not create the account on-chain.
+ */
+export abstract class Account {
+  /**
+   * Private key associated with the account.
+   * Note: this will be removed in the next major release,
+   *  as not all accounts have a private key.
+   */
+  abstract readonly privateKey: PrivateKey;
```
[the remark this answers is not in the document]
hardsetting: I don't think it's used anywhere in the codebase, so should be easy to remove, but it would be a breaking change.

src/core/account/Account.ts
```diff
@@ -0,0 +1,245 @@
...
+  scheme?: SigningSchemeInput.Ed25519;
+  legacy?: true;
+}
+
+/**
+ * Arguments for generating an `SingleKeyAccount` with ah underlying `Ed25519PrivateKey`.
+ * The `legacy` argument needs to be explicitly set to false,
+ * otherwise an `Ed25519Account` will be returned instead.
+ */
+export interface GenerateEd25519SingleKeyAccountArgs {
```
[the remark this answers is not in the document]
hardsetting: This is needed if we want exact types to be returned at compile time.
If that's not something desireable to people, we can remove.

src/core/crypto/multiKey.ts
```diff
@@ -17,13 +35,15 @@ export class MultiKey extends PublicKey {
    */
   public readonly signaturesRequired: number;
 
-  constructor(args: { publicKeys: PublicKey[]; signaturesRequired: number }) {
+  // region Constructors
+
+  constructor(args: { publicKeys: Array<PublicKey | AnyPublicKey>; signaturesRequired: number }) {
```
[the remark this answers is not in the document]
hardsetting: Good call, now that I re-merged back all public keys that union is not needed anymore

src/core/account/Account.ts
```diff
@@ -0,0 +1,245 @@
...
+  path: string;
+  mnemonic: string;
+}
+
+/**
+ * Interface for a generic Aptos account.
+ *
+ * The interface is defined as abstract class to provide a single entrypoint for account generation,
+ * either through `Account.generate()` or `Account.fromDerivationPath`.
+ * Despite this being an abstract class, it should be treated as an interface and enforced using
```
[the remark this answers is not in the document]
hardsetting: Dependency cycles 😢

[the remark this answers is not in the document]
heliuchuan: Sad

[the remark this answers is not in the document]
hardsetting: `Account` needs full imports for all accounts implementations.
If you do `class MyAccount extends Account` you need full import from `Account`

src/core/account/Account.ts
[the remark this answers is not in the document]
heliuchuan: Ok fair enough leave as is and i'll remove.

src/core/account/Account.ts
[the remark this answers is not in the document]
heliuchuan: i see i see.

src/core/account/Account.ts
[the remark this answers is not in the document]
0xmaayan: > This is needed if we want exact types to be returned at compile time.
Lets keep it

src/core/account/Account.ts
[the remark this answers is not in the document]
hardsetting: so the options are:
1) move the generalized constructors outside of the `Account` class and into a separate file
2) don't use `extends` (which is acceptable since there's no particular benefit in doing that)
3) putting everything in the same file, which is not sustainable

To me option 1 would be the best one, but I got pushbacks since "users won't be able to easily find how to create an account".

Another option would be to do 1, and then use mixins.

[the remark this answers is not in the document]
hardsetting: if we want to change this up in the future, it should be a quick fix

PR #300, opened 2024-02-21
Fix typos

rex4539:


rex4539:
@gregnazario Could you please take a look?

docs/@aptos-labs/ts-sdk-0.0.8/classes/TransactionWorker.html
```diff
@@ -1,6 +1,6 @@
...
 </div></section><section class="tsd-index-section"><h3 class="tsd-index-heading">Properties</h3><div class="tsd-index-list"><a href="TransactionWorker.html#account" class="tsd-index-link"><svg class="tsd-kind-icon" viewBox="0 0 24 24"><use href="#icon-1024"></use></svg><span>account</span></a>
-<a href="TransactionWorker.html#accountSequnceNumber" class="tsd-index-link"><svg class="tsd-kind-icon" viewBox="0 0 24 24"><use href="#icon-1024"></use></svg><span>account<wbr/>Sequnce<wbr/>Number</span></a>
+<a href="TransactionWorker.html#accountSequenceNumber" class="tsd-index-link"><svg class="tsd-kind-icon" viewBox="0 0 24 24"><use href="#icon-1024"></use></svg><span>account<wbr/>Sequence<wbr/>Number</span></a>
```
gregnazario, line 3: Please don't change this in the documentation.

Unfortunately, looks like the name is misspelled on the actual TransactionWorker, and that will need to be updated in the next release.  At that point the docs will be regenerated correctly.

My suggestion here would be to leave all the existing docs alone, fix any other spelling mistakes, and then with the next release, we'll update the autogenerated documentation for the next release.

gregnazario (commented):
Looks good, other than we shouldn't change the generated documentation.  cc @0xmaayan

tests/e2e/api/digitalAsset.test.ts
```diff
@@ -128,31 +128,31 @@ describe("DigitalAsset", () => {
   });
 
   test("it freezes transfer ability", async () => {
-    const transaction = await aptos.freezeDigitalAssetTransaferTransaction({
+    const transaction = await aptos.freezeDigitalAssetTransferTransaction({
```
0xmaayan, line 131: need to also change it on the api file https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/api/digitalAsset.ts#L328-L345

And should also marked as a breaking change on CHANGELOG

0xmaayan (commented):
Overall looks good, thanks for working on it!

There is a function API rename that needs to be renamed on the API file as well and marked as a breaking change on CHANGELOG.
Also, please revert docs changes as it is auto generated.

gregnazario:
> Thanks for the review!
> 
> Phew, it's a lot of hassle to revert the docs changes.
> 
> I better close it :)

Oh, it should have been as simple as `git checkout origin/main docs`. I think

PR #301, opened 2024-02-22
Add support for validator transactions

heliuchuan:
### Description
In release 1.10 a new transaction type 'validator_transaction' was added.  This adds its corresponding response type.  Fixed a few type bugs as well.

Example 
```
{
   "version":"3",
   "hash":"0xc2f1a301b4af32c817f624f7b6edb778c5ec522e1e122b1b0d74472a10809cef",
   "state_change_hash":"0xf8699a3b0194c10eb0242fdeb9361deee8a3c242a21222e74df857c3c0ac1961",
   "event_root_hash":"0x0db7799c6ee25bc8ca2539ba4fb04f1c701c9bc929c97c54e7a12d584ade84a1",
   "state_checkpoint_hash":null,
   "gas_used":"0",
   "success":true,
   "vm_status":"Executed successfully",
   "accumulator_root_hash":"0xec3a5fa8a61310432a89b23549cea590a44e79afe9d232d241efb23a6f18592f",
   "changes":[
      {
         "address":"0x1",
         "state_key_hash":"0x50578ada6e87f43cf8693b904e291f9a4b45e3dd8ba76fd7a5b6cfe2b04db116",
         "data":[
            "Object"
         ],
         "type":"write_resource"
      },
      {
         "address":"0x1",
         "state_key_hash":"0x5c4285a5bcbfb079714ca24ba9c766305b3d1ae093da3bce6c5c32aaa7f6cac0",
         "data":[
            "Object"
         ],
         "type":"write_resource"
      }
   ],
   "events":[
      {
         "guid":[
            "Object"
         ],
         "sequence_number":"0",
         "type":"0x1::jwks::ObservedJWKsUpdated",
         "data":[
            "Object"
         ]
      }
   ],
   "timestamp":"1708575651139233",
   "type":"validator_transaction"
}
```

### Test Plan
Added a static test against a sample JSON response to make sure the response satisfies the type.  

Perhaps later make this a runtime check against 'live' data but then we'd have to write/generate our own runtime type check.

Note the user_transaction test doesn't pass because of the single sender API issues.


src/types/index.ts
```diff
@@ -340,7 +348,7 @@ export type UserTransactionResponse = {
   hash: string;
   state_change_hash: string;
   event_root_hash: string;
-  state_checkpoint_hash?: string;
+  state_checkpoint_hash: string | null;
```
gregnazario, line 351: The `?` is specifically for undefined?  Is there a reason to switch to `string | null`?

src/types/index.ts
```diff
@@ -403,7 +411,7 @@ export type BlockMetadataTransactionResponse = {
...
```
gregnazario, line 414: Similarly

tests/e2e/api/transactionResponse.test.ts
```diff
@@ -0,0 +1,530 @@
...
+    } satisfies BlockMetadataTransactionResponse;
+  });
+
+  //   Comment out this test until the API is fixed for single sender txns
+  //   test("it tests a sample user_transaction JSON response satisfies UserTransactionResponse", async () => {
+  //    const txn = {
+  //       "version":"6224",
+  //       "hash":"0xa17bffc240c11f8e9bc2ba8b4b1a45cae44846303c96bf62ca9fafa0fc2d07a8",
+  //       "state_change_hash":"0x2d5c8a6f05e820e81b297b3cde83bf55f3e5ea86f6edc2dc11744e851beea5cb",
+  //       "event_root_hash":"0x1b7628f4bbd261df56e25b44bdce7665193e5014c8221481f631927e1c7974ef",
```
gregnazario, line 252: ? Why add a commented out test?

What does this mean "API is fixed for single sender txns"?

tests/e2e/api/transactionResponse.test.ts
```diff
@@ -0,0 +1,530 @@
...
+            },
+          },
+        },
+      ],
+      timestamp: "1708575651139233",
+      type: TransactionResponseType.Validator,
+    } satisfies ValidatorTransactionResponse;
+  });
+
+  test("it tests a sample block_metadata JSON response satisfies ValidatorTransactionResponse", async () => {
```
gregnazario, line 130: This name is misleading, since it's checking for block metadata to look like block metadata.

tests/e2e/api/transactionResponse.test.ts
```diff
@@ -0,0 +1,530 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
gregnazario, line 1: Shouldn't have to add no unused vars to a new file.

tests/e2e/api/transactionResponse.test.ts
```diff
@@ -0,0 +1,530 @@
...
+/* eslint-disable max-len */
```
gregnazario, line 2: Can you put this below the copyright lines?

gregnazario (commented):
Mostly comments around the commented out test, as well as why is the `?` changed to `| null`?

src/types/index.ts
heliuchuan: yes.  the API returns null, not undefined.  So the attribute exists (is defined), but is set to null.

If you didn't make this change the test would fail (compilation)

tests/e2e/api/transactionResponse.test.ts
heliuchuan: It fails because the API wasnt updated to support single sender transactions when @davidiw  implemented it.  This was the issue that bowen encountered.

0xmaayan: not following.. why do we add a non-working commented out test? can't we just not add it..?

0xmaayan (commented):
Overall lgtm. A comment around a new not working commented out test - a bit confused around it

Also, please update changelog

tests/e2e/api/transactionResponse.test.ts
gregnazario: Yeah we either add it, how it looks today, or we delete the commented out test.

heliuchuan: deleted

0xmaayan (approved):

PR #304, opened 2024-02-23
Add API to get module event 

0xaptosj:
### Description

Add new event API to get module event from indexer.

### Test Plan

Uint test.

### Related Links

https://github.com/aptos-labs/aptos-developers/discussions/6

### DIFF
--- a/src/api/event.ts
+++ b/src/api/event.ts
@@ -1,7 +1,12 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-import { getAccountEventsByCreationNumber, getAccountEventsByEventType, getEvents } from "../internal/event";
+import {
+  getAccountEventsByCreationNumber,
+  getAccountEventsByEventType,
+  getModuleEventsByEventType,
+  getEvents,
+} from "../internal/event";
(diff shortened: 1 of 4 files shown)


alinush (approved):
Blind stamp

src/api/event.ts
```diff
@@ -15,6 +20,27 @@ import { waitForIndexerOnVersion } from "./utils";
...
+  /**
+   * Get module events by event type
+   *
+   * @param args.eventType - The event type
+   * @param args.minimumLedgerVersion Optional ledger version to sync up to, before querying
+   *
+   * @returns Promise<GetEventsResponse>
+   */
+  async getModuleEventsByEventType(args: {
+    eventType: MoveStructId;
```
bowenyang007, line 32: Can you add a check here that if the event type is too long (wrapped for example) we should send a warning and possibly error. We cap indexed_event at length 300.

0xmaayan (approved):
LGTM. Please update changelog

src/api/event.ts
0xaptosj: Added to [internal function](https://github.com/aptos-labs/aptos-ts-sdk/pull/304/files#diff-e2ce900e3281d047225ea1fd842abcb5fa55614f33438d27e1b1b6e342127ccbR101)

PR #307, opened 2024-02-25
[bugfix] Fix TSC when compiling using SDK

gregnazario:
### Description
For some reason, TSC does not like this Network.DEVNET constant, even though the message specifically says, enum literals are okay.

This allows `tsc -b` to work properly on its own, and I've tested with the examples and other locations.

This has no actual behavior change (probably should be released with a patch version).  The DEFAULT_NETWORK constant is probably not very useful, given that most users would change it to the network they expect.

I discovered this trying to build a CLI with a prebuilt example, but tsc -b would always fail with:
```
node_modules/.pnpm/@aptos-labs+ts-sdk@1.8.0/node_modules/@aptos-labs/ts-sdk/dist/esm/utils/const.d.mts:9:33 - error TS2304: Cannot find name 'Network'.

9 declare const DEFAULT_NETWORK = Network.DEVNET;
                                  ~~~~~~~

node_modules/.pnpm/@aptos-labs+ts-sdk@1.8.0/node_modules/@aptos-labs/ts-sdk/dist/esm/utils/const.d.mts:9:33 - error TS1254: A 'const' initializer in an ambient context must be a string or numeric literal or literal enum reference.

9 declare const DEFAULT_NETWORK = Network.DEVNET;

```

### Test Plan
I tested this locally against the examples folder and ran `tsc -b`.

### DIFF
--- a/src/api/aptosConfig.ts
+++ b/src/api/aptosConfig.ts
@@ -4,7 +4,7 @@
 import aptosClient from "@aptos-labs/aptos-client";
 import { AptosSettings, ClientConfig, Client } from "../types";
 import { NetworkToNodeAPI, NetworkToFaucetAPI, NetworkToIndexerAPI, Network } from "../utils/apiEndpoints";
-import { AptosApiType, DEFAULT_NETWORK } from "../utils/const";
+import { AptosApiType } from "../utils/const";
 
 /**
  * This class holds the config information for the SDK client instance.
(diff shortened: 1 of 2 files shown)


0xaptosj (approved):

0xaptosj:
when do you need to run `tsc`? I always run `pnpm build` which runs `tsup`

gregnazario:
> when do you need to run `tsc`? I always run `pnpm build` which runs `tsup`

This is for third party packages that use our SDK.  We can't expect them all to use tsup

0xmaayan (approved):

PR #308, opened 2024-02-26
Use indexer API via API Gateway

banool:
### Description
This is part of the migration to API Gateway. More context here: https://www.notion.so/aptoslabs/Migration-1-Legacy-URL-to-API-Gateway-migration-b007742490684b9bab3debd674504a94.

I found the initial files to change like this:
```
rg -l 'aptoslabs.com'
```

### Test Plan
CI for now.


0xmaayan (approved):
Please update changelog

PR #316, opened 2024-02-29
Remove decimals from getTokenData

0xaptosj:
### Description

This field doesn't exist.


0xaptosj:
Indexer API will expose this field

PR #328, opened 2024-03-12
[types] Ensure string is checked before other types

gregnazario:
### Description
In Javascript, the types are erased, so inputs may not have the types associated.  However, string can still be checked, so all inputs should first be checked for if it's a string, then fall back on other options.

This was often manifesting as:
```
"TypeError: Cannot read properties of undefined (reading 'startsWith')",
    "    at n.fromString (file:///var/task/node_modules/@aptos-labs/ts-sdk/dist/esm/chunk-BZMO6Q6L.mjs:1:422)",
    "    at n.fromHexInput (file:///var/task/node_modules/@aptos-labs/ts-sdk/dist/esm/chunk-BZMO6Q6L.mjs:1:879)",
    "    at new r (file:///var/task/node_modules/@aptos-labs/ts-sdk/dist/esm/chunk-NUBFOYQZ.mjs:1:1038)",
    "    at Runtime.handler (file:///var/task/index.mjs:42:57)",
    "    at Runtime.handleOnceNonStreaming (file:///var/runtime/index.mjs:1173:29)"
```

### Test Plan
Tested by replacing one of the AccountAddress shortcuts with a toString()

### DIFF
--- a/src/core/hex.ts
+++ b/src/core/hex.ts
@@ -131,8 +131,11 @@ export class Hex {
    * @returns Hex
    */
   static fromHexInput(hexInput: HexInput): Hex {
-    if (hexInput instanceof Uint8Array) return new Hex(hexInput);
-    return Hex.fromString(hexInput);
+    // Note we have to be more defensive here, since Javascript doesn't have the concept of types
+    if (typeof hexInput === "string") {
+      return Hex.fromString(hexInput);
+    }
+    return new Hex(hexInput);
   }
 
   // ===
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 
 # Unreleased
 
+- [`Fix`] Ensure that hex input strings are loaded as strings in Javascript
+
 # 1.10.0 (2024-03-11)
 
 - [`Deprecate`] IIFE build support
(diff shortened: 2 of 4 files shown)


src/core/hex.ts
```diff
@@ -131,8 +131,11 @@ export class Hex {
    * @returns Hex
    */
   static fromHexInput(hexInput: HexInput): Hex {
-    if (hexInput instanceof Uint8Array) return new Hex(hexInput);
-    return Hex.fromString(hexInput);
+    // Note we have to be more defensive here, since Javascript doesn't have the concept of types
+    if (typeof hexInput === "string") {
```
0xmaayan, line 135: not completely sure this is a fix. HexInput is of type string | Uint8Array. Hex constructor accepts Uint8Array. if hexInput is of type string, it would never be instanceof Uint8Array and therefore would get to Hex.fromString. 

The thing is that javascript doesnt care for types, so technically you could pass in whatever you want (tho I think you would get a typing error from the sdk). So if you pass in anything that is not string | Uint8Array you.... screwed.

0xmaayan (commented):
I am confused about this change and if it is even a fix. Tests seem to pass if passing in a string/non-string from a js file before this change

gregnazario:
> I am confused about this change and if it is even a fix. Tests seem to pass if passing in a string/non-string from a js file before this change. Also, the description error complains on passing in an `undefined`

Yeah, this was my mistake.

The real question is if we want to bother protecting against undefined everywhere, or just leave it to the user to not mess up.

PR #332, opened 2024-03-21
Fixed keyless simulation

hardsetting:
### Description

Handling automatic authenticator generation for keyless transaction simulation

### DIFF
--- a/src/transactions/transactionBuilder/transactionBuilder.ts
+++ b/src/transactions/transactionBuilder/transactionBuilder.ts
@@ -10,15 +10,11 @@ import { AptosConfig } from "../../api/aptosConfig";
 import { AccountAddress, AccountAddressInput, Hex, PublicKey } from "../../core";
 import { AnyPublicKey, AnySignature } from "../../core/crypto";
 import { Ed25519PublicKey, Ed25519Signature } from "../../core/crypto/ed25519";
-import { Secp256k1PublicKey, Secp256k1Signature } from "../../core/crypto/secp256k1";
 import { getInfo } from "../../internal/account";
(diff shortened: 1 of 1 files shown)


src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -387,12 +383,7 @@ export function generateSignedTransactionForSimulation(args: InputSimulateTransa
 export function getAuthenticatorForSimulation(publicKey: PublicKey) {
   // TODO add support for AnyMultiKey
   if (publicKey instanceof AnyPublicKey) {
-    if (publicKey.publicKey instanceof Ed25519PublicKey) {
-      return new AccountAuthenticatorSingleKey(publicKey, new AnySignature(new Ed25519Signature(new Uint8Array(64))));
-    }
-    if (publicKey.publicKey instanceof Secp256k1PublicKey) {
-      return new AccountAuthenticatorSingleKey(publicKey, new AnySignature(new Secp256k1Signature(new Uint8Array(64))));
-    }
+    return new AccountAuthenticatorSingleKey(publicKey, new AnySignature(new Ed25519Signature(new Uint8Array(64))));
```
gregnazario, line 386: Is this because it's a hack around Secp256k1 and the signature not being checked properly?

Perhaps we can just leave instance of Secp, and then default to Ed25519

hardsetting: Technically not a hack, in rust you can verify any signature against a publicKey.
I think the code would be much cleaner this way, maybe I can add a comment?

heliuchuan (approved):

PR #335, opened 2024-03-22
Updates queries for ANS to ignore expired names

kaw2k:
### Description
**Update:**
This will be a two step refactor. This PR just fixes the imidiate issues. A future PR will address the points listed bellow. 

**Original context**
This PR addresses ANS queries that return values for expired or inactive names when they should not. 

Questions to answer: 
- The following functions directly hit the `router` and not the indexer. As a result, they will return values even if the name is expired. Do we want to shift these to using the indexer and return `null` for expired names? 
  - `getOwnerAddress` 
  - `getExpiration` 
  - `getTargetAddress`
  - `getPrimaryName` (Need to investigate this one.)
- The following APIs use the indexer and ensure that expiration date is valid, should be fine, but good to double check.
  - `getAccountNames`
  - `getAccountDomains`
  - `getAccountSubdomains`
  - `getDomainSubdomains`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 - Add function to create token object address locally
 - Add signers to entry function ABI for future signature count checking
 - [`Breaking`] Add type-safe view functions with ABI support
+- [`Fix`] ANS `getName` and `getDomainSubdomains` now appropriately ignores invalid and expired names
 
 # 1.10.0 (2024-03-11)
 
--- a/src/internal/ans.ts
+++ b/src/internal/ans.ts
@@ -340,6 +340,7 @@ export async function getName(args: {
   const where: CurrentAptosNamesBoolExp = {
     domain: { _eq: domainName },
     subdomain: { _eq: subdomainName },
+    is_active: { _eq: true },
   };
 
   const data = await queryIndexer<GetNamesQuery>({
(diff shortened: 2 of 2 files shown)


0xmaayan:
> ### Description
> This PR addresses ANS queries that return values for expired or inactive names when they should not.
> 
> Questions to answer:
> 
> * The following functions directly hit the `router` and not the indexer. As a result, they will return values even if the name is expired. Do we want to shift these to using the indexer and return `null` for expired names?
>   
>   * `getOwnerAddress`
>   * `getExpiration`
>   * `getTargetAddress`
>   * `getPrimaryName` (Need to investigate this one.)

What are the consideration in using router directly and not indexer?

> * The following APIs use the indexer and ensure that expiration date is valid, should be fine, but good to double check.
>   
>   * `getAccountNames`
>   * `getAccountDomains`
>   * `getAccountSubdomains`
>   * `getDomainSubdomains`

sounds good

kaw2k:
Looking back on the first question, I think switching to the indexer is a given at this point. I think the thing to answer is for names that are expired and past their grace period (30 days when they can be reclaimed by their owner and no one else), should we return null for these calls.

0xmaayan:
> Looking back on the first question, I think switching to the indexer is a given at this point. I think the thing to answer is for names that are expired and past their grace period (30 days when they can be reclaimed by their owner and no one else), should we return null for these calls.

I think null makes sense. Dont think they should be included on one's dashboard if they are expired unless we introduce "expired" view

0xmaayan (approved):
LGTM! Please update CHANGELOG

0xaptosj (approved):

Issue #258, opened 2024-01-18
[Feature Request] Allow setting per-backend client config


banool:
When building `Aptos` you can provide a `ClientConfig` that is used for the underlying client when building requests.

It should be possible to provide a per-backend `ClientConfig` as well. So the final client config is the base client config with the per-backend config applied on top.

This way you can set different headers per backend, among other things.

0xmaayan:
@banool Thanks! Does the full `ClientConfig` type make sense per-backend config?
```
export type ClientConfig = {
  AUTH_TOKEN?: string;
  API_KEY?: string;
  HEADERS?: Record<string, string | number | boolean>;
  WITH_CREDENTIALS?: boolean;
};
```

banool:
I think having both of these doesn't make sense, they're for different uses right?
```
AUTH_TOKEN?: string;
API_KEY?: string;
```

Maybe those need to split up into the per-backend configs or something?

0xmaayan:
Yea. `AUTH_TOKEN` is for faucet and `API_KEY` is part of the api gateway so node and indexer at the moment

PR #336, opened 2024-03-23
Add support to allow setting per-backend configuration

0xmaayan:
### Description
This PR adds the ability to set client configuration per back-end server (resolves https://github.com/aptos-labs/aptos-ts-sdk/issues/258 )

- Allow setting per-backend client config
- `AUTH_TOKEN` moved to be under `faucetConfig` as it is faucet config only
- Handle `Unauthorized` server error
- Clean up and simplify code

### Test Plan
- tests are passing
- added new test files

### DIFF
--- a/src/client/core.ts
+++ b/src/client/core.ts
@@ -32,13 +32,10 @@ export async function request<Req, Res>(options: ClientRequest<Req>, client: Cli
     "content-type": contentType ?? MimeType.JSON,
   };
 
-  // TODO - auth token is being used only for faucet, it breaks full node requests.
-  // Find a more sophisticated way than that but without the need to add the
-  // auth_token on every `aptos.fundAccount()` call
-  if (overrides?.AUTH_TOKEN && url.includes("faucet")) {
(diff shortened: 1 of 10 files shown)


src/client/core.ts
```diff
@@ -81,11 +78,16 @@ export async function aptosRequest<Req extends {}, Res extends {}>(
     url: fullUrl,
   };
 
+  // Handle case for `Unauthorized` error (i.e API_KEY error)
+  if (result.status === 401) {
```
gregnazario, line 82: May want to include 403 in here as well later.

gregnazario (approved):

src/client/core.ts
0xmaayan: interesting... what is 403 status for?

gregnazario: It's forbidden, rather than 401 unauthorized

It's honestly used interchangeably sometimes.  But, if 401 is the only one that's thrown right now we can stick with it

0xmaayan: ohh yea, server throws 401

PR #341, opened 2024-03-26
Bump version to 1.11.0

0xmaayan:


gregnazario (approved):

hariria (approved):
I wonder if we need to commit the html files? Can we just build and deploy the site without committing them?

Issue #259, opened 2024-01-18
[Feature Request] Explore having `SimpleTransaction` and `MultiAgentTransaction` types as serializable/deserializable Objects


0xmaayan:
# 🚀 Feature Request Description

Currently, when generating a transaction in the SDK with `aptos.transaction.build.simple/multiAgent` the return type is `SimpleTransaction` or `MultiAgentTransaction` - respectively. Since these are simple TS types, we cant BCS serialize or deserialize the whole object but only the `rawTransaction` prop. 

## Motivation

When the transaction submission flow includes different servers that do different tasks, this starts to get a bit complex/annoying in serializing and sending the whole object data to a different server.

Consider a frontend server that generates a transaction, signs as a fee payer, serializes it and sends it to a backend server to submit the transaction.
This flow requires the frontend server to send the fee payer address separately as when the backend server wants to submit the transaction, it needs to include the fee payer address. 

Frontend
```
const transaction = await aptos.transaction.build.simple({
    sender: alice.accountAddress,
    withFeePayer: true,
    data: {
      function: "0x1::aptos_account::transfer",
      functionArguments: [bob.accountAddress,"1_000_000],
    },
  });

// Alice signs
const senderAuth = aptos.transaction.sign({ signer: alice, transaction });

// Sponsor signs
const sponsorAuth = aptos.transaction.signAsFeePayer({ signer: sponsor, transaction });

// Serialize everything
const rawTransactionBytes = transaction.rawTransaction.bcsToBytes();
const feePayerAddressByes = sponsor.accountAddress.bcsToBytes();
const serializedSenderAuthenticatorBytes = senderAuth.bcsToBytes();
const serializedFeePayerAuthenticatorBytes = sponsorAuth.bcsToBytes();

const serializedData = {
    rawTransactionBytes,
    feePayerAddressByes,
    serializedSenderAuthenticatorBytes,
    serializedFeePayerAuthenticatorBytes,
};

// Send to server to submit
await sendToOtherServer(serializedData);
```

**Describe the solution you would like**
Have the `aptos.transaction.build.simple/multiAgent` return a structured object that each field can be serialized/deserialized and that can easily sent to another server 
```
class SimpleTransaction extends Serializable {...}
```

PR #342, opened 2024-03-28
Introduce Serializable transaction modules

0xmaayan:
### Description
solves https://github.com/aptos-labs/aptos-ts-sdk/issues/259 

Following requests to simplify the communication between servers when sending transactions, this PR introduces new modules for generated transactions - `SimpleTransaction` and `MultiAgentTransaction`.
With these changes, a server can easily serialize a transaction and send to another server for easy deserialization.

**Before:**
The generated `transaction` is of a simple type and can't be bcs serialized. Therefore we need to serialize the `rawTransaction` and send it to the server along with other needed transaction details.
```
const rawTransactionBytes = transaction.rawTransaction.bcsToBytes();
const feePayerAddressByes = sponsor.accountAddress.bcsToBytes();
const serializedSenderAuthenticatorBytes = senderAuth.bcsToBytes();
const serializedFeePayerAuthenticatorBytes = sponsorAuth.bcsToBytes();

const serializedData = {
  rawTransactionBytes,
  feePayerAddressByes,
  serializedSenderAuthenticatorBytes,
  serializedFeePayerAuthenticatorBytes,
};

await sendToOtherServer(serializedData);
```

**After:**
The generated transaction is of type Serializable and can be bcs serialized.
```
await sendToOtherServer(transaction.bcsToBytes(), senderAuth.bcsToBytes(), sponsorAuth.bcsToBytes());
```

### Test Plan
- tests are passing
- examples have been updated

### DIFF
--- a/src/transactions/instances/simpleTransaction.ts
+++ b/src/transactions/instances/simpleTransaction.ts
@@ -1,0 +1,60 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+/* eslint-disable @typescript-eslint/naming-convention */
+
+import { Deserializer } from "../../bcs/deserializer";
+import { Serializable, Serializer } from "../../bcs/serializer";
+import { AccountAddress } from "../../core";
+import { RawTransaction } from "./rawTransaction";
+
+/**
+ * Representation of a SimpleTransaction that can serialized and deserialized
+ */
+export class SimpleTransaction extends Serializable {
+  public rawTransaction: RawTransaction;
+
+  public feePayerAddress?: AccountAddress | undefined;
+
+  // We dont really need it, we add it for type checkings we do
+  // throughout the SDK
+  public readonly secondarySignerAddresses: undefined;
+
+  /**
+   * SimpleTransaction represents a simple transaction type of a single signer that
+   * can be submitted to Aptos chain for execution.
+   *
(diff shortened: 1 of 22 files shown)


src/transactions/instances/simpleTransaction.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  serialize(serializer: Serializer): void {
+    this.rawTransaction.serialize(serializer);
+
+    if (this.feePayerAddress === undefined) {
+      serializer.serializeBool(false);
+    } else {
+      serializer.serializeBool(true);
+      this.feePayerAddress.serialize(serializer);
+    }
```
gregnazario, line 47: Interesting, using a bool instead of `null` or `[]`

Though... actually I think that 

`bool,address` serializes the same in BCS by accident as the "optional" address
`[address]` and `[]`

Up to you here, because you could just serialize a vector of 0 for none, and a vector of 1 for some.

gregnazario (approved):

src/transactions/instances/simpleTransaction.ts
0xmaayan: yea I think you are right, we use `bool` [here](https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/transactions/instances/transactionPayload.ts#L356) so thought just going with it

Issue #309, opened 2024-02-26
[Bug] `verifySignature` errors out when passing in a `string` type


0xmaayan:
# 🐛 Bug Description

Public Key [verifySignature](https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/core/crypto/asymmetricCrypto.ts#L17) accepts a `HexInput` type which is of `string | Uint8Array` type. 

But when passing it the output of a wallet [signMessage](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L446) function it errors our with

```
E:\Anto\Aptos\Aptos SDK\aptos-sdk-v2\node_modules\@aptos-labs\ts-sdk\src\bcs\serializer.ts:331
  const valueBigInt = BigInt(value);
                      ^
F [Error]: Hex string must be an even number of hex characters.
    at Function.fromString (E:\Anto\Aptos\Aptos SDK\aptos-sdk-v2\node_modules\@aptos-labs\ts-sdk\src\bcs\serializer.ts:331:23)
    at Function.fromHexInput (E:\Anto\Aptos\Aptos SDK\aptos-sdk-v2\node_modules\@aptos-labs\ts-sdk\src\core\accountAddress.ts:19:3)
    at verifySignature (E:\Anto\Aptos\Aptos SDK\aptos-sdk-v2\node_modules\@aptos-labs\ts-sdk\src\core\crypto\multiEd25519.ts:206:47)
```

## How to reproduce

**Code snippet to reproduce**
```typescript
const main = async () => {
  const inputString =
    "APTOS\nmessage: Hello from Aptos Wallet Adapter\nnonce: random_string";
  const signature = new Ed25519Signature(
    "70b6da5e8d8a22a45f582cec5b724cce201545bef614216db645237d9ff3391e201b0807a1dd7d240f3623173d6f681d46f153043552ef55fc06853530836704"
  );

  let publicKey = new Ed25519PublicKey(
    "0xa9d8dd9ac2240a5fd7975ee92f959476ea69ed0a369ed46011a6a2156a383bf8"
  );
  console.log(
    "Verification:",
    publicKey.verifySignature({ message: inputString, signature })
  );
};
main();
```

**Stack trace / error message**
```
// Paste the output here
```

## System information

**System details:**
- Typescript SDK Version
- Platform (e.g. Node, browser, etc.)

0xmaayan:
[...]

PR #344, opened 2024-03-29
Verify correct message format when sign a message and/or verify signature

0xmaayan:
### Description
resolves https://github.com/aptos-labs/aptos-ts-sdk/issues/309

To verify or sign a message, we should pass the message as a Hex string or Uint8Array type. We can't enforce Hex string type on TS level (since how do you know if it a Hex string or a string? i.e `123abc`). Therefore we check if the message is of type `string`, if it does we check if it is a valid `Hex string` and if it is not we convert it into a `Buffer` type.

In addition, change `Hex.fromString` to `Hex.fromHexString` for clarity

### Test Plan
added tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 # Unreleased
 
 - [`Breaking`] Change `getOwnerAddress` and `getTargetAddress` return type to `AccountAddress`
+- Add `message` input type verification on `sign` and `verifySignature` functions and convert it into a correct type if needed
+- [`Breaking`] Change `fromString` to `fromHexString` on `Hex` class
 
 # 1.11.0 (2024-03-26)
 
--- a/src/core/crypto/ed25519.ts
+++ b/src/core/crypto/ed25519.ts
@@ -11,6 +11,7 @@ import { CKDPriv, deriveKey, HARDENED_OFFSET, isValidHardenedPath, mnemonicToSee
 import { PrivateKey } from "./privateKey";
 import { AccountPublicKey, VerifySignatureArgs } from "./publicKey";
 import { Signature } from "./signature";
+import { convertMessage } from "./utils";
 
 /**
  * Represents the public key of an Ed25519 key pair.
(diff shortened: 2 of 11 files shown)


src/core/crypto/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+    // If message is not a valid Hex string, convert it into a Buffer
+    if (!isValid.valid) {
+      return Buffer.from(message, "utf8");
+    }
+    // If message is a valid Hex string, return it
+    return message;
+  }
+  // message is a Uint8Array
+  return message;
+};
```
gregnazario, line 24: Trying to come up with a better name than `convertMessage`

So, this lets you take in:

1. Hex string
2. A UTF-8 encoded string (as long as it isn't a hex string)
3. a Uint8array

Let's put that comment there, and maybe change it to... convertSigningMessage

src/core/crypto/ed25519.ts
```diff
@@ -213,11 +214,12 @@ export class Ed25519PrivateKey extends Serializable implements PrivateKey {
   /**
    * Sign the given message with the private key.
    *
-   * @param message in HexInput format
+   * @param message a message as a Hex string or Uint8Array
```
gregnazario, line 217: The convert message lets you accept UTF-8 strings that aren't hex though now?

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 # Unreleased
 
 - [`Breaking`] Change `getOwnerAddress` and `getTargetAddress` return type to `AccountAddress`
+- Add `message` input type verification on `sign` and `verifySignature` functions and convert it into a correct type if needed
+- [`Breaking`] Change `fromString` to `fromHexString` on `Hex` class
```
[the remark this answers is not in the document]
0xmaayan: Some users including me were confused by it as `fromString` seems like it converts any string to Hex, but it actually converts Hex string to Hex

src/core/crypto/utils.ts
0xmaayan: yea I wasnt sure what name makes sense. convertSigningMessage sounds good

From TS point of view, it lets you pass in `string | Uint8array` where `string` can be any string (regular string, hex string, etc)

gregnazario (approved):
Approved pending fixes

Issue #321, opened 2024-03-06
[Feature Request] Add an API to compute TX hash


0xaptosj:
# 🚀 Feature Request Description

Here's how to compute the hash according to [comments in aptos-core ](https://github.com/aptos-labs/aptos-core/blob/main/api/src/transactions.rs#L186)

```
To create a transaction hash by yourself, do the following:
1. Hash message bytes: "RawTransaction" bytes + BCS bytes of [Transaction](https://aptos-labs.github.io/aptos-core/aptos_types/transaction/enum.Transaction.html).
2. Apply hash algorithm `SHA3-256` to the hash message bytes.
3. Hex-encode the hash bytes with `0x` prefix.
```

## Motivation

**Is your feature request related to a problem? Please describe you use case.**

## Pitch

**Describe the solution you would like**

**Describe alternatives you've considered**

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))

## Additional context

[Someone requested this feature on stack overflow](https://stackoverflow.com/questions/77462495/get-transaction-hash-from-raw-transaction-in-aptos-network).

0xaptosj:
> Is this to do local calculation? Or simulation can help here?

I guess he wants to calculate the hash locally, but i answered on stack overflow whether simulation works

PR #345, opened 2024-03-31
[transaction] Add ability to get transaction hash

gregnazario:
### Description
This adds the ability to get the transaction hash locally.

How it works is that all Rust structs are signed as follows:
1. SHA3_256 Hash `APTOS::` + `StructName` = `prefix`
2. SHA3_256 Hash the above `prefix` + BCS encoded bytes of the struct

Note that, enums are encoded by a u8, followed by the actual bytes associated.  So, in the case of the transaction hash, it must be a `UserTransaction` = 0 for the enum type before the rest of the bytes.

Resolves https://github.com/aptos-labs/aptos-ts-sdk/issues/321

Note this code, only hashes user transactions

### Test Plan
E2E test ensures it's the same as the API

### DIFF
--- a/src/transactions/transactionBuilder/transactionBuilder.ts
+++ b/src/transactions/transactionBuilder/transactionBuilder.ts
@@ -528,6 +528,36 @@ export function generateSignedTransaction(args: InputSubmitTransactionData): Uin
 }
 
 /**
+ * Hashes the set of values with a SHA-3 256 hash
+ * @param input array of UTF-8 strings or Uint8array byte arrays
+ */
+export function hashValues(input: (Uint8Array | string)[]): Uint8Array {
+  const hash = sha3Hash.create();
+  for (const item of input) {
+    hash.update(item);
+  }
+  return hash.digest();
+}
+
+/**
+ * The domain separated prefix for hashing transacitons
+ */
+const TRANSACTION_PREFIX = hashValues(["APTOS::Transaction"]);
+
+/**
+ * Generates a transaction hash for the given transaction payload.  It must already have an authenticator
+ * @param args
+ */
+export function generateTransactionHash(args: InputSubmitTransactionData): string {
+  const signedTransaction = generateSignedTransaction(args);
+
+  // Transaction signature is defined as, the domain separated prefix based on struct (Transaction)
+  // Then followed by the type of the transaction for the enum, UserTransaction is 0
+  // Then followed by BCS encoded bytes of the signed transaction
+  return new Hex(hashValues([TRANSACTION_PREFIX, new Uint8Array([0]), signedTransaction])).toString();
+}
+
+/**
  * Derive the raw transaction type - FeePayerRawTransaction or MultiAgentRawTransaction or RawTransaction
  *
  * @param transaction A aptos transaction type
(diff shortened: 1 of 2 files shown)


src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -527,6 +527,36 @@ export function generateSignedTransaction(args: InputSubmitTransactionData): Uin
...
+}
+
+/**
+ * The domain separated prefix for hashing transacitons
+ */
+const TRANSACTION_PREFIX = hashValues(["APTOS::Transaction"]);
+
+/**
+ * Generates a transaction hash for the given transaction payload.  It must already have an authenticator
+ * @param args
```
0xmaayan, line 549: nit
```suggestion
 * @param args InputSubmitTransactionData
```

src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -527,6 +527,36 @@ export function generateSignedTransaction(args: InputSubmitTransactionData): Uin
...
+ */
+export function generateTransactionHash(args: InputSubmitTransactionData): string {
```
0xmaayan, line 551:
```suggestion
export function generateUserTransactionHash(args: InputSubmitTransactionData): string {
```

tests/e2e/transaction/transactionBuilder.test.ts
```diff
@@ -554,6 +555,102 @@ describe("transaction builder", () => {
       expect(signedTransaction instanceof SignedTransaction).toBeTruthy();
     });
   });
+
+  describe("generateTransactionHash", () => {
```
0xmaayan, line 559:
```suggestion
  describe("generateUserTransactionHash", () => {
```

src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -527,6 +527,36 @@ export function generateSignedTransaction(args: InputSubmitTransactionData): Uin
...
+  return hash.digest();
+}
+
+/**
+ * The domain separated prefix for hashing transacitons
+ */
+const TRANSACTION_PREFIX = hashValues(["APTOS::Transaction"]);
+
+/**
+ * Generates a transaction hash for the given transaction payload.  It must already have an authenticator
```
0xmaayan, line 548:
```suggestion
 * Generates a user transaction hash for the given transaction payload.  It must already have an authenticator
```

tests/e2e/transaction/transactionBuilder.test.ts
```diff
@@ -554,6 +555,102 @@ describe("transaction builder", () => {
...
+
+      // Generate hash
+      const signedTxnInput = {
+        transaction,
+        senderAuthenticator,
+      };
+      const transactionHash = generateTransactionHash(signedTxnInput);
+
+      // Submit transaction
+      const submitted = await aptos.transaction.submit.simple(signedTxnInput);
```
0xmaayan, line 579: do we need to actually submit it to get the transaction hash or is it enough to simulate?

0xmaayan (commented):
LGTM.
Left some comments on naming.
Please update CHANGELOG

0xmaayan (approved):

tests/e2e/transaction/transactionBuilder.test.ts
gregnazario: Simulation would give a different hash, because the signature is zeroed out.

PR #348, opened 2024-04-03
Read local testnet URLs from env vars if available

banool:


tests/e2e/helper.ts
```diff
@@ -0,0 +1,22 @@
...
+ * aren't set, it falls back to the values for Network.LOCAL. Use this whenever you
+ * want a client to use with a local testnet.
+ */
+export function getAptosClient(additionalConfig?: Partial<AptosConfig>): { aptos: Aptos; config: AptosConfig } {
+  // TODO: Do I need to use the dotenv lib for this?
+  const nodeApiUrl = process.env.APTOS_NODE_API_URL || NetworkToNodeAPI.local;
+  const indexerApiUrl = process.env.APTOS_INDEXER_API_URL || NetworkToIndexerAPI.local;
+  const faucetApiUrl = process.env.APTOS_FAUCET_API_URL || NetworkToFaucetAPI.local;
+  const config = new AptosConfig({
+    network: Network.LOCAL,
```
0xmaayan, line 14: l think we should only change the network. Running tests with fullnode points to server A and indexer points to server B would fail.

Also, assume I want to test a complete different network, for example devnet, I should just be able to set the network name and not each endpoint. As it is now, to test against devnet I would need to change fullnode/indexer/faucet specifically - a bit annoying.

tests/e2e/helper.ts
```diff
@@ -0,0 +1,22 @@
+import { Aptos, AptosConfig, Network, NetworkToFaucetAPI, NetworkToIndexerAPI, NetworkToNodeAPI } from "../../src";
+
+/**
+ * This function attempts to read env vars first to configure an Aptos client. If they
+ * aren't set, it falls back to the values for Network.LOCAL. Use this whenever you
+ * want a client to use with a local testnet.
+ */
+export function getAptosClient(additionalConfig?: Partial<AptosConfig>): { aptos: Aptos; config: AptosConfig } {
+  // TODO: Do I need to use the dotenv lib for this?
```
0xmaayan, line 9: might be

tests/e2e/helper.ts
```diff
@@ -0,0 +1,22 @@
+import { Aptos, AptosConfig, Network, NetworkToFaucetAPI, NetworkToIndexerAPI, NetworkToNodeAPI } from "../../src";
+
+/**
+ * This function attempts to read env vars first to configure an Aptos client. If they
+ * aren't set, it falls back to the values for Network.LOCAL. Use this whenever you
+ * want a client to use with a local testnet.
```
0xmaayan, line 6: not completely understand "Use this whenever you want a client to use with a local testnet"? we should use this always and just change the network if we want to run against a different network

0xmaayan (changes requested):

tests/e2e/helper.ts
0xmaayan: I think what we would want to do is
```
const network = process.env.APTOS_NETWORK || Network.LOCAL;
 const nodeApiUrl = process.env.APTOS_NODE_API_URL || NetworkToNodeAPI.local;
  const indexerApiUrl = process.env.APTOS_INDEXER_API_URL || NetworkToIndexerAPI.local;
  const faucetApiUrl = process.env.APTOS_FAUCET_API_URL || NetworkToFaucetAPI.local;
```

and then for you to use it in the other PR, you could set the network to `Network.CUSTOM` along with all other endpoints

tests/e2e/helper.ts
banool: There are places in the E2E tests which test against devnet / testnet. Though I think that's sus and shouldn't be the case.

tests/e2e/helper.ts
banool: The flexibility is intentional, what if I want to test against a different node API but use the normal indexer API? Or in my case, they're using a different port than Network.LOCAL says.

We could add the env var that we have in the examples as well, `APTOS_NETWORK`. I'll push that up so you can see what I mean.

tests/e2e/helper.ts
0xmaayan: I believe we use it to test staking endpoints - which makes sense

0xmaayan (approved):
LGTM

PR #350, opened 2024-04-04
Fix shared state problem with post.test.ts, use localnet

banool:
### Description
Previously this would pass:
```
npx jest tests/e2e/client/post.test.ts
```

This would not:
```
pnpm run e2e-test -t 'faucet it sets correct headers'
```

This is because the faucet test was relying on changes made by the first two tests.

This PR fixes that by using separate clients for each test and fixing the invalid assertions.

I also make the tests run against localnet so there is no need for this try / catch (which was only there bc of some API key rejection stuff that doesn't happen locally) and update all the assertions accordingly.

### Test Plan
See description.

### Related Links
https://aptos-org.slack.com/archives/C05NLAKJM9U/p1712252143249939


0xmaayan (approved):
noice!

PR #355, opened 2024-04-08
Fix flakiness in E2E tests

banool:
## Summary
A few changes in one PR. I'll rebase rather than squash:
- **Increase wait time for local testnet startup**
- **Hit / in api_token for full node requests test, log on error**
  - Log to help debug why it failed.
- **Use decimals function rather than is_registered in function generics test**

## Test Plan
See that CI passes, hopefully with fewer retries than normally required.

### DIFF
--- a/tests/e2e/api/general.test.ts
+++ b/tests/e2e/api/general.test.ts
@@ -174,13 +174,13 @@ describe("general api", () => {
       expect(symbol).toEqual("APT");
 
       const payload2: InputViewFunctionData = {
-        function: "0x1::coin::is_account_registered",
+        function: "0x1::coin::decimals",
         typeArguments: ["0x1::aptos_coin::AptosCoin"],
-        functionArguments: ["0x1"],
+        functionArguments: [],
       };
 
-      const isRegistered = (await aptos.view<[boolean]>({ payload: payload2 }))[0];
-      expect(isRegistered).toEqual(false);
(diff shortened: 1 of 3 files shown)


tests/e2e/api/general.test.ts
```diff
@@ -174,13 +174,13 @@ describe("general api", () => {
...
 
       const payload2: InputViewFunctionData = {
-        function: "0x1::coin::is_account_registered",
+        function: "0x1::coin::decimals",
         typeArguments: ["0x1::aptos_coin::AptosCoin"],
-        functionArguments: ["0x1"],
+        functionArguments: [],
       };
 
-      const isRegistered = (await aptos.view<[boolean]>({ payload: payload2 }))[0];
```
0xmaayan, line 182: have we figured out why it returns true instead of false? might be a higher level problem?

0xmaayan (approved):

tests/e2e/api/general.test.ts
banool: Nope, good point though, we can investigate separately.

banool: I'm guessing some other test transfers coin to that account, but bc the test order is not deterministic the result is different sometimes.

banool:
Bingo, tests passed on the first attempt.

PR #361, opened 2024-04-10
Support generic challenge signing. Introduce ABI deserialization.

0xmaayan:
### Description

Following a community and a foundation request, adding support for generic challenge signing.

With the new generic challenge signing support this PR also introduces deserialization process with the types coming from a remote abi. 
There are some limitation to the deserialization process
1. `vector<u8>` can be a serialization representative to some types such as: public keys (ed25519, secp256k1, etc), string, etc. Since we can't really know anything about the value type other than it is a `vector of u8`, deserializing it is a bit challenging. Therefore, the sdk would try to deserialize it as a ed25519 public key type and if it fails it would return the array of u8 itself.
2. Deserialization of `Generic` or `Option` types are not supported by the SDK as the `deserializer` class in the SDK can only handle a single class, i.e not a class with generics.

### Future work
To not overload this work, will follow with next PRs to
- `remoteAbi.ts` file needs to be refactored
- deserialization flow to accept a local abi

### Test Plan
added a new test for generic challenge
added tests for abi deserialization
tests are passing

### DIFF
--- a/tests/e2e/api/proofChallenge.test.ts
+++ b/tests/e2e/api/proofChallenge.test.ts
@@ -1,15 +1,15 @@
 import { Account, U8, MoveVector, U64, AccountAddress } from "../../../src";
+import { ProofChallenge } from "../../../src/transactions/instances/proofChallenge";
 import { getAptosClient } from "../helper";
 
 const { aptos } = getAptosClient();
 
 describe("proof challenge", () => {
-  test("generic challenge with simple arguments", async () => {
+  test("it creates generic challenge with simple arguments", async () => {
     const fromAccount = Account.generate();
     const newAccount = Account.generate();
 
     await aptos.fundAccount({ accountAddress: fromAccount.accountAddress, amount: 1_000_000_000 });
-    await aptos.fundAccount({ accountAddress: newAccount.accountAddress, amount: 1_000_000_000 });
 
     const accountInfo = await aptos.getAccountInfo({
       accountAddress: fromAccount.accountAddress,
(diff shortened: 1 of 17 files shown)


tests/e2e/transaction/signTransaction.test.ts
```diff
@@ -216,3 +218,46 @@ describe("sign transaction", () => {
...
+        new U8(newAccount.signingScheme), // to scheme
+        MoveVector.U8(newAccount.publicKey.toUint8Array()),
+        MoveVector.U8(proofSignedByCurrentPrivateKey.toUint8Array()),
+        MoveVector.U8(proofSignedByNewPrivateKey.toUint8Array()),
+      ],
+    },
+  });
+
+  const response = await aptos.signAndSubmitTransaction({ signer: fromAccount, transaction });
+  console.log("response", response);
```
kent-white, line 262: This looks good to me, but would be nice to test the whole flow (ie. 1. Create Prroof. 2. Create transaction with proof. 3. submit transaction and make sure it goes through)

The reason for this is I've had a lot of issues in the past actually getting the challenge in the right format. And it also seems really easy to break. Everything looks fine in the PR for contructing the challenges but...

0xmaayan: The test here 
- creates a proof
- signs the proof
- create transaction with proof
- submits successfully to chain

Were you looking for a different flow example?

BriungRi (commented):
Could the DevX be simpler here if `createProofChallenge` took two `Account` instead?

```
const account1 = new Account();
const account2 = new Account()
const proofChallenge = createProofChallenge({currentAccount: account1, newAccount: account2})
```

Seems like having to pass in all these different keys could be mistake-prone.

0xmaayan:
> Could the DevX be simpler here if `createProofChallenge` took two `Account` instead?
> 
> ```
> const account1 = new Account();
> const account2 = new Account()
> const proofChallenge = createProofChallenge({currentAccount: account1, newAccount: account2})
> ```
> 
> Seems like having to pass in all these different keys could be mistake-prone.

It is very specific - not sure all proof challenge require 2 signatures. Also, it can be tricky from a wallet perspective when there is only one signer/wallet

tests/e2e/api/proofChallenge.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    await aptos.fundAccount({ accountAddress: fromAccount.accountAddress, amount: 1_000_000_000 });
+    await aptos.fundAccount({ accountAddress: newAccount.accountAddress, amount: 1_000_000_000 });
+
+    const accountInfo = await aptos.getAccountInfo({
+      accountAddress: fromAccount.accountAddress,
+    });
+
+    const challenge = await aptos.createProofChallenge({
+      struct: "0x1::account::RotationProofChallenge",
```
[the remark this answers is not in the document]
0xmaayan: wdym? it is a generic implementation so any struct serves as a proof challenge is valid here

PR #364, opened 2024-04-15
add new apt to get all user tx versions and update get user txs

0xaptosj:


src/types/generated/operations.ts
```diff
@@ -55,7 +55,6 @@ export type CurrentTokenOwnershipFieldsFragment = {
     token_properties: any;
     token_standard: string;
     token_uri: string;
-    decimals: any;
```
0xaptosj, line 58: this is the result of running `indexer-codegen`

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 - [`Breaking`] Change any generate transaction function to return `SimpleTransaction` or `MultiAgentTransaction` instance
 - Adds `getUserTransactionHash` which can generate a transaction hash after signing, but before submission
 - Add function to create resource address locally
+- Add `getAccountAllTransactionVersions` which returns versions of all transactions related to the given address.
```
0xmaayan, line 24: These 2 should be under a new dated release version

src/types/generated/operations.ts
0xmaayan: This is a breaking change - have we talked with Indexer team on why they removed it?

0xmaayan (changes requested):
I am not sure about this solution. This can increase the dapp server requests and can potentially lead to rate limit or just worst performance (like the Explorer that suffers from a bad performance - site load, etc)

I think we should push for an Indexer solution @bowenyang007

src/internal/account.ts
```diff
@@ -137,12 +141,45 @@ export async function getTransactions(args: {
...
+  // TODO: Ideally indexer should provide one API that returns all transactions details in one go
+  // But now we have to query all transaction versions first and then query each transaction by version
+  const versions = await getAccountAllTransactionVersions({ aptosConfig, accountAddress, options });
+  const results = [];
+  for (const version of versions) {
+    const tx = getTransactionByVersion({ aptosConfig, ledgerVersion: version });
+    results.push(tx);
+  }
+  return Promise.all(results);
+}
```
gregnazario, line 153: This significantly changes the behavior of this function from being a paginated API that is associated with an account, to having a non-paginated API across two different services.

If this is the case that this is important, it should do whatever query the explorer and wallets do, or have a dedicated query for it.

Additionally, this changes what is a 1 network call, to 1 + n network calls.

In a limited network environment, it would be slow or eat up all your bandwidth.

gregnazario (changes requested):
I'd like to see an indexer only solution for a single request that is paginated if we choose to go down this route.

Otherwise, this is way too costly.

0xaptosj:
discussion happening in this [thread](https://aptos-org.slack.com/archives/C03MN5F7WUV/p1712848849421019)

0xaptosj:
Synced with Bowen offline, tldr: i think we have to keep this slow approach, because user_transactions table would be gone within a month cuz it's too expensive. Indexer team might come up with a better solution in the future but for now this is the recommended way by them.

CHANGELOG.md
0xaptosj: fixed

0xaptosj:
Not sure why `decimals` field is removed in some queries, as I don't see it being removed on hasura. I added them back manually. We should probably investigate it later.

src/internal/account.ts
```diff
@@ -137,12 +141,45 @@ export async function getTransactions(args: {
   options?: PaginationArgs;
 }): Promise<TransactionResponse[]> {
   const { aptosConfig, accountAddress, options } = args;
-  return paginateWithCursor<{}, TransactionResponse[]>({
+  // TODO: Ideally indexer should provide one API that returns all transactions details in one go
+  // But now we have to query all transaction versions first and then query each transaction by version
+  const versions = await getAccountAllTransactionVersions({ aptosConfig, accountAddress, options });
```
0xmaayan, line 146: can we keep this query as is, and add another query like `getAllUserTransactions` that does this multiple queries? and mention it on the function comment

0xmaayan:
> Not sure why `decimals` field is removed in some queries, as I don't see it being removed on hasura. I added them back manually. We should probably investigate it later.

might be related https://github.com/aptos-labs/aptos-ts-sdk/pull/314
Ideally, we should not add anything manually but generate whatever is on Hasura so we wont introduce any weird/wrong behavior

tests/e2e/api/account.test.ts
```diff
@@ -137,6 +137,71 @@ describe("account api", () => {
...
+        transaction: rawTxn2,
+      });
+      const response2 = await aptos.transaction.submit.simple({
+        transaction: rawTxn2,
+        senderAuthenticator: authenticator2,
+      });
+      const txn2 = await aptos.waitForTransaction({ transactionHash: response2.hash });
+
+      // sleep for 500 ms to ensure the txns are in the ledger
+      await new Promise((resolve) => setTimeout(resolve, 500));
```
0xaptosj, line 187: without the sleep tx2 wouldn't be included, pretty weird considering i already waitForTransaction on tx 2

tests/e2e/api/account.test.ts
```diff
@@ -137,6 +137,73 @@ describe("account api", () => {
...
+      });
+      const response2 = await aptos.transaction.submit.simple({
+        transaction: rawTxn2,
+        senderAuthenticator: authenticator2,
+      });
+      const txn2 = await aptos.waitForTransaction({ transactionHash: response2.hash });
+
+      // sleep for 500 ms to ensure the txns are in the ledger
+      await new Promise((resolve) => {
+        setTimeout(resolve, 500);
```
0xaptosj, line 188: without the sleep tx2 wouldn't be included, pretty weird considering i already waitForTransaction on tx2

PR #366, opened 2024-04-16
Change the 'waitForTransaction' SDK API to try long poll

manudhundi:
In aptos-core we now have exposed the API to do a long poll while waiting for a txn (commit https://github.com/aptos-labs/aptos-core/commit/29a9f0c3f38bb3ddab0517b0fa7aa20804b57f37). This waits for about 1s for the txn to complete.

In this change, the 'waitForTransaction' API in the SDK tries to call the long wait with the goal of exposing the subsecond E2E latencies to the end users. In the long wait is not sufficient then we fall back to polling based wait we already have

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 
 - Export `LocalNode` module using the module relative path
 
+- Change the `waitForTransaction` SDK API to try long poll
+
 # 1.12.2 (2024-04-10)
 
 - Revert export `LocalNode` module
--- a/src/internal/transaction.ts
+++ b/src/internal/transaction.ts
@@ -149,14 +149,14 @@ export async function waitForTransaction(args: {
 
   // If the transaction is pending, we do a long wait once to avoid polling
   if (isPending) {
-    const startTime = performance.now();
+    const startTime = Date.now();
     try {
       lastTxn = await longWaitForTransaction({ aptosConfig, transactionHash });
       isPending = lastTxn.type === TransactionResponseType.Pending;
     } catch (e) {
       handleAPIError(e);
     }
-    timeElapsed = (performance.now() - startTime) / 1000;
+    timeElapsed = (Date.now() - startTime) / 1000;
   }
 
   // Now we do polling to see if the transaction is still pending
--- a/src/internal/transaction.ts
+++ b/src/internal/transaction.ts
@@ -97,6 +97,19 @@ export async function isTransactionPending(args: {
   }
 }
 
+async function longWaitForTransaction(args: {
+  aptosConfig: AptosConfig;
+  transactionHash: HexInput;
+}): Promise<TransactionResponse> {
+  const { aptosConfig, transactionHash } = args;
+  const { data } = await getAptosFullNode<{}, TransactionResponse>({
+    aptosConfig,
+    path: `transactions/wait_by_hash/${transactionHash}`,
+    originMethod: "waitTransactionByHash",
+  });
+  return data;
+}
+
 export async function waitForTransaction(args: {
   aptosConfig: AptosConfig;
   transactionHash: HexInput;
@@ -108,6 +121,8 @@ export async function waitForTransaction(args: {
 
   let isPending = true;
   let timeElapsed = 0;
+  let longWaitOnce = false;
+  let isFirstTry = true;
   let lastTxn: TransactionResponse | undefined;
   let lastError: AptosApiError | undefined;
   let backoffIntervalMs = 200;
@@ -118,8 +133,16 @@ export async function waitForTransaction(args: {
       break;
     }
     try {
-      // eslint-disable-next-line no-await-in-loop
-      lastTxn = await getTransactionByHash({ aptosConfig, transactionHash });
+      if (!isFirstTry && !longWaitOnce) {
+        // We are greedy, hence do a 'long wait' once to shave off latencies resulting from polling with backoff
+        longWaitOnce = true;
+        // eslint-disable-next-line no-await-in-loop
+        lastTxn = await longWaitForTransaction({ aptosConfig, transactionHash });
+      } else {
+        isFirstTry = false;
+        // eslint-disable-next-line no-await-in-loop
+        lastTxn = await getTransactionByHash({ aptosConfig, transactionHash });
+      }
 
       isPending = lastTxn.type === TransactionResponseType.Pending;
 
--- a/src/internal/transaction.ts
+++ b/src/internal/transaction.ts
@@ -141,7 +141,6 @@ export async function waitForTransaction(args: {
 
   // check to see if the txn is already on the blockchain
   try {
-    // eslint-disable-next-line no-await-in-loop
     lastTxn = await getTransactionByHash({ aptosConfig, transactionHash });
     isPending = lastTxn.type === TransactionResponseType.Pending;
   } catch (e) {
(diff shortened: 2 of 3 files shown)


src/internal/transaction.ts
```diff
@@ -118,8 +133,16 @@ export async function waitForTransaction(args: {
...
     try {
-      // eslint-disable-next-line no-await-in-loop
-      lastTxn = await getTransactionByHash({ aptosConfig, transactionHash });
+      if (!isFirstTry && !longWaitOnce) {
+        // We are greedy, hence do a 'long wait' once to shave off latencies resulting from polling with backoff
+        longWaitOnce = true;
+        // eslint-disable-next-line no-await-in-loop
+        lastTxn = await longWaitForTransaction({ aptosConfig, transactionHash });
+      } else {
+        isFirstTry = false;
```
gregnazario, line 142: These two variables are the same right?

Maybe just make isFirstTry = false when you first enter the first section
```suggestion
      if (!triedOnce) {
        // We are greedy, hence do a 'long wait' once to shave off latencies resulting from polling with backoff
        triedOnce = true;
        // eslint-disable-next-line no-await-in-loop
        lastTxn = await longWaitForTransaction({ aptosConfig, transactionHash });
      } else {
```

src/internal/transaction.ts
```diff
@@ -108,6 +121,8 @@ export async function waitForTransaction(args: {
 
   let isPending = true;
   let timeElapsed = 0;
+  let longWaitOnce = false;
+  let isFirstTry = true;
```
gregnazario, line 125:
```suggestion
  let triedOnce = false;
```

src/internal/transaction.ts
manudhundi: No, the logic we want is:
Try the polling way in the first attempt hoping there is no need for a wait.
If that doesn't work try to do long wait for 1s.
If the long wait also is not enough, then fallback to polling

gregnazario: Okay so:
1. Short wait
2. Long wait
3. Short wait loop

src/internal/transaction.ts
manudhundi: Yes

src/internal/transaction.ts
manudhundi: Yup

banool:
Could we get a proper title, summary, test plan, etc?

src/internal/transaction.ts
```diff
@@ -97,6 +97,19 @@ export async function isTransactionPending(args: {
...
 
+async function longWaitForTransaction(args: {
+  aptosConfig: AptosConfig;
+  transactionHash: HexInput;
+}): Promise<TransactionResponse> {
+  const { aptosConfig, transactionHash } = args;
+  const { data } = await getAptosFullNode<{}, TransactionResponse>({
+    aptosConfig,
+    path: `transactions/wait_by_hash/${transactionHash}`,
+    originMethod: "waitTransactionByHash",
```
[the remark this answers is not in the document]
manudhundi: Thanks for pointing this, updated

src/internal/transaction.ts
```diff
@@ -113,6 +126,41 @@ export async function waitForTransaction(args: {
...
+    }
+    lastError = e;
+    const isRequestError = e.status !== 404 && e.status >= 400 && e.status < 500;
+    if (isRequestError) {
+      throw e;
+    }
+  }
+
+  // check to see if the txn is already on the blockchain
+  try {
```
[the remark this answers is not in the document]
manudhundi: We introduced a new API for long poll in aptos-core. Essentially it waits for about 1s for a txn to complete (goal of this is to reflect the subsecond E2E to the client). 

So now with this change, we want this behavior in SDK:
// Short lookup
if (short successful) {
   return short
}

// Long lookup
if (long successful) {
  return long
}

// Short lookup loop
loop until timeout {
   if (short successful) {
      return short
   }
}

throw error

src/internal/transaction.ts
```diff
@@ -113,6 +126,41 @@ export async function waitForTransaction(args: {
...
+  // If the transaction is pending, we do a long wait once to avoid polling
+  if (isPending) {
+    try {
+      // eslint-disable-next-line no-await-in-loop
+      lastTxn = await longWaitForTransaction({ aptosConfig, transactionHash });
+      isPending = lastTxn.type === TransactionResponseType.Pending;
+    } catch (e) {
+      handleAPIError(e);
+    }
+    timeElapsed = 1; // Long wait takes about 1 second
```
sitalkedia, line 160: Why don't we measure the actual time here?

0xmaayan:
Please update CHANGELOG

src/internal/transaction.ts
manudhundi: done

manudhundi:
> Please update CHANGELOG

done

tests/e2e/api/transaction.test.ts
```diff
@@ -86,4 +87,52 @@ describe("transaction api", () => {
...
+    let txn: TransactionResponse;
+    beforeAll(async () => {
+      const senderAccount = Account.generate();
+      await aptos.fundAccount({ accountAddress: senderAccount.accountAddress, amount: FUND_AMOUNT });
+      const bob = Account.generate();
+      const rawTxn = await aptos.transaction.build.simple({
+        sender: senderAccount.accountAddress,
+        data: {
+          function: "0x1::aptos_account::transfer",
+          functionArguments: [bob.accountAddress, new U64(10)],
```
0xmaayan, line 101: nit
```suggestion
          functionArguments: [bob.accountAddress, 10],
```

tests/e2e/api/transaction.test.ts
```diff
@@ -86,4 +87,52 @@ describe("transaction api", () => {
...
+        transaction: rawTxn,
+      });
+      const response = await aptos.transaction.submit.simple({
+        transaction: rawTxn,
+        senderAuthenticator: authenticator,
+      });
+      txn = await longWaitForTransaction({ aptosConfig: aptos.config, transactionHash: response.hash });
+    });
+
+    test("it queries for transactions on the chain", async () => {
```
0xmaayan, line 115: not completely sure what we test here? we already have tests for those "get transaction by * " queries

tests/e2e/api/transaction.test.ts
```diff
@@ -86,4 +87,52 @@ describe("transaction api", () => {
...
+      const transactions = await aptos.getTransactions();
+      expect(transactions.length).toBeGreaterThan(0);
+    });
+
+    test("it queries for transactions by version", async () => {
+      if (!("version" in txn)) {
+        throw new Error("Transaction is still pending!");
```
0xmaayan, line 122: I dont think we want to throw while testing

tests/e2e/api/transaction.test.ts
manudhundi: done

tests/e2e/api/transaction.test.ts
manudhundi: removed

tests/e2e/api/transaction.test.ts
manudhundi: ok removing it. 
Will only test aptos.getTransactionByHash(..) to ensure that longWaitForTransaction(...) works as expected

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 
 - Export `LocalNode` module using the module relative path
```
gregnazario, line 11: nit:you can remove the space here

src/internal/transaction.ts
```diff
@@ -113,6 +126,40 @@ export async function waitForTransaction(args: {
...
+  try {
+    lastTxn = await getTransactionByHash({ aptosConfig, transactionHash });
+    isPending = lastTxn.type === TransactionResponseType.Pending;
+  } catch (e) {
+    handleAPIError(e);
+  }
+
+  // If the transaction is pending, we do a long wait once to avoid polling
+  if (isPending) {
+    const startTime = performance.now();
```
gregnazario, line 152: I don't know what performance is, but I'm assuming that's used for timing

gregnazario (approved):

src/internal/transaction.ts
manudhundi: will use Date.now()

PR #367, opened 2024-04-16
[deps] Switch crypto library to @noble/curves

gregnazario:
### Description
We already use `@noble/curves/secp256k1` for other signing, this brings it in alignment so we only use one signing library `@noble/curves`.

It generally reads better and is more recently updated.

### Test Plan
CI


alinush (approved):
LGTM!

0xmaayan (approved):
Please update changelog

PR #370, opened 2024-04-17
[crypto] Add malleability check on EDD2519 verification

gregnazario:
### Description
Adds malleability verification to SDK, copied from Rust code

### Test Plan
See unit tests

### DIFF
--- a/src/core/crypto/ed25519.ts
+++ b/src/core/crypto/ed25519.ts
@@ -14,6 +14,14 @@ import { Signature } from "./signature";
 import { convertSigningMessage } from "./utils";
 
 /**
+ * L is the value that greater than or equal to will produce a non-canonical signature, and must be rejected
+ */
+const L: number[] = [
+  0xed, 0xd3, 0xf5, 0x5c, 0x1a, 0x63, 0x12, 0x58, 0xd6, 0x9c, 0xf7, 0xa2, 0xde, 0xf9, 0xde, 0x14, 0x00, 0x00, 0x00,
+  0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x10,
+];
+
+/**
  * Represents the public key of an Ed25519 key pair.
  *
  * Since [AIP-55](https://github.com/aptos-foundation/AIPs/pull/263) Aptos supports
@@ -65,6 +73,10 @@ export class Ed25519PublicKey extends AccountPublicKey {
     const messageBytes = Hex.fromHexInput(messageToVerify).toUint8Array();
     const signatureBytes = signature.toUint8Array();
     const publicKeyBytes = this.key.toUint8Array();
+    // Also verify malleability
+    if (!signature.checkSMalleability()) {
+      return false;
+    }
 
     return ed25519.verify(signatureBytes, messageBytes, publicKeyBytes);
   }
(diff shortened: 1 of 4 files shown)


src/core/crypto/ed25519.ts
```diff
@@ -313,5 +325,23 @@ export class Ed25519Signature extends Signature {
     return new Ed25519Signature(bytes);
   }
 
+  /**
+   * Checks if an ED25519 signature is non-canonical.
+   */
+  checkSMalleability(): boolean {
```
alinush, line 331: This is a bad name. I realize it is inherited from our Rust codebase.

I would rename this to `isCanonicalSignature`.

tests/unit/ed25519.test.ts
```diff
@@ -49,6 +49,16 @@ describe("Ed25519PublicKey", () => {
     ).toBe(false);
   });
 
+  it("should fail malleable signatures", () => {
+    const signature = new Ed25519Signature(
+      // eslint-disable-next-line max-len
+      "ad0828843228048fee9ead35cf0decc8155342e967993851f70bfc0bb9e97719a094c6f956e5cd65872fa769e7dce26f6e637eb4ff70c3f6aafc5f06fbc300ad",
```
alinush, line 55: Can you document how you obtained this?

src/core/crypto/ed25519.ts
```diff
@@ -313,5 +325,23 @@ export class Ed25519Signature extends Signature {
...
+    const s = this.toUint8Array().slice(32);
+
+    for (let i = s.length - 1; i >= 0; i -= 1) {
+      if (s[i] < L[i]) {
```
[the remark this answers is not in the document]
gregnazario: This is copied from your implementation in the Rust code.  It's possible I got it backwards

tests/unit/ed25519.test.ts
gregnazario: This is from the GitHub issue... I've been unable to generate a non-canonical signature

src/core/crypto/ed25519.ts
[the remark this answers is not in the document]
alinush: Oh, I see. This is a short-circuiting trick.

Then the only Q left is about the byte ordering. For this to work, `L` and `s` have to be in LE form, with the MSB starting at `s[s.length - 1]`

tests/unit/ed25519.test.ts
alinush: Cool! Just add a link to the GH issue.

To obtain a non canonical signature, just take `L` and increment it. Then set `s` to be that value. :) Then, set the other component of the signature to be junk, since verification will short-circuit on `s`.

alinush: Alternatively, create a normal canonical signature, take its `s` and set it to `s + L` or `s + 2L` or any `s + kL` for `k > 0`.

src/core/crypto/ed25519.ts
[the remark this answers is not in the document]
alinush: I guess this is the case since the SDK produces signatures that pass the Rust verification.

alinush (approved):

tests/unit/ed25519.test.ts
gregnazario: Fixed to test L and L + 1

PR #379, opened 2024-04-26
Refactor core/accounts  Remove private key from the Account interface.  Separate out the signing message functionality.  Adds MultiKeyAccount

heliuchuan:
### Description

This is all groundwork refactor for the keyless PR.  I encountered circular dependencies when developing keyless due to accounts being under /core and a lot of packages importing Account. 

- Moves src/core/account to src/account
- Adds MultiKeyAccount class
- Removes privateKey from Account interface
- Refactors signing message functions

Trying to merge this while Keyless work is ongoing to reduce diff and merge conflicts.


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 
 # Unreleased
 
+- [`Breaking`] Removes private key from the Account class to support keyless and passkey
```
0xmaayan, line 7: do we really need this change part of this PR? why not having it as part of the keyless PR?

src/transactions/transactionBuilder/signingMessage.ts
```diff
@@ -0,0 +1,70 @@
...
+  const body = bytes;
+
+  const mergedArray = new Uint8Array(prefix.length + body.length);
+  mergedArray.set(prefix);
+  mergedArray.set(body, prefix.length);
+
+  return mergedArray;
+}
+
+export function generateSigningMessageForSerializable(obj: Serializable): Uint8Array {
```
0xmaayan, line 57: Please provide a test for this function

0xmaayan, line 57: Can we think on a better variable name other than `obj`?
```suggestion
export function generateSigningMessageForSerializable(obj: Serializable): Uint8Array {
```

src/transactions/transactionBuilder/signingMessage.ts
```diff
@@ -0,0 +1,70 @@
...
+  return generateSigningMessage(obj.bcsToBytes(), obj.constructor.name);
```
0xmaayan, line 58: what is the `obj.constructor.name`?

0xmaayan (changes requested):
This PR is missing
- Method doc comments
- Method return types
- Class variable doc comments
- Tests for new functions

Also, removing the private key from the Account class might need to be released with a major release, lets keep it part of the keyless PR.

I think we can keep this PR to only adding `Account implementation for MultiKey accounts` and `file organize`

src/account/MultiKeyAccount.ts
```diff
@@ -0,0 +1,113 @@
...
+    const signatures = [];
+    for (const signer of this.signers) {
+      signatures.push(signer.sign(data));
+    }
+    return new MultiSignature({ signatures, bitmap: this.signaturesBitmap });
+  }
+
+  signTransaction(transaction: AnyRawTransaction) {
+    const signatures = [];
+    for (const signer of this.signers) {
```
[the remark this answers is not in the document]
heliuchuan: Not sure what you mean, there is no main signer.

CHANGELOG.md
heliuchuan: Yes, multikey accounts don't have a private key

src/core/crypto/multiKey.ts
```diff
@@ -139,9 +139,19 @@ export class MultiKey extends AccountPublicKey {
...
+
+    if (index !== -1) {
+      return index;
+    }
+    throw new Error("Public key not found in MultiKey");
+  }
 }
 
-export class MultiKeySignature extends Signature {
+export class MultiSignature extends Signature {
```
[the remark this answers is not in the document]
heliuchuan: I thought it made more sense.  Its a signature, not a key signature.  I view it as key/signature and just prefixing with multi.

src/transactions/transactionBuilder/signingMessage.ts
heliuchuan: The name of the object/class type.

https://stackoverflow.com/questions/46217853/get-constructor-name-of-object

src/transactions/transactionBuilder/signingMessage.ts
heliuchuan: idk hows serializable

CHANGELOG.md
0xmaayan: yes, [MultiKey](https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/core/crypto/multiKey.ts#L27) does not have a private key.

This change however is to remove the private key from the Account to support keyless and passkey - consider it is a breaking change, could we include it on the keyless PR instead of this PR?

src/core/crypto/multiKey.ts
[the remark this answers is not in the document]
0xmaayan: renaming exported class methods is a breaking change - if there is no special reason we should avoid it.

tests/unit/signingMessage.test.ts
```diff
@@ -0,0 +1,101 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { Account, Ed25519PrivateKey, generateSigningMessageForSerializable, generateSigningMessageForTransaction } from "../../src";
+import { getAptosClient } from "../e2e/helper";
+
+const { aptos } = getAptosClient();
+const TRANSFER_AMOUNT = 100;
+
+describe("generateSigningMessage ", () => {
+  const alice = Account.fromPrivateKey({
```
0xmaayan, line 11: lets use this file https://github.com/aptos-labs/aptos-ts-sdk/blob/main/tests/unit/helper.ts for pre-generated wallets/accounts

src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -565,48 +537,52 @@ export function generateUserTransactionHash(args: InputSubmitTransactionData): s
...
-      transaction.rawTransaction,
-      transaction.secondarySignerAddresses ?? [],
-      transaction.feePayerAddress,
+export function generateMultiSignersSignedTransaction(
+  transaction: MultiAgentRawTransaction | FeePayerRawTransaction,
+  senderAuthenticator: AccountAuthenticator,
+  feePayerAuthenticator?: AccountAuthenticator,
+  additionalSignersAuthenticators?: Array<AccountAuthenticator>,
+) {
+  if (transaction instanceof FeePayerRawTransaction) {
```
0xmaayan, line 554: `instanceOf` creates bundle issues, we switched to use type checking, check out https://github.com/aptos-labs/aptos-ts-sdk/pull/393/files

src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -451,12 +446,7 @@ export function generateSignedTransactionForSimulation(args: InputSimulateTransa
 export function getAuthenticatorForSimulation(publicKey: PublicKey) {
   // TODO add support for AnyMultiKey
   if (publicKey instanceof AnyPublicKey) {
-    if (publicKey.publicKey instanceof Ed25519PublicKey) {
-      return new AccountAuthenticatorSingleKey(publicKey, new AnySignature(new Ed25519Signature(new Uint8Array(64))));
-    }
-    if (publicKey.publicKey instanceof Secp256k1PublicKey) {
-      return new AccountAuthenticatorSingleKey(publicKey, new AnySignature(new Secp256k1Signature(new Uint8Array(64))));
```
0xmaayan, line 458: why remove?

src/account/Ed25519Account.ts
```diff
@@ -75,13 +77,17 @@ export class Ed25519Account implements Account {
     return this.publicKey.verifySignature(args);
   }
 
-  signWithAuthenticator(message: HexInput) {
-    const signature = this.privateKey.sign(message);
+  signWithAuthenticator(transaction: AnyRawTransaction) {
```
0xmaayan, line 80: why this change?

CHANGELOG.md
heliuchuan: I think it makes sense to have it in this PR imo.  the [MultiKey](https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/core/crypto/multiKey.ts#L27) abstraction is justification enough i feel.

either way, the keyless stuff will be in the same cut.

src/transactions/transactionBuilder/transactionBuilder.ts
heliuchuan: It simplifies the logic no?

All AnyPublicKey will use AccountAuthenticatorSingleKey.

No need to check if its Ed25519PublicKey or Secp256k1PublicKey (or KeylessPublicKey for that matter)

src/account/Ed25519Account.ts
heliuchuan: Signing with an authenticator really makes sense if it's an AnyRawTransaction.

And I think the signing message generation logic should be explicit.  The motivation for this change is that Keyless will do it a bit differently.

src/transactions/transactionBuilder/transactionBuilder.ts
0xmaayan: dont you need different signatures for ed and secp?

CHANGELOG.md
0xmaayan: what "same cut"? the sdk doesnt follow aptos framework cut timeline. If we merge it in, next SDK release will include this change and would probably need to have a major version upgrade. 

If this change is to support keyless, we should get it in with the keyless integration PR

src/core/crypto/multiKey.ts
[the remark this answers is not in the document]
heliuchuan: It is a breaking change, however I think people would only be using 'MultiKeySignature' due to the lack of a MultiKeyAccount abstraction.

imo not many use this functionality, it's not a hard fix for devs, and its a good incentive for people to use the MultiKeyAccount when they upgrade

but I get not wanting to make unneeded breaking changes, but I also worry about that rationale forever locking us into naming that is just wrong or inconsistent.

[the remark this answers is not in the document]
0xmaayan: Let's mention it on changelog as `Breaking`

src/account/Account.ts
```diff
@@ -222,19 +216,26 @@ export abstract class Account {
...
+   * @returns Signature
    */
   abstract sign(message: HexInput): Signature;
 
+  /**
+   * Sign the given transaction.
+   * @param message in HexInput format
+   * @returns Signature
+   */
+  abstract signTransaction(transaction: AnyRawTransaction): Signature;
```
0xmaayan, line 237: is there a reason we need this function? when building the sdk, we intentionally decided to return  `AccountAuthenticator` from a sign functionality as it makes more sense and help with dev exp cc @hardsetting

0xmaayan (commented):
I am not completely sure about the motivation for this PR, it feels like a lot of different changes squeezed into one PR. Can we split it into separated PRs for each change?

Also, I can't find justifications for different changes like return Signature from sign function, change signWithAuth to accept only a transaction, and MultiKeySignature rename. 

Also, iiuc, some breaking changes are only to support the future keyless integration - can we include whatever is keyless relevant in the future keyless PR?

src/transactions/transactionBuilder/signingMessage.ts
```diff
@@ -0,0 +1,93 @@
...
+    return new MultiAgentRawTransaction(transaction.rawTransaction, transaction.secondarySignerAddresses);
+  }
+
+  return transaction.rawTransaction;
+}
+
+/**
+ * Generates the 'signing message' form of a message to be signed.
+ *
+ * @param bytes The byte representation of the message to be signed and sent to the chain
```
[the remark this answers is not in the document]
heliuchuan: these are signing messages, not signed message.

this takes a thing to be signed and then is transforms it into a 'signing message'

where it prepends the hash of the domain separator.  its a thing specific to our blockchain. 

see bcscryptohash
https://github.com/aptos-labs/aptos-core/blob/e2f261d0a5b75bb9f7db998bd28f28acd3b741fd/crates/aptos-crypto-derive/src/lib.rs#L444

src/transactions/transactionBuilder/transactionBuilder.ts
heliuchuan: Oh fair enough I missed that distinction.  In practice though it doesn't matter because this is for simulation.  Like it's just returning an empty signature at the end of the day.

heliuchuan: I'll just change it back

src/transactions/transactionBuilder/signingMessage.ts
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Generates the 'signing message' form of a transaction. It derives the type of transaction and
+ * applies the appropriate domain separator based on if there is extra data such as a fee payer or
+ * secondary signers.
+ *
+ * @param transaction A transaction that is to be signed
+ *
+ * @returns The Uint8Array of the signing message
+ */
+export function generateSigningMessageForTransaction(transaction: AnyRawTransaction): Uint8Array {
```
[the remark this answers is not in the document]
heliuchuan: rawTransactionToBytes would be misleading because its not simply "to bytes".

it's rawTransactionToSigningMessage, which is not that much better

src/internal/transactionSubmission.ts
```diff
@@ -204,8 +202,8 @@ export function getSigningMessage(args: { transaction: AnyRawTransaction }): Uin
  * @return The signer AccountAuthenticator
  */
 export function signTransaction(args: { signer: Account; transaction: AnyRawTransaction }): AccountAuthenticator {
-  const accountAuthenticator = sign({ ...args });
-  return accountAuthenticator;
+  const { signer, transaction } = args;
+  return signer.signWithAuthenticator(transaction);
```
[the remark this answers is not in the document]
heliuchuan: Not sure what you mean.  I'm deleting the sign function which is just a layer of unneeded indirection

src/account/Account.ts
heliuchuan: Right, we still maintain `signWithAuthenticator` so I'm not taking anything away.  I add `signTransaction` for the ease of implementing signWithAuthenticator in the MultiKeyAccount.  I just think the decomposition makes the code more readable and the abstract class easier to implement.  I'd be totally fine making it private if you feel it would be misused by devs or confuse them.  But I figured why not *shrug*.  lmk what you prefer.

heliuchuan:
> I am not completely sure about the motivation for this PR, it feels like a lot of different changes squeezed into one PR. Can we split it into separated PRs for each change?

I could but ultimately am lazy and don't quite have a flow for stacked PRs yet (I should try graphite).  In the future I will do this though.  This PR is already split from the main Keyless PR to be more reviewable.

> Also, I can't find justifications for different changes like 
return Signature from sign function

Not sure what that is referring to

> change signWithAuth to accept only a transaction

Ok, so I need to move the generateSigningMessage logic to accounts.  It used to live [here](https://github.com/aptos-labs/aptos-ts-sdk/blob/5b66470cfd454a95201d10e1b415e724e57a805d/src/transactions/transactionBuilder/transactionBuilder.ts#L477) in a function called `sign` before this PR.  I need to do this because Keyless breaks the assumption of how we normally generate these signing messages.  It does a lot of custom stuff.  But this custom stuff is specific to the context of signing transactions as in the `sign` function.  I realize now an Authenticator is a (publicKey, signature) pair and can be used outside of a context of signing txns (I confused Authenticator with SignedTransaction). 

Perhaps what I want is a separate signTransactionWithAuth where the input is typed as a AnyRawTransaction.  I assume that that's ok.  Will update to do this.

> MultiKeySignature rename

This is just for 'correctness'.  I'm happy to not rename.  I'll remove this.

src/transactions/transactionBuilder/signingMessage.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @param bytes The byte representation of the message to be signed and sent to the chain
+ * @param domainSeparator A domain separator that starts with 'APTOS::'
+ *
+ * @returns The Uint8Array of the signing message
+ */
+export function generateSigningMessage(bytes: Uint8Array, domainSeparator: string): Uint8Array {
+  const hash = sha3Hash.create();
+
+  if (!domainSeparator.startsWith("APTOS::")) {
```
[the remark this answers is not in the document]
heliuchuan: Because if it's not prefixed with APTOS:: ur using it wrong

tests/e2e/transaction/transactionBuilder.test.ts
```diff
@@ -564,7 +548,7 @@ describe("transaction builder", () => {
         amount: 1,
       });
 
-      const senderAuthenticator = sign({ signer: alice, transaction });
+      const senderAuthenticator = alice.signTransactionWithAuthenticator(transaction);
```
0xmaayan, line 551: This should also be included as `[Breaking]` on changelog. Some users out there use the `transactionBuilder` directly

0xmaayan (commented):
Once it lands, a new major version will need to be released because of the remove of the private key from the Account class

src/account/Ed25519Account.ts
```diff
@@ -76,12 +78,19 @@ export class Ed25519Account implements Account {
   }
 
   signWithAuthenticator(message: HexInput) {
-    const signature = this.privateKey.sign(message);
-    return new AccountAuthenticatorEd25519(this.publicKey, signature);
+    return new AccountAuthenticatorEd25519(this.publicKey, this.privateKey.sign(message));
+  }
+
+  signTransactionWithAuthenticator(transaction: AnyRawTransaction) {
```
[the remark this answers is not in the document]
heliuchuan: These are inherited from the base account class...(at least in vs code it shows when i hover over?) do we need to copy paste them everywhere?

src/account/MultiKeyAccount.ts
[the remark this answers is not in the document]
heliuchuan: not quite sure.  not an easy thing to deal with UX wise.  but regardless in 1 of N cases its easy to imagine use.

src/account/Ed25519Account.ts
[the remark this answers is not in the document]
heliuchuan: Done

0xmaayan (approved):
Amazing! Thanks for all the iterations.

PR #380, opened 2024-04-29
Fix transaction submission from different bundle

hardsetting:
### Description
When using a wallet extension, objects returned from the wallet belong to the wallet's bundle.
If we pass those objects directly to the ts-sdk, the `instanceof` operator doesn't work.

The change implemented in this PR is a temporary fix, to make sure that the submission flow from a walle extension works as expected.

Ideally, we should try to avoid `instanceof` as much as possible, unless we can guarantee the object was instantiated from the same bundle.

### Test Plan
CI

https://github.com/aptos-labs/aptos-ts-sdk/assets/7707719/51e8b0a8-5f69-44f2-9e20-249cec4f2ce4

### DIFF
--- a/src/transactions/transactionBuilder/transactionBuilder.ts
+++ b/src/transactions/transactionBuilder/transactionBuilder.ts
@@ -27,10 +27,10 @@ import { normalizeBundle } from "../../utils/normalizeBundle";
 import {
   AccountAuthenticator,
   AccountAuthenticatorEd25519,
-  AccountAuthenticatorMultiKey,
   AccountAuthenticatorSingleKey,
 } from "../authenticator/account";
 import {
+  TransactionAuthenticator,
   TransactionAuthenticatorEd25519,
   TransactionAuthenticatorFeePayer,
   TransactionAuthenticatorMultiAgent,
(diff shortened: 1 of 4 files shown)


src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -494,14 +495,12 @@ export function sign(args: { signer: Account; transaction: AnyRawTransaction }):
...
+  const { transaction, feePayerAuthenticator, additionalSignersAuthenticators } = args;
+  const senderAuthenticator = normalizeBundle(AccountAuthenticator, args.senderAuthenticator);
 
   const transactionToSubmit = deriveTransactionType(transaction);
 
-  if (
-    (feePayerAuthenticator || additionalSignersAuthenticators) &&
-    (transactionToSubmit instanceof MultiAgentRawTransaction || transactionToSubmit instanceof FeePayerRawTransaction)
-  ) {
+  if ("secondary_signer_addresses" in transactionToSubmit) {
```
0xmaayan, line 503: this is a hack that works because 
- feePayer has secondarySignerAddresses as an array of addresses or an empty array, so it "passes" this check
- multiAgent always has secondarySignerAddresses
right? 

Let's add a comment that this is a hack so one wont think and expect feePayerTransaction to always need secondarySignerAddresses

src/utils/normalizeBundle.ts
```diff
@@ -0,0 +1,21 @@
...
+};
+
+/**
+ * Utility function that serializes and deserialize an object back into the same bundle as the sdk.
+ * This is a workaround to have the `instanceof` operator work when input objects come from a different
+ * bundle.
+ * @param cls The class of the object to normalize
+ * @param value the instance to normalize
+ */
+export function normalizeBundle<T extends Serializable>(cls: DeserializableClass<T>, value: T) {
```
0xmaayan, line 17: you are a wild person

0xmaayan (commented):
U R WILD!

Have you tested it with Petra?

hardsetting:
> Have you tested it with Petra?

Just added a video of me using Petra with the new wallet standard.

src/transactions/transactionBuilder/transactionBuilder.ts
hardsetting: Not really a hack, just looks hacky :(
Everything is strong typed and typescript correctly picks it up.
`transactionToSubmit` is `RawTransaction | FeePayerRawTransaction | MultiAgentRawTransaction`, and the last two always have that field defined.

src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -295,9 +296,8 @@ export async function generateRawTransaction(args: {
 
   const { maxGasAmount, gasUnitPrice, expireTimestamp } = {
     maxGasAmount: options?.maxGasAmount ? BigInt(options.maxGasAmount) : BigInt(DEFAULT_MAX_GAS_AMOUNT),
-    gasUnitPrice: BigInt(gasEstimate),
-    expireTimestamp: BigInt(Math.floor(Date.now() / 1000) + DEFAULT_TXN_EXP_SEC_FROM_NOW),
-    ...options,
+    gasUnitPrice: options?.gasUnitPrice ?? BigInt(gasEstimate),
```
0xmaayan, line 299: ❤️

0xmaayan (approved):
LGTM. Please update changelog =)

src/transactions/transactionBuilder/transactionBuilder.ts
hardsetting: Futher cleaned it up and removed the unnecessary `deriveTransactionType` step

PR #381, opened 2024-04-29
Use `@aptos-labs/aptos-cli` as a regular dependency

0xmaayan:
### Description
Following https://aptos-org.slack.com/archives/C06KSKYNN7R/p1714309654711519


justinr1234:
@0xmaayan I'd also suggest modifying the code to not always call `npx` and also allow calling global `aptos` alone

gregnazario (approved):

0xmaayan:
> @0xmaayan I'd also suggest modifying the code to not always call `npx` and also allow calling global `aptos` alone

We are able to use the global `aptos` command. Using the `npx` is for when we want to run `aptos` in a node env and for the sdk to support built-in `aptos` functions (like run the local testnet)

PR #386, opened 2024-05-01
Typescript strict

hariria:
### Description
- [x] Added [`strict`](https://www.typescriptlang.org/tsconfig/#strict) typing for better correctness

Example:
Using `undefined` as an index will not compile when using typescript strictly
```ts
// Error: if process.env.APTOS_NETWORK is undefined, this will fail when strict because 
// it does not allow you to use an undefined type to index values
NetworkToNetworkName[process.env.APTOS_NETWORK];
```
In javascript however, `undefined` will convert to a string `undefined`

```js
// Passes: if process.env.APTOS_NETWORK is undefined, it will pass in "undefined" as an index to NetworkToNetworkName
NetworkToNetworkName[process.env.APTOS_NETWORK];
```
Thus, if you have a key in `NetworkToNetworkName` called `"undefined"`, it will unintentionally use that key / value pair

- [x] Added a `tsc` step (build) to package.json and CI to ensure all ts files are built correctly in the `examples/typescript` directory

### Test Plan
Build passes

### DIFF
--- a/.github/actions/run-examples/action.yaml
+++ b/.github/actions/run-examples/action.yaml
@@ -50,7 +50,7 @@ runs:
       with:
         max_attempts: 3
         timeout_minutes: 25
-        command: cd examples/typescript && pnpm install && pnpm test
+        command: cd examples/typescript && pnpm install && pnpm build && pnpm test
 
     - name: Print local testnet logs on failure
       shell: bash
--- a/examples/typescript/batch_funds.ts
+++ b/examples/typescript/batch_funds.ts
@@ -30,7 +30,7 @@ import {
   UserTransactionResponse,
 } from "@aptos-labs/ts-sdk";
 
-const APTOS_NETWORK: Network = NetworkToNetworkName[process.env.APTOS_NETWORK] || Network.LOCAL;
+const APTOS_NETWORK: Network = NetworkToNetworkName[process.env.APTOS_NETWORK ?? Network.DEVNET];
 
 const config = new AptosConfig({ network: APTOS_NETWORK });
 const aptos = new Aptos(config);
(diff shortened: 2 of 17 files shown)


.github/actions/run-examples/action.yaml
```diff
@@ -50,7 +50,7 @@ runs:
       with:
         max_attempts: 3
         timeout_minutes: 25
-        command: cd examples/typescript && pnpm install && pnpm test
+        command: cd examples/typescript && pnpm install && pnpm build && pnpm test
```
0xmaayan, line 53: is there a specific reason we need to build it on ci? i think we only should care that nothing breaks in the example logic on ci and not in the package build

0xmaayan (approved):

.github/actions/run-examples/action.yaml
hariria: Yeah, a couple reasons:

1. The `pnpm test` command only checks a subset of the typescript files, not all (e.g., `batch_mint.ts` is not included in `pnpm test`)
2. `tsc` runs full type checking on all `.ts` files in that directory and is typically stricter than `ts-node` which mostly just transpiles and executes the code

the build step runs almost instantly so shouldn't impact CI time at all

0xmaayan: yea, I am not worried about the time but more on the false failures we can have - anyway lets roll with it and see how it goes

hariria: Which false failures?

0xmaayan: I mean, not sure how much we care about type checking, build success, etc when running those examples. We mostly care nothing breaks if we change something in the SDK code

hariria: > I mean, not sure how much we care about type checking, build success, etc when running those examples

I feel like its important to know if the examples don't compile properly no? Especially if others are trying to use them for their own projects? I think it is achieving the goal you mentioned in the second part of your statement in determining if nothing has broken, especially since it's checking examples that are not included in the `test` script

PR #391, opened 2024-05-08
Add support for cli move commands

0xmaayan:
### Description
To enhance the SDK, and to support future work in another area, this PR adds support for commonly used CLI commands
- aptos init
- aptos move compile
- aptos move test
- aptos move publish


hariria (approved):
Cool

PR #396, opened 2024-05-14
[crypto] Add AIP-80 input types for private keys

gregnazario:
### Description
Adds support for AIP-80 private keys

### Test Plan
See unit tests

### Related Links
Discussion https://github.com/aptos-foundation/AIPs/issues/405

### DIFF
--- a/src/core/crypto/ed25519.ts
+++ b/src/core/crypto/ed25519.ts
@@ -168,17 +168,30 @@ export class Ed25519PrivateKey extends Serializable implements PrivateKey {
    * Create a new PrivateKey instance from a Uint8Array or String.
    *
    * @param hexInput HexInput (string or Uint8Array)
+   * @param strict If set, enforces private key has to use AIP-80
    */
-  constructor(hexInput: HexInput) {
+  constructor(hexInput: HexInput, strict: boolean = false) {
     super();
-
-    const privateKeyHex = Hex.fromHexInput(hexInput);
-    if (privateKeyHex.toUint8Array().length !== Ed25519PrivateKey.LENGTH) {
+    let data: Hex;
+    if (typeof hexInput === "string") {
+      // A non-prefixed input
+      if (!strict && hexInput.startsWith("0x")) {
+        data = Hex.fromHexString(hexInput);
+      } else if (hexInput.startsWith("ed25519-priv-")) {
+        // AIP-80 compatible
+        data = Hex.fromHexString(hexInput.split("-")[2]);
+      } else {
+        throw new Error("Invalid private key input type");
+      }
+    } else {
+      data = Hex.fromHexInput(hexInput);
+    }
+    if (data.toUint8Array().length !== Ed25519PrivateKey.LENGTH) {
       throw new Error(`PrivateKey length should be ${Ed25519PrivateKey.LENGTH}`);
     }
 
     // Create keyPair from Private key in Uint8Array format
-    this.signingKey = privateKeyHex;
+    this.signingKey = data;
   }
 
   /**
@@ -272,8 +285,8 @@ export class Ed25519PrivateKey extends Serializable implements PrivateKey {
    *
    * @returns string representation of the private key
    */
-  toString(): string {
-    return this.signingKey.toString();
+  toString(): `ed25519-priv-${string}` {
+    return `ed25519-priv-${this.signingKey.toString()}`;
   }
 
   // endregion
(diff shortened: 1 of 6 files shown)


src/core/crypto/ed25519.ts
```diff
@@ -251,8 +264,8 @@ export class Ed25519PrivateKey extends Serializable implements PrivateKey {
    *
    * @returns string representation of the private key
    */
-  toString(): string {
-    return this.signingKey.toString();
+  toString(): `ed25519-priv-${string}` {
```
0xmaayan, line 267: isnt it depends on how you generate the privateKey?

src/core/crypto/ed25519.ts
```diff
@@ -147,17 +147,30 @@ export class Ed25519PrivateKey extends Serializable implements PrivateKey {
...
     super();
-
-    const privateKeyHex = Hex.fromHexInput(hexInput);
-    if (privateKeyHex.toUint8Array().length !== Ed25519PrivateKey.LENGTH) {
+    let data: Hex;
+    if (typeof hexInput === "string") {
+      // A non-prefixed input
+      if (!strict && hexInput.startsWith("0x")) {
+        data = Hex.fromHexString(hexInput);
+      } else if (hexInput.startsWith("ed25519-priv-")) {
```
0xmaayan, line 159: lets have `ed25519-priv-` a global const

src/core/crypto/secp256k1.ts
```diff
@@ -107,16 +107,29 @@ export class Secp256k1PrivateKey extends Serializable implements PrivateKey {
...
     super();
-
-    const privateKeyHex = Hex.fromHexInput(hexInput);
-    if (privateKeyHex.toUint8Array().length !== Secp256k1PrivateKey.LENGTH) {
+    let data: Hex;
+    if (typeof hexInput === "string") {
+      // A non-prefixed input
+      if (!strict && hexInput.startsWith("0x")) {
+        data = Hex.fromHexInput(hexInput);
+      } else if (hexInput.startsWith("secp256k1-priv-")) {
```
0xmaayan, line 119: same. global const.

src/core/crypto/secp256k1.ts
```diff
@@ -207,8 +220,8 @@ export class Secp256k1PrivateKey extends Serializable implements PrivateKey {
    *
    * @returns string representation of the private key
    */
-  toString(): string {
-    return this.key.toString();
+  toString(): `secp256k1-priv-${string}` {
+    return `secp256k1-priv-${this.key.toString()}`;
```
0xmaayan, line 224: same question as in ed

src/core/crypto/ed25519.ts
```diff
@@ -147,17 +147,30 @@ export class Ed25519PrivateKey extends Serializable implements PrivateKey {
...
-  constructor(hexInput: HexInput) {
+  constructor(hexInput: HexInput, strict: boolean = false) {
     super();
-
-    const privateKeyHex = Hex.fromHexInput(hexInput);
-    if (privateKeyHex.toUint8Array().length !== Ed25519PrivateKey.LENGTH) {
+    let data: Hex;
+    if (typeof hexInput === "string") {
+      // A non-prefixed input
+      if (!strict && hexInput.startsWith("0x")) {
```
0xmaayan, line 157: `fromHexInput` uses `fromHexString` - do we really need this check?

0xmaayan (commented):
how does the account creation look like with aip-80? 
it looks like, with this change, there is no way for you to create an account with an aip-80 compatible private key
```
const account = Account.generate() // always produces a random private key with no prefix
```

gregnazario:
> how does the account creation look like with aip-80? it looks like, with this change, there is no way for you to create an account with an aip-80 compatible private key
> 
> ```
> const account = Account.generate() // always produces a random private key with no prefix
> ```

The private key is the same, it's just the string representation is different with a prefix to differentiate between other similarly hex encoded inputs

src/core/crypto/ed25519.ts
gregnazario: Need it to tell that it's not prefixed

0xmaayan:
> > how does the account creation look like with aip-80? it looks like, with this change, there is no way for you to create an account with an aip-80 compatible private key
> > ```
> > const account = Account.generate() // always produces a random private key with no prefix
> > ```
> 
> The private key is the same, it's just the string representation is different with a prefix to differentiate between other similarly hex encoded inputs

So the real change here is how we represent the private key as a string (with a prefix) and then adding support to when users pass this string into the sdk

0xmaayan:
@gregnazario what is the progress with that?

PR #400, opened 2024-05-18
[scripts] auto convert string type tags

gregnazario:
### Description
Allows strings to be inputted as type arguments for scripts, similar to entry functions

### Test Plan
E2E tests with type tags and with strings

### DIFF
--- a/src/transactions/types.ts
+++ b/src/transactions/types.ts
@@ -69,6 +69,25 @@ export type ScriptFunctionArgumentTypes =
   | FixedBytes;
 
 /**
+ * TypeArgument inputs for Entry functions, view functions, and scripts
+ *
+ * This can be a string version of the type argument such as:
+ * - u8
+ * - u16
+ * - u32
+ * - u64
+ * - u128
+ * - u256
+ * - bool
+ * - address
+ * - signer
+ * - vector<Type>
+ * - address::module::struct
+ * - address::module::struct<Type1, Type2>
+ */
+export type TypeArgument = TypeTag | string;
+
+/**
  * Type that holds all raw transaction instances Aptos SDK supports
  */
 export type AnyRawTransactionInstance = RawTransaction | MultiAgentRawTransaction | FeePayerRawTransaction;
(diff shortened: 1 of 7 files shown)


src/transactions/types.ts
```diff
@@ -144,7 +144,7 @@ export type InputMultiSigDataWithRemoteABI = {
  */
 export type InputScriptData = {
   bytecode: HexInput;
-  typeArguments?: Array<TypeTag>;
+  typeArguments?: Array<TypeTag | string>;
```
0xmaayan, line 147: can we use a safer type here? like `${string}::${string}::${string}` and whatever format is accepted? `string` feels too weak

tests/move/transfer/sources/script_coin_transfer.move
```diff
@@ -0,0 +1,12 @@
+script {
```
0xmaayan, line 1: can we reuse https://github.com/aptos-labs/aptos-ts-sdk/blob/main/tests/move/transfer/sources/script_transfer.move instead of creating another move script test file?

0xmaayan (commented):
please update changelog

tests/move/transfer/sources/script_coin_transfer.move
gregnazario: We can, though I wanted to ensure there was an example with and without a type argument

src/transactions/types.ts
gregnazario: It will be converted to a TypeTag, so if it's not formatted correctly it will be rejected.

Oh, and we can't use that type because `u8` is a valid TypeTag

0xmaayan (approved):

tests/e2e/transaction/transactionBuilder.test.ts
```diff
@@ -155,6 +155,52 @@ describe("transaction builder", () => {
       expect(rawTxn.payload instanceof TransactionPayloadScript).toBeTruthy();
     });
 
+    test("it generates a raw transaction with script payload and string type arguments", async () => {
+      const alice = Account.generate();
+      await aptos.fundAccount({ accountAddress: alice.accountAddress, amount: FUND_AMOUNT });
+      const payload = await generateTransactionPayload({
+        bytecode: TYPED_SCRIPT_TEST,
+        typeArguments: ["0x1::aptos_coin::AptosCoin"],
+        functionArguments: [
```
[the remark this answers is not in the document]
gregnazario: yes... that's interesting that it's still working?  strange...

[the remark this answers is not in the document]
gregnazario: Ah, it doesn't submit it to on-chain, just checks for building

PR #401, opened 2024-05-20
Demonstrate integration tests with the SDK

0xmaayan:
### Description
An example demonstrate how one can write integration tests with the TS SDK against a local or live node.

Notes:
1. The example uses the sdk `move.publish` function that uses `aptos move publish`. This method requires to provide a `privateKey` and a `url` so the CLI would know how and where to publish the package with. There is an alternative flow using `aptos move build-publish-payload --json-output-file` - an example can be found [here](https://github.com/aptos-labs/aptos-ts-sdk/blob/main/examples/typescript/your_fungible_asset.ts#L152-L161)
2. Currently, the sdk CLI functions are not generic and doesnt provide the user a way to pass in additional arguments (such as --assume-yes, --private-key, etc), we should modify the functions so users ideally can provide any arguments supported by the CLI


0xmaayan:
created for demo reasons

PR #406, opened 2024-05-22
Support generating transactions with loose types for SDK V1 backward …

0xmaayan:
…compatibility

Previously, SDK V1 (i.e `aptos`) let you pass in `Integer` and `Boolean` types as `string`. We removed this support in SDK v2 to have a strong type system in the SDK.

This causes dapps to break as they migrate from SDK V1 to SDK V2 - so adding this support back. 

Reference to SDK V1 support https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/transaction_builder/builder_utils.ts#L39-L72


gregnazario (approved):

hardsetting (approved):
Nice!

PR #409, opened 2024-05-24
[block] Change block APIs to fetch all transactions in the block

gregnazario:
### Description
Additionally, this breakingly changes the block APIs to be under transaction rather than general, due to a circular dependency.

### Test Plan
I tested this live against Mainnet, which has over 100 transactions per block.  This helped me work out some of the edge conditions of off by 1.

### DIFF
--- a/src/internal/transaction.ts
+++ b/src/internal/transaction.ts
@@ -19,6 +19,7 @@ import {
   type TransactionResponse,
   WaitForTransactionOptions,
   CommittedTransactionResponse,
+  Block,
 } from "../types";
 import { DEFAULT_TXN_TIMEOUT_SEC, ProcessorType } from "../utils/const";
 import { sleep } from "../utils/helpers";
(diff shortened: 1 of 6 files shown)


src/internal/transaction.ts
```diff
@@ -279,3 +280,94 @@ export class FailedTransactionError extends Error {
...
+      return block;
+    }
+
+    // For now, we will grab all the transactions in groups of 100, but we can make this more efficient by trying larger
+    // amounts
+    const fetchFutures = [];
+    const pageSize = 100n;
+    for (let i = latestVersion + 1n; i < lastVersion; i += BigInt(100)) {
+      fetchFutures.push(
+        getTransactions({
```
0xmaayan, line 355: I'd prefer this change to come from the API, there is always the chance to hit "too many requests" error

src/internal/transaction.ts
```diff
@@ -279,3 +280,94 @@ export class FailedTransactionError extends Error {
...
+  if (options?.withTransactions) {
+    // Transactions should be filled, but this ensures it
+    block.transactions = block.transactions ?? [];
+
+    const lastTxn = block.transactions[block.transactions.length - 1];
+    const firstVersion = BigInt(block.first_version);
+    const lastVersion = BigInt(block.last_version);
+
+    // Convert the transaction to the type
+    const curVersion: string | undefined = (lastTxn as any)?.version;
```
0xmaayan, line 334: could we not use `any`?

src/internal/transaction.ts
```diff
@@ -279,3 +280,94 @@ export class FailedTransactionError extends Error {
...
+    let latestVersion;
```
0xmaayan, line 335:
```suggestion
    let latestVersion = lastTxn.version ? BigInt(lastTxn.version) : firstVersion - 1n;
```

src/internal/transaction.ts
```diff
@@ -279,3 +280,94 @@ export class FailedTransactionError extends Error {
...
+  options?: { withTransactions?: boolean };
+}): Promise<Block> {
+  const { aptosConfig, blockHeight, options } = args;
+  const { data: block } = await getAptosFullNode<{}, Block>({
+    aptosConfig,
+    originMethod: "getBlockByHeight",
+    path: `blocks/by_height/${blockHeight}`,
+    params: { with_transactions: options?.withTransactions },
+  });
+  return fillBlockTransactions({ block, ...args });
```
0xmaayan, line 312: this api endpoint doesnt support `cursor`? so we could do `paginateWithCursor` instead

0xmaayan (approved):
stamp to unblock. but a better result would be to implement it on the API side

src/internal/transaction.ts
gregnazario: Yep, it's a mix of both.  Those using the block APIs are really only making indexers, so rate limiting should be obvious. 

API change is in as well

PR #415, opened 2024-06-03
Add TypeTagString

BriungRi:
### Description
`0x1::string::String` is pretty commonly used, so I think it would be nice to provide an export that calls into `TypeTagStruct`

### DIFF
--- a/src/transactions/typeTag/index.ts
+++ b/src/transactions/typeTag/index.ts
@@ -340,6 +340,12 @@ export class TypeTagStruct extends TypeTag {
   }
 }
 
+export class TypeTagString extends TypeTagStruct {
+  constructor() {
+    super(new StructTag(AccountAddress.ONE, new Identifier("string"), new Identifier("String"), []));
+  }
+}
+
 export class StructTag extends Serializable {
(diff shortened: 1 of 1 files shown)


src/transactions/typeTag/index.ts
```diff
@@ -340,6 +340,12 @@ export class TypeTagStruct extends TypeTag {
   }
 }
 
+export class TypeTagString extends TypeTagStruct {
+  constructor() {
+    super(new StructTag(AccountAddress.ONE, new Identifier("string"), new Identifier("String"), []));
+  }
+}
```
gregnazario, line 347: There's a type for this?  `stringTypeTag()`

It's not an actual TypeTag to be a string, so... it'll get real confusing really quick to have types that don't match the underlying data.  Not to mention we already have fragmentation between SDKs

gregnazario: Each `TypeTag` type represents an actual underlying type in the rust enum `TypeTag` used for the MoveVM.  Each one has a different encoding associated.

String is not a native type in Move, it's a `Struct` type tag, with specific properties.

PR #418, opened 2024-06-03
Removes use of `is_active` for ANS queries

kaw2k:
### Description
[An issue was raised for the ANS api](https://github.com/aptos-labs/aptos-names/pull/313). This PR brings the logic from that fix the the SDK. 

### Test Plan
See the test plan for the linked PR.

### DIFF
--- a/src/internal/ans.ts
+++ b/src/internal/ans.ts
@@ -68,6 +68,40 @@ export function isValidANSName(name: string): { domainName: string; subdomainNam
   };
 }
 
+export enum SubdomainExpirationPolicy {
+  Independent = 0,
+  FollowsDomain = 1,
+}
+
+/**
+ * A helper function to determine if a given ANS name is considered active or
+ * not. Domains are considered active if their expiration date is in the
+ * future. Subdomains have two policies which modify their behavior. They can
+ * follow their parent's expiration (1) in which they ignore their own
+ * expiration timestamp or they can expire independently (0) in which they can
+ * expire before their parent but not afterwards.
+ *
+ * @param name - An ANS name returned from one of the functions of the SDK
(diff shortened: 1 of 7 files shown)


src/internal/ans.ts
```diff
@@ -68,6 +68,28 @@ export function isValidANSName(name: string): { domainName: string; subdomainNam
   };
 }
 
+enum SubdomainExpirationPolicy {
+  Independent = 0,
+  FollowsDomain = 1,
+}
+export function isActiveANSName(name: GetANSNameResponse[0]): boolean {
```
0xmaayan, line 75: can we get a test for this function?

src/internal/ans.ts
```diff
@@ -68,6 +68,28 @@ export function isValidANSName(name: string): { domainName: string; subdomainNam
   };
 }
 
+enum SubdomainExpirationPolicy {
```
0xmaayan, line 71:
```suggestion
export enum SubdomainExpirationPolicy {
```

src/internal/ans.ts
0xmaayan, line 75: lets add doc string comments for this function and what it is for

src/internal/ans.ts
```diff
@@ -365,7 +386,7 @@ export async function getName(args: {
     res = sanitizeANSName(res);
   }
 
-  return res;
+  return isActiveANSName(res) ? res : undefined;
```
0xmaayan, line 389: dow e need to update this function test accordingly?

banool: We deploy ANS to the localnet as part of the E2E test suite, maybe now is the time to switch the test to use that.

0xmaayan (approved):
LGTM

PR #419, opened 2024-06-04
Fix keyless transaction simulation

heliuchuan:
### Description

This sends a 'null' signature of Keyless type instead of Ed25519


0xmaayan (approved):
please update changelog

PR #431, opened 2024-06-13
[abi] Allow overriding View function ABI with a local version

gregnazario:
### Description
This doesn't allow skipping the ABI, as it still checks inputted types are the proper types, but it should allow for things like private view functions, and skipping the ABI pull, with the type safety of the ABI.

This ABI check was originally added, because of the move to BCS view functions.  This was the only way to allow for more complex arguments to be parsed, without a significantly complex parser on both the fullnode side and the SDK side

The `undefined` checks tend not to do well, when there are multiple systems that parse from one to the next.  Quite often some pieces will return `null` instead of `undefined`, and we will likely need to do some extra handling of that elsewhere too.

### DIFF
--- a/src/transactions/transactionBuilder/transactionBuilder.ts
+++ b/src/transactions/transactionBuilder/transactionBuilder.ts
@@ -109,7 +109,7 @@ export async function generateTransactionPayload(
   }
   const { moduleAddress, moduleName, functionName } = getFunctionParts(args.function);
 
-  const functionAbi = await fetchAbi({
+  const functionAbi = await fetchOrUseProvidedAbi({
     key: "entry-function",
     moduleAddress,
     moduleName,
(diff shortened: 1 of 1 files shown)


src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -188,7 +188,7 @@ export async function generateViewFunctionPayload(args: InputViewFunctionDataWit
   });
 
   // Fill in the ABI
-  return generateViewFunctionPayloadWithABI({ abi: functionAbi, ...args });
+  return generateViewFunctionPayloadWithABI({ ...args, abi: functionAbi });
```
0xmaayan, line 191: why are we doing it that way if view functions can take in json arguments? when there is no abi but json arguments we should not do any server call

gregnazario: View functions take in BCS arguments?

The JSON parser on the server side isn't actually well featured enough to support most inputs.

If we really want to support the JSON inputs, we can, but it's the same problem as why transactions are encoded in BCS rather than JSON.

0xmaayan: https://aptos-org.slack.com/archives/C05NLAKJM9U/p1718147823459949?thread_ts=1718144833.994919&channel=C05NLAKJM9U&message_ts=1718147823.459949

PR #432, opened 2024-06-14
[api] Add viewJson as a way to migrate from the legacy SDK

gregnazario:
### Description
This keeps the original behavior from the Legacy SDK but as a different function on the SDK.

### Test Plan
See unit tests

### DIFF
--- a/src/api/general.ts
+++ b/src/api/general.ts
@@ -10,7 +10,7 @@ import {
   queryIndexer,
 } from "../internal/general";
 import { getBlockByHeight, getBlockByVersion } from "../internal/transaction";
-import { view } from "../internal/view";
+import { view, viewJson } from "../internal/view";
 import {
   AnyNumber,
   Block,
@@ -22,7 +22,7 @@ import {
   MoveValue,
 } from "../types";
 import { ProcessorType } from "../utils/const";
-import { InputViewFunctionData } from "../transactions";
+import { InputViewFunctionData, InputViewFunctionJsonData } from "../transactions";
 
 /**
  * A class to query all `General` Aptos related queries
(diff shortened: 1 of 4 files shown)


src/internal/view.ts
```diff
@@ -33,3 +38,24 @@ export async function view<T extends Array<MoveValue> = Array<MoveValue>>(args:
...
+
+export async function viewJson<T extends Array<MoveValue> = Array<MoveValue>>(args: {
+  aptosConfig: AptosConfig;
+  payload: InputViewFunctionJsonData;
+  options?: LedgerVersionArg;
+}): Promise<T> {
+  const { aptosConfig, payload, options } = args;
+  const { data } = await postAptosFullNode<ViewFunctionJsonPayload, MoveValue[]>({
+    aptosConfig,
+    originMethod: "view",
```
0xmaayan, line 50:
```suggestion
    originMethod: "viewJson",
```

src/api/general.ts
```diff
@@ -132,6 +132,29 @@ export class General {
...
+   * Queries for a Move view function with JSON, this provides compatability with the old `aptos` package
+   * @param args.payload Payload for the view function
+   * @param args.options.ledgerVersion The ledger version to query, if not provided it will get the latest version
+   *
+   * @example
+   * const data = await aptos.view({
+   *  payload: {
+   *   function: "0x1::coin::balance",
+   *   typeArguments: ["0x1::aptos_coin::AptosCoin"],
+   *   functionArguments: [accountAddress],
```
0xmaayan, line 145: `accountAddress` is of type `AccountAddress` - you would need to do `accountAddress.toString()`

0xmaayan (approved):

PR #441, opened 2024-06-20
[examples] Add rotate key example back

gregnazario:
### Description
This example used to be in the old SDK, but disappeared from the new one.

### Test Plan
CI around the example, it seems to be failing on indexer for me locally


0xmaayan:
> ### Description
> This example used to be in the old SDK, but disappeared from the new one.
> 
> ### Test Plan
> CI around the example, it seems to be failing on indexer for me locally
> 
> ### Related Links

hmm its here https://github.com/aptos-labs/aptos-ts-sdk/blob/main/examples/typescript/rotate_key.ts

gregnazario:
> > ### Description
> > This example used to be in the old SDK, but disappeared from the new one.
> > ### Test Plan
> > CI around the example, it seems to be failing on indexer for me locally
> > ### Related Links
> 
> hmm its here https://github.com/aptos-labs/aptos-ts-sdk/blob/main/examples/typescript/rotate_key.ts

I missed it, esm, not esm fun fun

PR #443, opened 2024-06-21
Add move script, object package deployment to cli

0xaptosj:
### Description
Add these functions to cli
- Run move script
- Create a new object and publish package to it
- Upgrade package published to object

### Test Plan
Tested MoonCoin example on testnet.

AND tested on windows and it works fine!


src/cli/move.ts
```diff
@@ -72,7 +72,7 @@ export class Move {
   }
 
   /**
-   * Function to publishe the modules in a Move package to the Aptos blockchain
+   * Function to publish the modules to the deployer account on the Aptos blockchain
```
0xaptosj, line 75:
```suggestion
   * Function to publish the modules to the publisher account on the Aptos blockchain
```

examples/typescript/local_node.ts
```diff
@@ -0,0 +1,113 @@
...
+      MoonCoin: "0x123",
+    },
+  });
+}
+
+// create a new object and publish a Move package to it
+async function createObjectAndPublishPackage() {
+  const move = new cli.Move();
+
+  await move.createObjectAndPublishPackage({
```
0xmaayan, line 68: what is the output we are getting from this command? is this the object address?

0xaptosj: yes, in this format

```sh
➜  sh_scripts ./deploy_nft_template.sh
##### Deploy module under a new object #####
Compiling, may take a little while to download git dependencies...
UPDATING GIT DEPENDENCY https://github.com/aptos-labs/aptos-core.git
UPDATING GIT DEPENDENCY https://github.com/aptos-labs/aptos-core.git
UPDATING GIT DEPENDENCY https://github.com/aptos-labs/token-minter.git
UPDATING GIT DEPENDENCY https://github.com/aptos-labs/aptos-core.git
UPDATING GIT DEPENDENCY https://github.com/aptos-labs/aptos-core.git
UPDATING GIT DEPENDENCY https://github.com/aptos-labs/aptos-core.git
INCLUDING DEPENDENCY AptosFramework
INCLUDING DEPENDENCY AptosStdlib
INCLUDING DEPENDENCY AptosToken
INCLUDING DEPENDENCY AptosTokenObjects
INCLUDING DEPENDENCY MoveStdlib
INCLUDING DEPENDENCY TokenMinter
BUILDING digital_asset_launchpad
package size 12447 bytes
Code was successfully deployed to object address 0x2608d20d0712a847296ccf4a768e63df351dc53153fba43458ad166048a8a1da.
{
  "Result": "Success"
}
```

0xaptosj: I used to have [this shell](https://github.com/aptos-labs/move-by-examples/blob/main/fungible-asset-launchpad/aptos/move/sh_scripts/deploy.sh#L13-L22) script to extract it and save it to txt, will replicate it in cad

0xmaayan: could we return the object address? and then, in a project (cad for example), use this output to run upgradeObjectPackage command?

src/cli/move.ts
```diff
@@ -106,6 +106,114 @@ export class Move {
...
+    return this.runCommand(cliArgs);
+  }
+
+  /**
+   * Function to upgrade a Move package previously published to an object on the Aptos blockchain
+   * Caller must be the object owner to call this function
+   *
+   * @param args.packageDirectoryPath Path to a move package (the folder with a Move.toml file)
+   * @param args.objectAddress Address of the object that the Move package published to
+   * @example
```
0xmaayan, line 156: example is wrong

src/cli/move.ts
```diff
@@ -106,6 +106,114 @@ export class Move {
     return this.runCommand(cliArgs);
   }
 
+  /**
+   * Function to create a new object and publish the Move package to it on the Aptos blockchain
+   *
+   * @param args.packageDirectoryPath Path to a move package (the folder with a Move.toml file)
+   * @param args.addressName Address name for the Move package
+   * @example
```
0xmaayan, line 114: example is wrong

0xaptosj: this indeed should be the name instead of value, here's the shell script version

```sh
aptos move create-object-and-publish-package \
  --package-dir $MOVE_DIR \
  --address-name launchpad_addr \
  --named-addresses "launchpad_addr=$PUBLISHER_ADDR,initial_creator_addr=0x7fe2b067b33642d6e5fea417452abeb131bb80ea07ad5d2ff0a0f58292238393"\
  --profile $PUBLISHER_PROFILE \
  --assume-yes

```

src/cli/move.ts
0xaptosj: fixed

src/cli/move.ts
0xmaayan: i see. a bit confusing as it is now. can we get 1 example that shows what arguments to pass?
```
{
 packageDirectoryPath:"some/path/to/move.mv;
 addressName: "mode_address_name";
 namedAddresses: {
     alice:0x1234, bob:0x5678
    }
}
```

0xmaayan (approved):
lgtm. just have the doc string clearer. 

also, please update changelog

src/cli/move.ts
0xaptosj: included a complete example

examples/typescript/local_node.ts
0xmaayan: If this is the recommended way to publish a move module - why not implement it in the publish function? so that would be the default way

0xaptosj: my reason is to keep the ts function name consistent to cli name.

0xmaayan: I understand. Tho I think it is an option for us to "update" the cli API and make it more consistent with the standard and expectations we have from our dev users and provide a better dev exp. Also, on CAD we still want to use `npm run move:publish` to publish the module instead  of using a confusing name like `npm run move:createObjectAndPublishPackage` so names will still be different

PR #448, opened 2024-06-26
Add a publish command

0xmaayan:
### Description
This PR adds a `publish` command to run with `pnpm run publish` to make sure we first build the package and then publish it to NPM

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
   "scripts": {
     "build:clean": "rm -rf dist",
     "build": "pnpm build:clean && tsup",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -35,6 +35,7 @@
   "scripts": {
     "build:clean": "rm -rf dist",
     "build": "pnpm build:clean && tsup",
+    "publish": "pnpm run build && npm publish",
```
gregnazario, line 38: Probably add a dry run publish command too.  Good to use these so we never use npm, only pnpm

gregnazario (approved):

PR #449, opened 2024-06-26
Fix publish scripts

0xmaayan:
### Description
Weird endless loop bug with https://github.com/aptos-labs/aptos-ts-sdk/pull/448 when running `npm publish --dry-run`, fixing it to only run the built-in `prepublishOnly` npm script to build the SDK before running `npm publish --dry-run` or `npm publish`. Updated the CONTRIBUTING guide also.

Will look at using `pnpm` only later

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   "scripts": {
     "build:clean": "rm -rf dist",
     "build": "pnpm build:clean && tsup",
-    "publish": "pnpm run build && npm publish",
+    "prepublishOnly": "pnpm run build",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -35,7 +35,7 @@
   "scripts": {
     "build:clean": "rm -rf dist",
     "build": "pnpm build:clean && tsup",
-    "publish": "pnpm run build && npm publish",
+    "prepublishOnly": "pnpm run build",
```
gregnazario, line 38: But this is just build then?  So, does it change anything from the instructions?

gregnazario: Oh I see it autoruns.

gregnazario (approved):

PR #451, opened 2024-06-26
add getprice to estimate irys upload cost

jillxuu:
### Description
add getprice to estimate irys upload cost

### Related Links
https://docs.irys.xyz/developer-docs/irys-sdk/api/getPrice

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,6 @@
   "devDependencies": {
     "@aptos-labs/aptos-cli": "^0.1.2",
     "@babel/traverse": "^7.23.6",
-    "bignumber.js": "^9.1.2",
     "@graphql-codegen/cli": "^5.0.0",
     "@graphql-codegen/import-types-preset": "^3.0.0",
     "@graphql-codegen/typescript": "^4.0.1",
--- a/package.json
+++ b/package.json
@@ -65,11 +65,13 @@
   "devDependencies": {
     "@aptos-labs/aptos-cli": "^0.1.2",
     "@babel/traverse": "^7.23.6",
+    "bignumber.js": "^9.1.2",
     "@graphql-codegen/cli": "^5.0.0",
     "@graphql-codegen/import-types-preset": "^3.0.0",
     "@graphql-codegen/typescript": "^4.0.1",
     "@graphql-codegen/typescript-graphql-request": "^6.0.1",
     "@graphql-codegen/typescript-operations": "^4.0.1",
+    "@irys/sdk": "0.1.15",
     "@types/jest": "^29.5.11",
     "@types/node": "^20.10.4",
     "@typescript-eslint/eslint-plugin": "^6.14.0",
@@ -90,8 +92,7 @@
     "ts-node": "^10.9.2",
     "tsup": "^8.0.1",
     "typedoc": "^0.25.4",
-    "typescript": "5.3.3",
-    "@irys/sdk": "0.1.15"
+    "typescript": "5.3.3"
   },
   "version": "1.9.1-tmp.0"
 }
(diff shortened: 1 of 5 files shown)


src/plugins/assetUploader/index.ts
```diff
@@ -80,4 +81,16 @@ export class AssetUploader {
...
+
+  /**
+   * Get cost estimate
+   *
+   * @param args.numBytes  The number of bytes to check the price for
+   *
+   * @return Cost to upload numBytes, unit is the token specified
+   * when instantiating the Irys object Return value is in atomic units
+   */
+  async getPrice(args: { numBytes: number }): Promise<BigNumber> {
```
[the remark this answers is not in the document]
jillxuu: do they have `estimateFolderPrice` for node irys tho? didn't see sdk reference in the doc

src/plugins/assetUploader/irys.ts
```diff
@@ -195,4 +196,24 @@ export class IrysAssetUploader implements IAssetUploader {
...
+   *
+   * @return Cost to upload numBytes, unit is the token specified
+   * when instantiating the Irys object Return value is in atomic units
+   */
+  async getPrice(args: { account: Account; numBytes: number }): Promise<BigNumber> {
+    const { account, numBytes } = args;
+    const irys = await this.getIrys({ account });
+
+    try {
+      return await irys.getPrice(numBytes);
```
0xmaayan, line 214: I would just return a `number` here and remove the use of `BigNumber`

```suggestion
      return await irys.getPrice(numBytes).toNumber;
```

src/plugins/assetUploader/index.ts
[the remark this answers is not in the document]
0xmaayan: I think utils is shared between web and node

package.json
```diff
@@ -65,11 +65,13 @@
   "devDependencies": {
     "@aptos-labs/aptos-cli": "^0.1.2",
     "@babel/traverse": "^7.23.6",
+    "bignumber.js": "^9.1.2",
     "@graphql-codegen/cli": "^5.0.0",
     "@graphql-codegen/import-types-preset": "^3.0.0",
     "@graphql-codegen/typescript": "^4.0.1",
     "@graphql-codegen/typescript-graphql-request": "^6.0.1",
     "@graphql-codegen/typescript-operations": "^4.0.1",
+    "@irys/sdk": "0.1.15",
```
0xmaayan, line 74: Irys is already at `0.2.1` - might worth to update?

jillxuu: in the past ran into version mismatch issue. wanna get the functionality works first and then do version update in a separate PR.

0xmaayan (approved):

0xaptosj:
not related to this PR, but when we calculate upload cost, how do we handle the size change after we update the media url in  metadata files? If the media url is blank in metadata file at the beginning, after the media upload and it's filled, the size of metadata file will grow a bit.

0xmaayan:
> not related to this PR, but when we calculate upload cost, how do we handle the size change after we update the media url in metadata files? If the media url is blank in metadata file at the beginning, after the media upload and it's filled, the size of metadata file will grow a bit.
> 
> Maybe a potential solution is we calculate how many media url there would be, e.g. for a collection of 1000 NFTs, there would be 1009+1 media urls need to be updated. And Irys URL should have fixed length, we can add the total file size by 1001 * bytes_of(Irys URL)

do you mean updating a folder after a folder has been uploaded to irys? or it the change in cost with vs without the media url?

jillxuu:
> not related to this PR, but when we calculate upload cost, how do we handle the size change after we update the media url in metadata files? If the media url is blank in metadata file at the beginning, after the media upload and it's filled, the size of metadata file will grow a bit.
> 
> Maybe a potential solution is we calculate how many media url there would be, e.g. for a collection of 1000 NFTs, there would be 1009+1 media urls need to be updated. And Irys URL should have fixed length, we can add the total file size by 1001 * bytes_of(Irys URL)

this shouldn't be a problem from beginning because the cost should be calculated before the submission where the json file is already updated?

PR #454, opened 2024-07-02
Implement a base version of verifySignature for KeylessAccounts

heliuchuan:
### Description

This adds very basic signature verification in order to enable sign and verify for the wallet adapter.  This implementation is incomplete as it only checks the ephemeral signature.  

Full verification of the proof and training wheels will be done later as it requires querying the chain and the right bn254 library to deserialize the g1/g2 points of the VK.

### Test Plan

Added e2e test

### DIFF
--- a/src/account/KeylessAccount.ts
+++ b/src/account/KeylessAccount.ts
@@ -268,10 +268,26 @@ export class KeylessAccount extends Serializable implements Account {
     const signMess = txnAndProof.hash();
     return this.sign(signMess);
   }
-
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars, class-methods-use-this
-  verifySignature(args: { message: HexInput; signature: Signature }): boolean {
-    throw new Error("Not implemented");
+  /**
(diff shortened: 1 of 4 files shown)


src/account/KeylessAccount.ts
```diff
@@ -269,9 +268,26 @@ export class KeylessAccount extends Serializable implements Account {
     return this.sign(signMess);
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars, class-methods-use-this
-  verifySignature(args: { message: HexInput; signature: Signature }): boolean {
-    throw new Error("Not implemented");
+  /**
+   * Note - This function is currently incomplete and should only be used to verify ownership of the KeylessAccount
```
0xmaayan, line 272: what is missing to consider it as complete?

0xmaayan (approved):

src/account/KeylessAccount.ts
heliuchuan: Proof verification

PR #455, opened 2024-07-02
[fungibleAssets] Ensure query for coin balances uses correct query

gregnazario:
### Description
For migrated coins, they need to query both the address and the coin type associated.

### DIFF
--- a/src/internal/account.ts
+++ b/src/internal/account.ts
@@ -52,9 +52,10 @@ import {
   GetAccountTransactionsCount,
 } from "../types/generated/queries";
 import { memoizeAsync } from "../utils/memoize";
-import { Secp256k1PrivateKey, AuthenticationKey, Ed25519PrivateKey } from "../core";
+import { Secp256k1PrivateKey, AuthenticationKey, Ed25519PrivateKey, createObjectAddress } from "../core";
 import { CurrentFungibleAssetBalancesBoolExp } from "../types/generated/types";
 import { getTableItem } from "./table";
+import { APTOS_COIN } from "../utils";
 
 export async function getInfo(args: {
   aptosConfig: AptosConfig;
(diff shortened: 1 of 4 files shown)


src/internal/account.ts
```diff
@@ -404,16 +405,39 @@ export async function getAccountTransactionsCount(args: {
...
+    faAddress = AccountAddress.from(faMetadataAddress).toStringLong();
+  } else if (coinType !== undefined && faMetadataAddress === undefined) {
+    // TODO Move to a separate function as defined in the AIP for coin migration
+    if (args.coinType === APTOS_COIN) {
+      faAddress = AccountAddress.A.toStringLong();
+    } else {
+      faAddress = createObjectAddress(AccountAddress.A, coinType).toStringLong();
+    }
+  } else if (coinType === undefined && faMetadataAddress !== undefined) {
+    faAddress = AccountAddress.from(faMetadataAddress).toStringLong();
```
rtso, line 424: If faMetadataAddress is set, should this convert to the coinType?

src/internal/account.ts
```diff
@@ -404,16 +405,39 @@ export async function getAccountTransactionsCount(args: {
...
+  let faAddress: string;
+
+  if (coinType !== undefined && faMetadataAddress !== undefined) {
+    faAddress = AccountAddress.from(faMetadataAddress).toStringLong();
+  } else if (coinType !== undefined && faMetadataAddress === undefined) {
+    // TODO Move to a separate function as defined in the AIP for coin migration
+    if (args.coinType === APTOS_COIN) {
+      faAddress = AccountAddress.A.toStringLong();
+    } else {
+      faAddress = createObjectAddress(AccountAddress.A, coinType).toStringLong();
```
rtso, line 421: @lightmark to check that this function is correct

src/internal/account.ts
gregnazario: There's no way to retrieve the coin type from the FA metadata address.

So, in this case I'm taking it as an FA by itself.

src/internal/account.ts
gregnazario: I took this from the code directly here, which should be extracted to a view function:

https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/coin.move#L320-L328

src/internal/account.ts
lightmark: This is the expected behavior.

src/internal/account.ts
lightmark: @rtso what is the `asset_type` is not changed from `0x1::aptos_coin::AptosCoin` to `0xa` yet?
From the client perspective, they don't know what to use. Although may be a short term issue.

lightmark (commented):
lgtm except the question for @rtso

lightmark (approved):

PR #458, opened 2024-07-08
ensure that errors have a meaningful message

geekflyer:
This improves the error message handling in the SDK. Previously we would receive `AptosApiError: [object Object]` and other non-descriptive errors in some cases. This came up in debugging the keyless e2e tests specifically.

### DIFF
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { AptosRequest } from "../types";
+import { AptosApiType } from "../utils/const.js";
 
 /**
  * The API response type
@@ -24,6 +25,12 @@ export interface AptosResponse<Req, Res> {
   request?: Req;
 }
 
+type AptosApiErrorOpts = {
+  apiType: AptosApiType;
+  aptosRequest: AptosRequest;
+  aptosResponse: AptosResponse<any, any>;
+};
+
 /**
  * The type returned from an API error
  *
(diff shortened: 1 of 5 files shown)


src/client/types.ts
```diff
@@ -45,14 +52,59 @@ export class AptosApiError extends Error {
...
+    aptosResponse.status
+  }) and response body: ${serializeAnyPayloadForErrorMessage(aptosResponse.data)}`;
+}
+
+const SERIALIZED_PAYLOAD_TRIM_TO_MAX_LENGTH = 400;
+
+// this function accepts a payload of any type (probably an object) and serializes it to a string
+// Since we don't know the type or size of the payload and we don't want to add a huge object in full to the error message
+// we limit the to the first 200 and last 200 characters of the serialized payload and put a "..." in the middle.
+function serializeAnyPayloadForErrorMessage(payload: any): string {
```
0xmaayan, line 101:
```suggestion
export function serializeAnyPayloadForErrorMessage(payload: any): string {
```

tests/unit/__snapshots__/client.test.ts.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
0xmaayan, line 1: whats this file? lets not include it? it looks like a complete different test mechanism than what we have in the SDK, might be good to revisit using it in the future.

tests/unit/client.test.ts
```diff
@@ -0,0 +1,102 @@
+import { AptosApiType } from "../../src/utils/const.js";
```
0xmaayan, line 1: lets have these tests part of this file https://github.com/aptos-labs/aptos-ts-sdk/blob/main/tests/e2e/client/aptosRequest.test.ts and a test for each AptosApiType

src/client/types.ts
```diff
@@ -45,14 +52,59 @@ export class AptosApiError extends Error {
...
+
+  // This is the generic/catch-all case. We received some response from the API but it doesn't appear to be a well-known structure.
+  // We print http status codes and the response body (after some trimming),
+  // in the hope that this gives enough context what went wrong without printing overly huge messages.
+  return `${errorPrelude} status: ${aptosResponse.statusText}(code:${
+    aptosResponse.status
+  }) and response body: ${serializeAnyPayloadForErrorMessage(aptosResponse.data)}`;
+}
+
+const SERIALIZED_PAYLOAD_TRIM_TO_MAX_LENGTH = 400;
```
0xmaayan, line 96:
```suggestion
export const SERIALIZED_PAYLOAD_TRIM_TO_MAX_LENGTH = 400;
```

src/client/types.ts
```diff
@@ -45,14 +52,59 @@ export class AptosApiError extends Error {
...
+  // Received well-known structured error response body - simply serialize and return it.
+  // We don't need http status codes etc. in this case.
+  if (aptosResponse.data?.message != null && aptosResponse.data?.error_code != null) {
+    return `${errorPrelude}: ${JSON.stringify(aptosResponse.data)}`;
+  }
+
+  // This is the generic/catch-all case. We received some response from the API but it doesn't appear to be a well-known structure.
+  // We print http status codes and the response body (after some trimming),
+  // in the hope that this gives enough context what went wrong without printing overly huge messages.
+  return `${errorPrelude} status: ${aptosResponse.statusText}(code:${
```
0xmaayan, line 91: verifying - the sdk supports request to non aptos api types (i.e custom endpoints) - does this handle those cases?

0xmaayan (commented):
Thanks! thats sweet! left some comments

src/client/types.ts
```diff
@@ -24,6 +25,12 @@ export interface AptosResponse<Req, Res> {
   request?: Req;
 }
 
+type AptosApiErrorOpts = {
```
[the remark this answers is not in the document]
geekflyer: the constructor args of AptosApiError are an internal interface/implementation detail of the SDK, hence we don't need/want to export this type

src/client/types.ts
```diff
@@ -45,14 +52,59 @@ export class AptosApiError extends Error {
 
   readonly request: AptosRequest;
 
-  constructor(request: AptosRequest, response: AptosResponse<any, any>, message: string) {
-    super(message);
+  constructor({ apiType, aptosRequest, aptosResponse }: AptosApiErrorOpts) {
```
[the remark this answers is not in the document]
geekflyer: this is not a breaking change since the constructor of AptosApiError is an SDK internal API detail.

src/client/types.ts
```diff
@@ -45,14 +52,59 @@ export class AptosApiError extends Error {
...
-    this.request = request;
+    this.url = aptosResponse.url;
+    this.status = aptosResponse.status;
+    this.statusText = aptosResponse.statusText;
+    this.data = aptosResponse.data;
+    this.request = aptosRequest;
+  }
+}
+
+function deriveErrorMessage({ apiType, aptosRequest, aptosResponse }: AptosApiErrorOpts): string {
```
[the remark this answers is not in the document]
geekflyer: this is internal

src/client/types.ts
geekflyer: yes it does

src/client/types.ts
geekflyer: this is internal

src/client/types.ts
geekflyer: internal too

tests/unit/client.test.ts
geekflyer: this is a unit test and hence doesn't belong to aptosRequest.test.ts

src/client/types.ts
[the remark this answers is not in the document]
0xmaayan: the class is exported, and might be used by external projects, so the constructor might also be used, and we need to have everything it uses exported and mentioned as changed

tests/unit/__snapshots__/client.test.ts.snap
0xmaayan: toMatchSnapshot is mostly good to use when we want to test against a UI and that it doesnt change unexpectedly, which is were we mostly use it (doc website, aptos-dashboard)

src/client/types.ts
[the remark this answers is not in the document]
geekflyer: why would external projects instantiate this class? This sounds like a non-intended usage of the SDK. I marked the constructor as @internal as per https://github.com/microsoft/TypeScript/issues/58891#issuecomment-2203609250 to make this clear.

tests/unit/__snapshots__/client.test.ts.snap
geekflyer: I disagree. Snapshot based testing is useful for more than UI stuff. In fact we have various examples of snapshot based testing in our codebase that use it for backend code (see the search above), incl. with other libraries. We also use snapshot based testing in our rust codebase in aptos-core (we call this golden files there but it's the same concept).
Furthermore you can argue that the error messages are sort of the UI surface of the SDK since that's what users/devs will see as output when working with it.

src/client/types.ts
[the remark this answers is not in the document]
0xmaayan: We had issues in the past where SDK users use exported functions, even if meant to use as internal. Since it is not an expected use case, we dont need to go the deprecated warning road but just mention it on CHANGELOG that the constructor interface has changed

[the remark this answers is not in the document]
geekflyer: I still think it's a bit unnecessary and noisy to add changes to clearly internally meant signatures to the changelog. If someone does indeed indeed instantiate an AptosApiError outside of the SDK they are clearly doing something outside of the norm/advanced use-case and in those rare cases I would expect that they have a sufficient sophisticated setup and understanding of TypeScript and TypeScript would tell them that the signature has changed.

I'm happy to add this as a one-time change to the changelog for now but I think in future we should more clearly draw a line between the external API surface and internal type signatures as stabilizing and adding every single internal function signature change to a changelog sounds extremely burdensome as an SDK developer and leads to unnecessary noisy changelogs for the SDK consumer.

0xmaayan (approved):

PR #462, opened 2024-07-11
[core] Remove `instanceof` where input might come from other bundle

hardsetting:
### Description
The usage of `instanceof` keeps causing a lot of problem, specifically when using the TS sdk from different bundles.
Sometimes this is due to misconfiguration, some other times there's no way around it (e.g. injected wallet plugins).

This PR attempts to remove the usage of `instanceof` from all flows were inputs might come from different bundles:
- When using `Account.from` with an instance of `AccountAddress`
- When verifying a signature using an instance of `Signature`
- When simulating a transaction using an instance of `PublicKey`

### Test Plan
CI, as well as manually testing Wapal that is currently unable to integrate with AptosConnect because of the above problem.

### DIFF
--- a/src/core/crypto/ed25519.ts
+++ b/src/core/crypto/ed25519.ts
@@ -22,6 +22,26 @@ const L: number[] = [
 ];
 
 /**
+ * Checks if an ED25519 signature is non-canonical.
+ *
+ * Comes from Aptos Core
+ * https://github.com/aptos-labs/aptos-core/blob/main/crates/aptos-crypto/src/ed25519/ed25519_sigs.rs#L47-L85
+ */
+export function isCanonicalEd25519Signature(signature: Signature): boolean {
+  const s = signature.toUint8Array().slice(32);
+  for (let i = L.length - 1; i >= 0; i -= 1) {
+    if (s[i] < L[i]) {
+      return true;
+    }
+    if (s[i] > L[i]) {
+      return false;
+    }
+  }
+  // As this stage S == L which implies a non-canonical S.
+  return false;
+}
+
+/**
  * Represents the public key of an Ed25519 key pair.
  *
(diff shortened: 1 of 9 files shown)


src/core/crypto/singleKey.ts
```diff
@@ -156,6 +160,12 @@ export class AnySignature extends Signature {
   // region AccountSignature
 
   toUint8Array() {
+    // TODO: keep this warning around for a bit, and eventually change this to return `this.signature.toUint8Array()`.
+    // eslint-disable-next-line no-console
+    console.warn(
+      "Calls to AnySignature.toUint8Array() will soon return the underlying signature bytes. " +
```
0xmaayan, line 166: nit
```suggestion
      "Calls to Aptos TypeScript SDK AnySignature.toUint8Array() will soon return the underlying signature bytes. " +
```

src/core/crypto/ed25519.ts
```diff
@@ -118,6 +135,10 @@ export class Ed25519PublicKey extends AccountPublicKey {
   static isPublicKey(publicKey: AccountPublicKey): publicKey is Ed25519PublicKey {
     return publicKey instanceof Ed25519PublicKey;
   }
+
+  static isInstance(publicKey: PublicKey): publicKey is Ed25519PublicKey {
```
0xmaayan, line 139: could we add tests to all those new `isInstance` functions?

0xmaayan (commented):
LGTM. need tests for the new functions and please update CHANGELOG

src/core/crypto/singleKey.ts
hardsetting: added prefix

src/core/crypto/ed25519.ts
hardsetting: done

0xmaayan (approved):

PR #194, opened 2023-11-15
Implement signWithPasskey

heliuchuan:

PR #474, opened 2024-07-22
[passkey] initial draft

mizu-flex:
### Description
This submission adds basic support for passkey to the SDK, and also includes corresponding tests.

### Test Plan
There still some errors in api tests, especially the submission part of passkey transactions

### Related Links
This submission mainly referenced the `passkeySigning` branch.


babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: [["@babel/preset-env", { targets: { node: "current" } }]],
```
0xmaayan, line 2: what is the reason for adding babel? the sdk uses tsup as the bundle tool

examples/typescript/passkey/src/App.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+  /**
+   * Prints the balance of an account
+   * @param aptos
+   * @param name
+   * @param address
+   * @returns {Promise<*>}
+   *
+   */
+  const balance = async (aptos: Aptos, name: string, address: AccountAddress) => {
```
0xmaayan, line 31: could we keep this example in a node env - same as all the other examples on the SDK? https://github.com/aptos-labs/aptos-ts-sdk/blob/main/examples/typescript/keyless.ts

src/internal/passkey.ts
```diff
@@ -0,0 +1,174 @@
...
+  };
+
+  return _generateRegistrationOptions({
+    ...args,
+    authenticatorSelection,
+    supportedAlgorithmIDs,
+  });
+}
+
+export async function registerCredential(
```
hariria, line 63: In addition to the registerCredential, there needs to be more methods verifying whether or not the credential was actually backedup (parse the registration response, verify backup eligibility and backup state), otherwise people will shoot themselves in the foot and create non-backedup credentials. 

See https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-66.md#authenticator-data

src/api/passkey.ts
```diff
@@ -0,0 +1,91 @@
...
+    timeout?: number;
+    rpID: string;
+    options?: {
+      allowCredentials?: AllowCredentialOption[];
+    };
+  }): Promise<PendingTransactionResponse> {
+    return signAndSubmitWithPasskey({ aptosConfig: this.config, ...args });
+  }
+
+  async signWithPasskey(args: {
```
0xmaayan, line 51: please add doc strings for each function

src/internal/passkey.ts
hariria: Additionally we need support for identifying incompetent / malicious authenticators via AAGUID.  see https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-66.md#incompetent--malicious-authenticators

src/core/crypto/passkey.ts
```diff
@@ -0,0 +1,73 @@
...
+import { AuthenticationKey } from "../authenticationKey";
+import { AccountPublicKey, VerifySignatureArgs } from "./publicKey";
+import { Secp256k1PublicKey } from "./secp256k1";
+
+/**
+ * Represents the Passkey public key
+ *
+ * Passkey public keys are Secp256r1 ecdsa public keys.
+ */
+export class PasskeyPublicKey extends AccountPublicKey {
```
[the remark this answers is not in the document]
mizu-flex: That is for `getPasskeyAccountAddress` which need a type `AccountPublicKey`

babel.config.js
mizu-flex: It is for `babel-jest`

Some imports from `@simplewebauthn/browser` broke jest, and I couldn't find a way to successfully run it using only ts-jest, so I introduced bable-jest.

examples/typescript/passkey/src/App.tsx
mizu-flex: I will have a try, but it maybe hard. Passkey (webauthn) requires browser API for interactive authentication process.

babel.config.js
0xmaayan: I see. I'd prefer not to introduce another bundle tool and keep the existing one. Also, to reduce potential errors, and to not increase the package size.

mizu-flex: Agree, but I have spent two days, still cannot run jest with only `ts-jest`, looking forward to any suggestions

src/core/crypto/passkey.ts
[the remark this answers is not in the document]
hariria: Why can't we make `Secp256r1PublicKey extends AccountPublicKey`?

src/core/crypto/passkey.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  /**
+   * Create a new PublicKey instance from a Uint8Array or String.
+   *
+   * @param hexInput A HexInput (string or Uint8Array)
+   */
+  constructor(hexInput: HexInput) {
+    super();
+
+    const key = new Secp256k1PublicKey(hexInput);
```
hariria, line 24: Incorrect Public Key, but this is no longer relevant based on the comment above

src/core/crypto/passkey.ts
```diff
@@ -0,0 +1,73 @@
+import { Serializer } from "../../bcs";
+import { AnyPublicKeyVariant, SigningScheme as AuthenticationKeyScheme, HexInput } from "../../types";
+import { AuthenticationKey } from "../authenticationKey";
+import { AccountPublicKey, VerifySignatureArgs } from "./publicKey";
+import { Secp256k1PublicKey } from "./secp256k1";
```
hariria, line 5: Wrong signature scheme

hariria:
Please also include functions for ECDSA Public Key recovery: https://github.com/aptos-labs/aptos-ts-sdk/pull/194/files#diff-0d9fbf024d51030ac58aa3709e3eed1a07f4d45b5f65affd1211bff56dacb9b4R1.

babel.config.js
mizu-flex: BTW, the solution I found is from https://github.com/MasterKale/SimpleWebAuthn/issues/293

mizu-flex:
> Please also include functions for ECDSA Public Key recovery: https://github.com/aptos-labs/aptos-ts-sdk/pull/194/files#diff-0d9fbf024d51030ac58aa3709e3eed1a07f4d45b5f65affd1211bff56dacb9b4R1.

Added

src/internal/passkey.ts
mizu-flex: 1. Can I just expose this method `verifyAuthenticationResponse` to user? Its return type `VerifiedAuthenticationResponse` include `credentialDeviceType` and `credentialBackedUp`

2. Should we implement a policy that only allows the use of password managers on the list https://github.com/passkeydeveloper/passkey-authenticator-aaguids/blob/main/aaguid.json ?

examples/typescript/passkey/src/recoverPubKey.ts
```diff
@@ -0,0 +1,65 @@
...
+  signature: Uint8Array,
+) {
+  const shaClientDataJSON = await sha256(clientDataJSON);
+  const message = concatenateUint8Arrays(authenticatorData, shaClientDataJSON);
+
+  const result = await recoverPublicKeyFromMessageAndSignature(message, signature);
+  return result;
+}
+
+export async function recoverPublicKeyFromMessageAndSignature(message: Uint8Array, signature: Uint8Array) {
```
hariria, line 29: Thanks for adding this in. 

1. Can we add tests for these methods that use the Passkey classes below? We want to make sure users can recover the public key associated with their passkey credential from one or more valid ECDSA signatures in the event that that the relying party loses or misplaces the user's public key. 
2. Also lets include a button in the UI for the `examples/typescript/passkey` folder that tests this

tests/e2e/api/passkey.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    userHandle: "tsSuOOp-4_KUAp9B7-M_EPFwl3giI4cdOzvFiUbDoMM",
+  },
+  type: "public-key",
+  clientExtensionResults: {},
+  authenticatorAttachment: "platform",
+} as AuthenticationResponseJSON;
+
+const { aptos } = getAptosClient();
+
+describe("passkey api", () => {
```
hariria, line 56: Were the api tests resolved after correcting the Secp256k1 private key issue?

src/internal/passkey.ts
hariria: For context on my general thought process here, overall I'd like to provide strong / safe defaults on passkey registration for the user with the ability to override them if a developer wants to.

1. Yes you're on the right track, the `verifyAuthenticationResponse` method [parses `BE` and `BS`](https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/server/src/authentication/verifyAuthenticationResponse.ts#L237C56-L237C72) and returns the `VerifiedAuthenticationResponse`. To make it easier for the user we can just check to make sure both are set to true to determine if the credential is backed up.
2. Per the point above, yes, I think it may make sense to have a strong default for users that allows for authenticators that are well known and ~guaranteed~ to be backed up. `iCloud Keychain (Managed)`, `iCloud Keychain`, and `Google Password Manager` for example. `1Password` and `Dashlane` may also be acceptable but I'm not confident on the rest of them at this moment.

hariria:
Also I'm not sure if SimpleWebAuthn provides this but I think it would be good to provide a method like `checkPlatformBackupable` that checks 
1. browser version (should be [Chrome 118+](https://developer.chrome.com/blog/passkeys-on-icloud-keychain) for backupable passkeys on MacOS iirc, safari, firefox)
2. operating system

See https://passkeys.dev/device-support/ for more info

PR #478, opened 2024-07-25
Add AsyncAccount

BriungRi:
### Description
Adds an `AsyncAccount` that matches `Account` but returns `Promise` for its abstract functions. Later, `AsyncAccount` needs to be used for making async signing requests via Turnkey

### DIFF
--- a/src/plugins/assetUploader/index.ts
+++ b/src/plugins/assetUploader/index.ts
@@ -1,4 +1,4 @@
-import { Account } from "../../account";
+import { Account, AsyncAccount } from "../../account";
 import { AptosConfig } from "../../api/aptosConfig";
 import { IrysAssetUploader } from "./irys";
 import { FundResponse, IAssetUploader, UploadResponse, TaggedFile, AssetUploaderProvider } from "./types";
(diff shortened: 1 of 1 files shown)


src/account/AsyncAccount.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Signing scheme used to sign transactions
+   */
+  abstract signingScheme: SigningScheme;
+
+  /**
+   * Sign a message using the available signing capabilities.
+   * @param message the signing message, as binary input
+   * @return the AccountAuthenticator containing the signature, together with the account's public key
+   */
+  abstract signWithAuthenticator(message: HexInput): AccountAuthenticator | Promise<AccountAuthenticator>;
```
gregnazario, line 38: I would just return a Promise only.  That way you always know you have to await on it (even if it doesn't make any special call).

Otherwise you have to do a lot of "if promise" await, "else do nothing"

src/account/AsyncAccount.ts
```diff
@@ -0,0 +1,70 @@
+import type { AccountAuthenticator } from "../transactions/authenticator/account";
+import { HexInput, SigningScheme } from "../types";
+import type { AccountAddress } from "../core/accountAddress";
+import { AccountPublicKey, Signature, VerifySignatureArgs } from "../core/crypto";
+import { AnyRawTransaction } from "../transactions/types";
+
+/**
+ * Interface for a generic Aptos account.
```
0xmaayan, line 8: please update doc string

src/account/AsyncAccount.ts
0xmaayan: why? if a dev wants to use the `Account` type they should use the Account class then. This can be annoying for external users when more of them start using the tmp branch

0xmaayan (approved):

PR #480, opened 2024-07-30
[bcs] Add BCS Spec

gregnazario:
### Description
This is a prototype for cross SDK specifications.  It points out some interesting differences between SDKs even just for BCS.  We'll need to go and sit down and write proper specifications that cover the behavior we want.

This is a mirror of the Go SDK spec PR https://github.com/aptos-labs/aptos-go-sdk/pull/78 with the same exact specification!

### Test Plan
I got everything running with the below command.  We will need to probably run it in CI, but until we settle on actual specs, I do not want to run them in CI for the time being.

```bash
pnpm _spec
> @aptos-labs/ts-sdk@1.26.0 _spec /opt/git/typescript-sdk
> ./node_modules/.bin/cucumber-js -p default

(node:47604) [DEP0180] DeprecationWarning: fs.Stats constructor is deprecated.
(Use `node --trace-deprecation ...` to show where the warning was created)
223 scenarios (223 passed)
669 steps (669 passed)
0m00.037s (executing steps: 0m00.011s)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "check-version": "scripts/checkVersion.sh",
     "update-version": "scripts/updateVersion.sh && pnpm doc",
     "spec": "pnpm build && pnpm _spec",
-    "_spec": "./node_modules/.bin/cucumber-js -p default"
+    "_spec": "cucumber-js -p default"
   },
   "dependencies": {
     "@aptos-labs/aptos-cli": "^0.2.0",
--- a/package.json
+++ b/package.json
@@ -45,11 +45,13 @@
     "indexer-codegen": "graphql-codegen --config ./src/types/codegen.yaml && pnpm fmt",
     "doc": "scripts/generateDocs.sh",
     "check-version": "scripts/checkVersion.sh",
-    "update-version": "scripts/updateVersion.sh && pnpm doc"
+    "update-version": "scripts/updateVersion.sh && pnpm doc",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -45,11 +45,13 @@
     "indexer-codegen": "graphql-codegen --config ./src/types/codegen.yaml && pnpm fmt",
     "doc": "scripts/generateDocs.sh",
     "check-version": "scripts/checkVersion.sh",
-    "update-version": "scripts/updateVersion.sh && pnpm doc"
+    "update-version": "scripts/updateVersion.sh && pnpm doc",
+    "spec": "pnpm build && pnpm _spec",
+    "_spec": "./node_modules/.bin/cucumber-js -p default"
```
0xmaayan, line 50: why are we using the bin file?

0xmaayan (commented):
have we tested the sdk build with the new dependency on both node and web envs?

package.json
gregnazario: This is a prototype for the moment, I don't expect to merge it in necessarily.

There is this weird conflict, where I needed to use ESM, but it kept giving me conflicts about loading CJS, rather than ESM.

Would love to sit down with someone who's more familiar with it than me and I'm sure we could knock it out.

gregnazario: Figured it out, need to build prior to running.  Should be fixed now.

PR #481, opened 2024-08-02
[New Type] Support Raw Type to Script txn

WGB5445:
### Description

Add BCS support for Script Raw type to the Ts SDK

### Test Plan

Serialization and deserialization for Raw type have been added.

### Related Links

for https://github.com/aptos-labs/aptos-core/pull/14097


src/bcs/serializable/scriptRaw.ts
```diff
@@ -0,0 +1,30 @@
+import { Serializer, Serializable } from "../serializer";
+import { Deserializer } from "../deserializer";
+import { HexInput } from "../../types";
+import { Hex } from "../../core/hex";
+import { TransactionArgument } from "../../transactions/instances/transactionArgument";
+
+export class Raw extends Serializable implements TransactionArgument {
```
0xmaayan, line 7: Either rename the class name or the file name
```suggestion
export class ScriptRaw extends Serializable implements TransactionArgument {
```

src/types/index.ts
```diff
@@ -61,6 +61,7 @@ export enum ScriptTransactionArgumentVariants {
   U16 = 6,
   U32 = 7,
   U256 = 8,
+  Raw = 9,
```
0xmaayan, line 64: I would consider using `Serialized` name - same as what the Rust code uses

0xmaayan (commented):
Sometimes we use `scriptRaw` and sometimes just `Raw` while the Rust code uses `Serialized` - Lets use the same name, consider using `Serialized` to be aligned with the Rust code

Also, please update CHANGELOG

0xmaayan (approved):

gregnazario (approved):

PR #482, opened 2024-08-02
Add helper conversion methods

0xmaayan:

### DIFF
--- a/tests/unit/conversion.test.ts
+++ b/tests/unit/conversion.test.ts
@@ -8,10 +8,38 @@ describe("conversion", () => {
     expect(result).toBe(50000000000);
   });
 
+  test("it converts float amount from human readable format to on chain format", () => {
+    const value = 500.5;
+    const decimals = 8;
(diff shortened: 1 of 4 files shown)


tests/unit/conversion.test.ts
```diff
@@ -0,0 +1,17 @@
...
+    expect(result).toBe(50000000000);
+  });
+
+  test("it converts amount from on chain format to human readable format ", () => {
+    const value = 50000000000;
+    const decimals = 8;
+    const result = convertAmountFromOnChainToHumanReadable(value, decimals);
+    expect(result).toBe(500);
+  });
+});
```
gregnazario, line 17: Do you want to check with decimal values?

Ideally, it probably should use the string version of the decimal, otherwise it may lose precision with 8 decimal points.

0xmaayan: done

0xbe1:
it is a very long name 😆

PR #492, opened 2024-08-23
Update README.md

jmintuitive:
### Description

As part of updating the docs, it's important that we connect new users of the library to aptos.dev and the reference docs. This rewrite also adds more comprehensive example code to help people understand what they have to do in concrete runnable code.

### Test Plan

Inspection via the preview feature.


README.md
```diff
@@ -48,32 +38,30 @@ Then, the SDK can be accessed through `window.aptosSDK`.
...
-```
-
-If you want to pass in a custom config
+import {{ Aptos, AptosConfig, Network }} from "@aptos-labs/ts-sdk"
 
-```ts
-// an optional config information for the SDK client instance.
-const config = new AptosConfig({ network: Network.LOCAL }); // default network is devnet
+// You can use AptosConfig to choose which network to connect to
+const config = new AptosConfig({{ network: Network.TESTNET }});
```
gregnazario, line 47: Why testnet and not devnet?

README.md
```diff
@@ -193,7 +214,13 @@ If you see import error when you do this
 import { Aptos, AptosConfig, Network } from "@aptos-labs/ts-sdk";
 ```
 
-It could be your `tsconfig.json` is not incompatible, make sure your `moduleResolution` is set to `node` instead of `bundler`.
```
gregnazario, line 196: AFAIK this no longer applies, but I guess there's no harm here

README.md
```diff
@@ -193,7 +214,13 @@ If you see import error when you do this
...
+It could be your `tsconfig.json` is not using `node`. Make sure your `moduleResolution` in the `tsconfig.json` is set to `node` instead of `bundler`.
+
+## Experimental Development Version
+
+[![NPM Experimental Version](https://img.shields.io/npm/v/%40aptos-labs/ts-sdk/experimental)][experimental-url]
+![Experimental Node Version](https://img.shields.io/node/v/%40aptos-labs%2Fts-sdk/experimental)
+![Experimental bundle size](https://img.shields.io/bundlephobia/min/%40aptos-labs/ts-sdk/experimental)
```
gregnazario, line 223: Is there an experimental version that's actually being put out?  I think we should just remove it now.

README.md
jmintuitive: That was the advice I got for which one was better for the developer experience. Devnet resets, which seems like less production ready, so if both are testing environments, TESTNET seems more recommended.

jmintuitive:
@gregnazario Updated and responded to comments - ready to merge if you're ok with things :)

gregnazario (approved):

Issue #496, opened 2024-08-26
[Feature Request] Multikey & Multi-ed25519 examples


gregnazario:
Replaces: https://github.com/aptos-labs/aptos-core/issues/10243

PR #502, opened 2024-09-01
Add MultiEd25519Account

yeptos:
### Description
Adds support for MultiEd25519Account

### Test Plan
(needs some work)

### Related Links
#496


gregnazario (commented):
This seems fine, can we add one test to ensure the signatures work correctly?

PR #467, opened 2024-07-16
Add the `NoAccountAuthenticator` variant

junkil-park:
### Description
This PR updates the TypeScript SDK to support the simulation enhancement AIP: https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-92.md.

Key Changes:

Simulation API Update: Allows users to simulate transactions without providing public keys by updating simulateTransaction to accept `PublicKey | null` instead of `PublicKey`. If null is provided, `NoAccountAuthenticator` is used as an authenticator. This new authenticator variant is supported by the Aptos VM as per [this PR](https://github.com/aptos-labs/aptos-core/pull/13714).

Multisig V2 Example Update: The multisig v2 example (multisig_v2.ts) is updated to reflect the change in the multisig transaction simulation behavior. To pre-check the multisig payload before creation, an entry function payload simulation with the multisig account as the sender and 0x0 as the fee payer is used.

### Test Plan
This PR includes the following test scenarios:
* Simulate a transaction without checking the authentication key when the public key is not provided (i.e., public key is null).
* Simulate a fee payer transaction with fee payer as `0x0` and no public key of the fee payer provided.
* Fail to execute a transaction when `NoAccountAuthenticator` is used.

### Related Links
N.A.

PR #522, opened 2024-10-02
Improve fee payer txn submission APIs

heliuchuan:
### Description

This simplifies the fee payer APIs by reducing the steps needed to submit such a transaction while still maintaining support for remote signing.

Now the txn may be submitted either on the fee payer side or the sender side.

This also improves the keyless experience as proofs may be fetched asynchronously, in those cases since submission is an async function where we can await the proof fetch.

### Test Plan

Tested via sponsored txn example.

More tests in progress

### DIFF
--- a/examples/typescript/simple_sponsored_transaction.ts
+++ b/examples/typescript/simple_sponsored_transaction.ts
@@ -74,7 +74,11 @@ const example = async () => {
   const sponsorSignature = aptos.transaction.signAsFeePayer({ signer: sponsor, transaction });
 
   // Submit the transaction to chain
-  const committedTxn = await aptos.signAndSubmitTransaction({ signer: alice, feePayerAuthenticator: sponsorSignature, transaction });
+  const committedTxn = await aptos.signAndSubmitTransaction({
+    signer: alice,
+    feePayerAuthenticator: sponsorSignature,
+    transaction,
+  });
 
   console.log(`Submitted transaction: ${committedTxn.hash}`);
   const response = await aptos.waitForTransaction({ transactionHash: committedTxn.hash });
--- a/examples/typescript/simple_sponsored_transaction.ts
+++ b/examples/typescript/simple_sponsored_transaction.ts
@@ -71,7 +71,7 @@ const example = async () => {
   });
 
   // Submit the transaction to chain
-  const committedTxn = await aptos.signAndSubmitFeePayerTransaction({signer: alice, feePayer: sponsor, transaction});
+  const committedTxn = await aptos.signAndSubmitAsFeePayer({signer: alice, feePayer: sponsor, transaction});
 
   console.log(`Submitted transaction: ${committedTxn.hash}`);
   const response = await aptos.waitForTransaction({ transactionHash: committedTxn.hash });
--- a/examples/typescript/simple_sponsored_transaction.ts
+++ b/examples/typescript/simple_sponsored_transaction.ts
@@ -71,7 +71,7 @@ const example = async () => {
   });
 
   // Submit the transaction to chain
-  const committedTxn = await aptos.signAndSubmitTransaction({signer: alice, feePayer: sponsor, transaction});
+  const committedTxn = await aptos.signAndSubmitTransaction({ signer: alice, feePayer: sponsor, transaction });
 
   console.log(`Submitted transaction: ${committedTxn.hash}`);
   const response = await aptos.waitForTransaction({ transactionHash: committedTxn.hash });
--- a/examples/typescript/simple_sponsored_transaction.ts
+++ b/examples/typescript/simple_sponsored_transaction.ts
@@ -70,8 +70,11 @@ const example = async () => {
     },
   });
 
+  // Sponsor signs
+  const sponsorSignature = aptos.transaction.signAsFeePayer({ signer: sponsor, transaction });
+
   // Submit the transaction to chain
-  const committedTxn = await aptos.signAndSubmitTransaction({ signer: alice, feePayer: sponsor, transaction });
+  const committedTxn = await aptos.signAndSubmitTransaction({ signer: alice, feePayerAuthenticator: sponsorSignature, transaction });
 
   console.log(`Submitted transaction: ${committedTxn.hash}`);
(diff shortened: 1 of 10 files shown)


0xmaayan (commented):
Have you seen this PR? following upcoming changes to the framework
https://github.com/aptos-labs/aptos-ts-sdk/pull/467

heliuchuan:
I have but is it relevant to this PR?  doesn't seem so afaict

heliuchuan:
> Have you seen this PR? following upcoming changes to the framework #467


I have but is it relevant to this PR? doesn't seem so afaict

examples/typescript/simple_sponsored_transaction.ts
```diff
@@ -70,18 +70,8 @@ const example = async () => {
...
-  // Sponsor signs
-  const sponsorSignature = aptos.transaction.signAsFeePayer({ signer: sponsor, transaction });
-
   // Submit the transaction to chain
-  const committedTxn = await aptos.transaction.submit.simple({
-    transaction,
-    senderAuthenticator: senderSignature,
-    feePayerAuthenticator: sponsorSignature,
-  });
+  const committedTxn = await aptos.signAndSubmitTransaction({ signer: alice, feePayer: sponsor, transaction });
```
gregnazario, line 74: The purpose of this example was to show you should sign with 2 separate users, rather than have them colocated on the same machine.

Is there another example that shows this?

src/account/MultiKeyAccount.ts
```diff
@@ -7,7 +7,7 @@ import { AccountAddress, AccountAddressInput } from "../core/accountAddress";
 import { HexInput, SigningScheme } from "../types";
 import { AccountAuthenticatorMultiKey } from "../transactions/authenticator/account";
 import { AnyRawTransaction } from "../transactions/types";
-import { KeylessAccount } from "./KeylessAccount";
+import { KeylessAccountCommon } from "./KeylessAccountCommon";
```
gregnazario, line 10: What's `KeylessAccountCommon`?  Seems like a strange name

src/api/transaction.ts
```diff
@@ -355,15 +346,44 @@ export class Transaction {
...
+    args: FeePayerOrFeePayerAuthenticatorOrNeither & {
+      signer: Account;
+      transaction: AnyRawTransaction;
+    },
+  ): Promise<PendingTransactionResponse> {
+    return signAndSubmitTransaction({
+      aptosConfig: this.config,
+      ...args,
+    });
+  }
```
gregnazario, line 359: This is very niche, is it just so when you're running your own fee payer, and you custody keys for users?

Thoughts on this @0xmaayan ?

gregnazario (commented):
KeylessAccountCommon seems like the wrong name for that interface, and it's a little weird of functionality here

heliuchuan:
> 4. (Missing) signAndSubmitAsFeePayer -> sender: Account, feePayer: Account (1 step sign)

I figure that this doesn't make sense.  If you are submitting as the fee payer you don't have access to the sender account and the sender ought to send you an authenticator that's valid for sponsorship.

Additionally the case is covered by the usual signAndSubmitTransaction.

src/account/MultiKeyAccount.ts
heliuchuan: Its the shared code between Keyless and FederatedKeyless.  

More context [here](https://github.com/aptos-labs/aptos-ts-sdk/pull/495#discussion_r1767098724)

src/api/transaction.ts
heliuchuan: My goal really was just to eliminate the extra step that people sign and then submit.  In the niche case where someone runs their own fee payer or custody for users, they can also use this by providing both accounts.

But mainly I just wanted an API where you have the authenticator from the other party (either as the sender or as the feepayer), and then you can just sign and submit

examples/typescript/simple_sponsored_transaction.ts
heliuchuan: updated it to better represent that they not supposed to be colocated.

src/account/MultiKeyAccount.ts
0xmaayan: I agree. we should have a more meaningful name that "common", at the very least something that tells me it is shared between Keyless and FederatedKeyless

src/api/transaction.ts
0xmaayan: I am abit confused, why we need these changes if we introduce signAndSubmitAsFeePayer?

IMO, the new function arguments is confusing

heliuchuan: So signAndSubmitAsFeePayer requires access to signing capabilities as the feePayer and the authenticator already signed by the sender.  Then the feepayer is left to sign and submission can be done by the feepayer (if they want to).

Alternatively if the sender asks a feepayer to sign and return an authenticator, this allows for them to use signAndSubmit. 

In the case that the sender and feepayer are colocated  then they can use signAndSubmit.  

Right now the paradigm for fee payer
1. get sender authenticator
2. get feepayer authenticator
3. submit via `await aptos.transaction.submit.simple`

I want to eliminate use of await aptos.transaction.submit.simple

I think in vast majority of cases people sign and submit immediately after.  So they should use signAndSubmit.

This is a better devX for Keyless because we support asyncronous proof fetching and it's just simpler.

gregnazario: I think this generally is going to give more confusion, but I'm not going to fight this one too long.

gregnazario (approved):

examples/typescript/simple_sponsored_transaction.ts
```diff
@@ -70,17 +70,14 @@ const example = async () => {
...
 
   // Submit the transaction to chain
-  const committedTxn = await aptos.transaction.submit.simple({
-    transaction,
-    senderAuthenticator: senderSignature,
+  const committedTxn = await aptos.signAndSubmitTransaction({
+    signer: alice,
     feePayerAuthenticator: sponsorSignature,
+    transaction,
   });
```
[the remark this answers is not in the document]
heliuchuan: i guess as i understand order doesn't really matter?  but i get that its the convention

Issue #520, opened 2024-10-01
[Bug] Trying to sign a non-hex message fails on browser platforms


hardsetting:
# 🐛 Bug Description

I was getting issues related to Buffer while signing messages and found out we’re using it [here](https://github.com/aptos-labs/aptos-ts-sdk/blob/37ecd6862fe33f3ee5bed501fa41b4779bde07b5/src/core/crypto/utils.ts#L17).
This should be easily replaceable with `new TextEncoder().encode(message)` which is available for all platforms
I would suggest adding an eslint rule for removing usages of Buffer or playing with the tsconfig.json to make it stricter :)

## How to reproduce

```typescript
const secretKey = Ed25519PrivateKey.generate();
const signature = secretKey.sign('some non-hex message');
```

this raises a`Buffer not defined` error on web platforms that don't polyfill `Buffer`

## Expected Behavior

Signs the utf-8 encoded message

PR #523, opened 2024-10-02
[fix] Allow signing non-hex messages on all platforms

gregnazario:
### Description
Fixes non-hex message signing on browser platforms
Resolves: https://github.com/aptos-labs/aptos-ts-sdk/issues/520


hardsetting (approved):
Nice thank you!

PR #526, opened 2024-10-03
Export the poseidon code in /core/crypto

heliuchuan:
### Description

Allow for this code to be used

### Test Plan

Already exists

### DIFF
--- a/src/core/crypto/poseidon.ts
+++ b/src/core/crypto/poseidon.ts
@@ -99,13 +99,12 @@ export function bytesToBigIntLE(bytes: Uint8Array): bigint {
 }
 
 export function bigIntToBytesLE(value: bigint | number, length: number): Uint8Array {
-  if (typeof value === "number") {
-    // eslint-disable-next-line no-param-reassign
-    value = BigInt(value);
-  }
+  let val = BigInt(value);
   const bytes = new Uint8Array(length);
   for (let i = 0; i < length; i += 1) {
-    bytes[i] = Number(value & BigInt(0xff));
+    bytes[i] = Number(val & BigInt(0xff));
+  }
+  val >>= BigInt(8);
     // eslint-disable-next-line no-param-reassign
     value >>= BigInt(8);
   }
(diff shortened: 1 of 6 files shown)


src/core/crypto/poseidon.ts
```diff
@@ -98,7 +98,11 @@ export function bytesToBigIntLE(bytes: Uint8Array): bigint {
...
 
-export function bigIntToBytesLE(value: bigint, length: number): Uint8Array {
+export function bigIntToBytesLE(value: bigint | number, length: number): Uint8Array {
+  if (typeof value === "number") {
+    // eslint-disable-next-line no-param-reassign
+    value = BigInt(value);
+  }
   const bytes = new Uint8Array(length);
   for (let i = 0; i < length; i += 1) {
     bytes[i] = Number(value & BigInt(0xff));
```
gregnazario, line 108: Let's simplify this, getting rid of the if, and getting rid of the lint ignores

```suggestion
export function bigIntToBytesLE(value: bigint | number, length: number): Uint8Array {
  let val = BigInt(value);
  const bytes = new Uint8Array(length);
  for (let i = 0; i < length; i += 1) {
    bytes[i] = Number(val & BigInt(0xff));
  }
  val >>= BigInt(8);
```

src/core/crypto/poseidon.ts
```diff
@@ -98,10 +98,13 @@ export function bytesToBigIntLE(bytes: Uint8Array): bigint {
...
+export function bigIntToBytesLE(value: bigint | number, length: number): Uint8Array {
+  let val = BigInt(value);
   const bytes = new Uint8Array(length);
   for (let i = 0; i < length; i += 1) {
-    bytes[i] = Number(value & BigInt(0xff));
+    bytes[i] = Number(val & BigInt(0xff));
+  }
+  val >>= BigInt(8);
     // eslint-disable-next-line no-param-reassign
     value >>= BigInt(8);
```
gregnazario, line 109: Remove these two

src/core/crypto/poseidon.ts
```diff
@@ -98,10 +98,13 @@ export function bytesToBigIntLE(bytes: Uint8Array): bigint {
   return result;
 }
 
-export function bigIntToBytesLE(value: bigint, length: number): Uint8Array {
+export function bigIntToBytesLE(value: bigint | number, length: number): Uint8Array {
```
gregnazario, line 101: Do we have correctness tests to ensure that a number is encoded correctly as LE?

I see there's a test that converts using these two functions nested in each other, but that just proves that they're inverses of each other, but not necessarily that the middle format is correct.

heliuchuan: done

gregnazario (approved):

PR #542, opened 2024-10-16
[AccountAddress] Make AccountAddress.fromString more strict

GhostWalker562:
### Description
"The fromString (the relaxed parsing one) function should take a param that says how many chars it is allowed to be missing before it is considered a failure"

`AccountAddress.fromString` takes in a parameter that prevents it from being padded if the address is shorter than "64 - maximum characters missing". This helps protect developers from accidentally passing in accidental address like EVM but providing stricter rules.

### Test Plan

Updated tests to support the new restrictions

### Checklist


src/core/accountAddress.ts
```diff
@@ -322,6 +324,14 @@ export class AccountAddress extends Serializable implements TransactionArgument
       );
     }
 
+    // Ensure that the maxMissing is between 0 and 63, inclusive.
+    if (maxMissingChars > 63 || maxMissingChars < 0) {
```
0xmaayan, line 328: this is not inclusive...?

```
if (maxMissingChars >= 63 || maxMissingChars <= 0)
```

gregnazario (approved):
Please update changelog.

We should probably allow any padding for anytime we read from onchain data, e.g. struct names, which auto shorten, but also... rare

gregnazario:
We will also need to probably make this apply to the other SDKs, mainly user facing ones like Unity

A7610605:
@GhostWalker562 @gregnazario  Hi, I'm guguru. From the genesis block, I'm using a vanity address as my node operator address. Today I noticed that I can't view my address [on explorer](https://explorer.aptoslabs.com/account/0x000000007c9f6724cfc3ead36713389148bbeb48a310dacd43a767b52ada4be7?network=mainnet). After a bit of digging, I found this. ETH addresses are 40 bits by default, I think we should increase maxMissingChars to a bigger value like 23.

PR #550, opened 2024-10-21
[PrivateKey] Add AIP-80 compliance to private keys

GhostWalker562:
### Description
Added static functions onto the PrivateKey class to parse and format hex values. Updated the existing private keys to properly toString into the format as well as parse the correct AIP-80 format.

Updated tests to default to the AIP-80 format, and created a test that checks for non-compliant string.

### Test Plan
Added tests for compliant and non-compliant strings.

### Related Links
https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-80.md

### Checklist

### DIFF
--- a/src/core/crypto/ed25519.ts
+++ b/src/core/crypto/ed25519.ts
@@ -319,6 +319,15 @@ export class Ed25519PrivateKey extends Serializable implements PrivateKey {
    *
    * @returns string representation of the private key.
    */
+  toString(): string {
+    return this.toHexString();
+  }
+
+  /**
+   * Get the private key as a hex string with the 0x prefix.
+   *
+   * @returns string representation of the private key.
+   */
   toHexString(): string {
     return this.signingKey.toString();
   }
@@ -330,7 +339,7 @@ export class Ed25519PrivateKey extends Serializable implements PrivateKey {
    *
    * @returns AIP-80 compliant string representation of the private key.
    */
-  toString(): string {
+  toAIP80String(): string {
     return PrivateKey.formatPrivateKey(this.signingKey.toString(), PrivateKeyVariants.Ed25519);
   }
 
--- a/src/core/crypto/secp256k1.ts
+++ b/src/core/crypto/secp256k1.ts
@@ -253,6 +253,15 @@ export class Secp256k1PrivateKey extends Serializable implements PrivateKey {
    *
    * @returns string representation of the private key
    */
+  toString(): string {
+    return this.toHexString();
+  }
+
+  /**
+   * Get the private key as a hex string with the 0x prefix.
+   *
+   * @returns string representation of the private key.
+   */
   toHexString(): string {
     return this.key.toString();
   }
@@ -264,7 +273,7 @@ export class Secp256k1PrivateKey extends Serializable implements PrivateKey {
    *
    * @returns AIP-80 compliant string representation of the private key.
    */
-  toString(): string {
+  toAIP80String(): string {
     return PrivateKey.formatPrivateKey(this.key.toString(), PrivateKeyVariants.Secp256k1);
   }
 
--- a/tests/unit/ed25519.test.ts
+++ b/tests/unit/ed25519.test.ts
@@ -113,13 +113,13 @@ describe("PrivateKey", () => {
   it("should create the instance correctly without error with AIP-80 compliant private key", () => {
     const privateKey2 = new Ed25519PrivateKey(ed25519.privateKey, false);
     expect(privateKey2).toBeInstanceOf(Ed25519PrivateKey);
-    expect(privateKey2.toString()).toEqual(ed25519.privateKey);
+    expect(privateKey2.toAIP80String()).toEqual(ed25519.privateKey);
   });
 
   it("should create the instance correctly without error with non-AIP-80 compliant private key", () => {
     const privateKey = new Ed25519PrivateKey(ed25519.privateKey, false);
     expect(privateKey).toBeInstanceOf(Ed25519PrivateKey);
-    expect(privateKey.toString()).toEqual(ed25519.privateKey);
+    expect(privateKey.toAIP80String()).toEqual(ed25519.privateKey);
   });
 
   it("should create the instance correctly without error with Uint8Array private key", () => {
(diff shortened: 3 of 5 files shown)


src/core/crypto/privateKey.ts
```diff
@@ -27,3 +30,91 @@ export interface PrivateKey {
...
+  /**
+   * Returns the AIP-80 compliant prefix for a given private key type. Append this to a PrivateKey HexString
+   * to get an AIP-80 compliant string.
+   *
+   * [Read about AIP-80](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-80.md)
+   *
+   * @param type
+   * @returns
+   */
+  public static getAIP80Prefix(type: "ed25519" | "secp256k1"): string {
```
0xmaayan, line 44: we should define a type for the argument. 
do we accept the other types? https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/types/index.ts#L105

src/core/crypto/privateKey.ts
```diff
@@ -27,3 +30,91 @@ export interface PrivateKey {
...
+  /**
+   * Parse a HexInput that may be a HexString, Uint8Array, or a AIP-80 compliant string to a Hex instance.
+   *
+   * [Read about AIP-80](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-80.md)
+   *
+   * @param value - A HexString, Uint8Array, or a AIP-80 compliant string.
+   * @param privateKeyType - The private key type
+   * @param strict - If true, the value MUST be compliant with AIP-80.
+   */
+  public static parseHexInput(value: HexInput, type: "ed25519" | "secp256k1", strict?: boolean): Hex {
```
0xmaayan, line 79: same comment on the type

src/core/crypto/ed25519.ts
```diff
@@ -211,11 +211,12 @@ export class Ed25519PrivateKey extends Serializable implements PrivateKey {
    * Create a new PrivateKey instance from a Uint8Array or String.
    *
    * @param hexInput HexInput (string or Uint8Array)
+   * @param strict If true, private key must AIP-80 compliant.
```
0xmaayan, line 214: lets add the link to the AIP

src/core/crypto/ed25519.ts
```diff
@@ -316,10 +317,21 @@ export class Ed25519PrivateKey extends Serializable implements PrivateKey {
...
   }
 
+  /**
+   * Get the private key as a AIP-80 compliant hex string.
+   *
+   * [Read about AIP-80](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-80.md)
+   *
+   * @returns AIP-80 compliant string representation of the private key.
+   */
+  toString(): string {
```
0xmaayan, line 331: this might be a breaking change

src/core/crypto/privateKey.ts
```diff
@@ -27,3 +30,91 @@ export interface PrivateKey {
...
+    let data: Hex;
+
+    const aip80Prefix = PrivateKey.getAIP80Prefix(type);
```
0xmaayan, line 82: why not have it as a static class variable?

src/core/crypto/privateKey.ts
```diff
@@ -27,3 +30,91 @@ export interface PrivateKey {
...
+    if (typeof value === "string") {
+      if (strict !== true && !value.startsWith(aip80Prefix)) {
```
0xmaayan, line 84:
```suggestion
      if (!strict && !value.startsWith(aip80Prefix)) {
```

src/core/crypto/privateKey.ts
```diff
@@ -27,3 +30,91 @@ export interface PrivateKey {
...
+        // HexString input
+        data = Hex.fromHexInput(value);
+        // If the strictness is false, the user has opted into non-AIP-80 compliant private keys.
+        if (strict !== false) {
```
0xmaayan, line 88:
```suggestion
        if (strict) {
```

src/core/crypto/secp256k1.ts
```diff
@@ -149,11 +149,12 @@ export class Secp256k1PrivateKey extends Serializable implements PrivateKey {
...
    *
    * @param hexInput A HexInput (string or Uint8Array)
+   * @param strict If true, private key must AIP-80 compliant.
    */
-  constructor(hexInput: HexInput) {
+  constructor(hexInput: HexInput, strict?: boolean) {
     super();
 
-    const privateKeyHex = Hex.fromHexInput(hexInput);
+    const privateKeyHex = PrivateKey.parseHexInput(hexInput, "secp256k1", strict);
```
0xmaayan, line 157: lets have global consts for `secp256k1` and `ed25519`. can add it in the [types file](https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/types/index.ts#L105)

src/core/crypto/secp256k1.ts
```diff
@@ -250,10 +251,21 @@ export class Secp256k1PrivateKey extends Serializable implements PrivateKey {
...
 
+  /**
+   * Get the private key as a AIP-80 compliant hex string.
+   *
+   * [Read about AIP-80](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-80.md)
+   *
+   * @returns AIP-80 compliant string representation of the private key.
+   */
+  toString(): string {
+    return PrivateKey.formatPrivateKey(this.key.toString(), "secp256k1");
```
0xmaayan, line 266: if now I am getting the AIP-80 compliant string, does it require me to use the ts-sdk when doing operations with this string? like when submitting a transactions?

tests/unit/ed25519.test.ts
```diff
@@ -111,24 +111,29 @@ describe("Ed25519PublicKey", () => {
 
 describe("PrivateKey", () => {
   it("should create the instance correctly without error", () => {
-    // Create from string
-    const privateKey = new Ed25519PrivateKey(ed25519.privateKey);
+    // Create from non-AIP-80 compliant string
```
0xmaayan, line 114: can we separate the tests for no-AIP-80 and AIP-80 tests?
```
it("should create the instance correctly without error with AIP-80 compliant private key", () => { }
it("should create the instance correctly without error with non-AIP-80 compliant private key", () => { }
```

src/core/crypto/ed25519.ts
0xmaayan: then we would need to do a major release

PR #571, opened 2024-11-10
Add AccountUtils which allows for general serialization of the account classes

heliuchuan:
### Description

This adds serialization functionality to all the account classes.  Clients that use Keyless will want to support the ability to add backup keys or turn the account into a multikey.  This allows clients to cache the account agnostic of the actual signing method impl underneath.  i.e., caching a `MultiKeyAccount` and `KeylessAccount` can work the same and use `AccountUtils.toBytes`

### Test Plan

Added unit tests

### Checklist


src/account/AccountUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+  Secp256k1PrivateKey,
+  ZeroKnowledgeSig,
+} from "../core";
+import { deserializeSchemeAndAddress } from "./utils";
+import { EphemeralKeyPair } from "./EphemeralKeyPair";
+
+/**
+ * Utility functions for working with accounts.
+ */
+export class AccountUtils {
```
hardsetting, line 25: I don't think we need a static class to namespace these functions.
How about either exporting functions directly e.g.

```typescript
export function accountToBytes(account: Account): Uint8Array { ... }
export function accountFromHexInput(input: HexInput): Account { ... }
export function serializeAccount(serializer: Serializer, account: Account) { ... }
export function deserializeAccount(deserializer: Deserializer): Account { ... }
```

or alternatively, a namespace

```typescript
export namespace AccountUtils {
  export function ...
}
```

src/account/AccountUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+    return account;
+  }
+
+  static ed25519AccountFromHex(hex: HexInput): Ed25519Account {
+    const account = this.fromHex(hex);
+    if (!(account instanceof Ed25519Account)) {
+      throw new Error("Deserialization of Ed25519Account failed");
+    }
+    return account;
+  }
```
hardsetting, line 207: Do we need these?
It's a very simple check that users can do on their app if needed.
They can also just do the following if they're handling a single type.
```typescript
const ed25519Account = AccountUtils.fromHex(hex) as Ed25519Account;
```

src/account/MultiKeyAccount.ts
```diff
@@ -85,26 +87,53 @@ export class MultiKeyAccount implements Account, KeylessSigner {
...
-    this.signingScheme = SigningScheme.MultiKey;
 
     this.accountAddress = address ? AccountAddress.from(address) : this.publicKey.authKey().derivedAddress();
 
-    // Get the index of each respective signer in the bitmap
+    // For each signer, find its corresponding position in the MultiKey's public keys array
     const bitPositions: number[] = [];
     for (const signer of signers) {
-      bitPositions.push(this.publicKey.getIndex(signer.publicKey));
+      bitPositions.push(this.publicKey.getIndex(signer.getAnyPublicKey()));
```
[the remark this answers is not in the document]
heliuchuan: No, it could be a KeylessAccount

src/account/AccountUtils.ts
heliuchuan: I could remove, but its just for convenience and for type safety.  Currently there is no Ed25519Account.deserialize so I'm providing this instead.

src/account/AccountUtils.ts
heliuchuan: Ah didn't know namespace was a thing

src/account/AbstractKeylessAccount.ts
```diff
@@ -225,6 +225,10 @@ export abstract class AbstractKeylessAccount extends Serializable implements Key
     }
   }
 
+  getAnyPublicKey(): AnyPublicKey {
+    return new AnyPublicKey(this.publicKey);
+  }
```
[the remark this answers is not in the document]
heliuchuan: I'm down for this, but this is a breaking change

src/account/Account.ts
```diff
@@ -206,10 +206,9 @@ export abstract class Account {
    * @category Account (On-Chain Model)
    */
   static fromPrivateKey(args: CreateEd25519AccountFromPrivateKeyArgs): Ed25519Account;
-  static fromPrivateKey(args: CreateEd25519SingleKeyAccountFromPrivateKeyArgs): SingleKeyAccount;
   static fromPrivateKey(args: CreateSingleKeyAccountFromPrivateKeyArgs): SingleKeyAccount;
-  static fromPrivateKey(args: CreateAccountFromPrivateKeyArgs): Account;
-  static fromPrivateKey(args: CreateAccountFromPrivateKeyArgs) {
+  static fromPrivateKey(args: CreateAccountFromPrivateKeyArgs): SingleKeyAccount;
+  static fromPrivateKey(args: CreateAccountFromPrivateKeyArgs): Ed25519Account | SingleKeyAccount {
```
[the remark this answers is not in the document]
heliuchuan: I don't think we lose type inference here.  I made this changes to simplify things a bit.  I'll remove from the PR though since it's not core to the PR.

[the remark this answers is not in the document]
heliuchuan: Oh I know why.  Because I wanted to change SingleKeyAccount.privateKey to not be of type PrivateKey and be Ed25519PrivateKey | Secp256k1PrivateKey instead.  These are just the downstream changes.

src/core/crypto/multiKey.ts
```diff
@@ -218,7 +218,7 @@ export class MultiKey extends AccountPublicKey {
     if (index !== -1) {
       return index;
     }
-    throw new Error("Public key not found in MultiKey");
+    throw new Error(`Public key ${publicKey} not found in MultiKey ${this}`);
```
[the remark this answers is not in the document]
heliuchuan: Public key 0x002072d1cb1e30c99a1bbd2901eeb14806bdba4efd83bc9b7effa504aa6a0aab826e not found in MultiKey 0x002019288a67479d4a06cd971dcb45dd442600721d2feec31ccba630ae9f0dbfd2da,0x002098892ad12b370280b5694d2ff9c36d5e796da220aeb7f78a0924238e51a2918e,0x0141041ca5c5e67533a9b89948031b0972c94dc80d6427deec52ba224c3ec66aa5a624b5ef9fdaf1cde8ca3c43425473e9f3126d164bebdc93538ea5dc76ff88ffc643

src/core/crypto/multiKey.ts
```diff
@@ -218,7 +218,7 @@ export class MultiKey extends AccountPublicKey {
     if (index !== -1) {
       return index;
     }
-    throw new Error("Public key not found in MultiKey");
+    throw new Error(`Public key ${publicKey} not found in MultiKey ${this.publicKeys}`);
```
hardsetting, line 221: Do you know what does `this.publicKeys` serialize to?
Could possibly be `[Object object]`

src/account/AccountUtils.ts
```diff
@@ -0,0 +1,216 @@
...
+    }
+  }
+
+  export function toHexStringWithoutPrefix(account: Account): string {
+    return Hex.hexInputToStringWithoutPrefix(toBytes(account));
+  }
+
+  export function toHexString(account: Account): string {
+    return Hex.hexInputToString(toBytes(account));
+  }
```
hardsetting, line 113: What do you think about just having `toHex`?
It would be consistent with `bcsToBytes()` and `bcsToHex()` we currently have.
And you can achieve both with `AccountUtils.toHex().toString()` or `AccountUtils.toHex().toStringWithoutPrefix()`

hardsetting (approved):

src/account/MultiKeyAccount.ts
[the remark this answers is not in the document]
hardsetting: A few lines above:
```typescript
const signers: SingleKeySigner[] = args.signers.map((signer) =>
      signer instanceof Ed25519Account ? SingleKeyAccount.fromEd25519Account(signer) : signer,
    );
```

afaiu `KeylessAccount` is not a subclass of `SingleKeySigner`

src/account/AccountUtils.ts
heliuchuan: Me personally I prefer not dealing with the toHex wrapper.  toHex is kind of misleading because in many libraries such a function returns bytes or hex string, not a custom Hex class.

Issue #576, opened 2024-11-12
aptos.getAccountModules()


RosalinaLuma620:
# 🐛 Bug Description

When I use this SDK to build React Front-end , I invoked the function below .
## How to reproduce

**Code snippet to reproduce**
```typescript
export const getAccountModulesWrapper = async (accountAddress: string) => {
    return await aptos.getAccountModules({
        accountAddress: accountAddress,
        options: {
            limit: 10,
            offset: 1
        }
    })
}
```

**Stack trace / error message**
![image](https://github.com/user-attachments/assets/09fb6cca-8aec-49ac-b45e-f7111e5b94da)

So I checked network records . 

The api request payload is like ( This is how the SDK translates the aptos.getAccountModules() api to do )
**https://api.testnet.aptoslabs.com/v1/accounts/0xf4fd53f1334877ef7aae3241ba87708bf38c618aaa2a22f09e84720fab3328ee/modules?start=11&limit=10**

Notice the query params, is **start** , not **offset**

so I change the link like this to try 
**https://api.testnet.aptoslabs.com/v1/accounts/0xf4fd53f1334877ef7aae3241ba87708bf38c618aaa2a22f09e84720fab3328ee/modules?offset=11&limit=10**

it worked.

## System information

**System details:**
- Typescript SDK Version : Version 5.0.4
- Aptos SDK Version : Version 1.27.1
- Platform : Mac OS M2 

Basically , it's a query params convert issue.

gregnazario:
Yeah, looks like a bug.  Actually amazed it hadn't been hit until now.

Offset is actually a long input, and not just a number to start with, so the PR attached will fix this and the documentat

PR #581, opened 2024-11-14
[account] Fix pagination on AccountModules and AccountResources

gregnazario:
### Description
Resolves https://github.com/aptos-labs/aptos-ts-sdk/issues/576

### Test Plan
Added pagination tests for both calls

### Checklist

### DIFF
--- a/src/client/get.ts
+++ b/src/client/get.ts
@@ -191,3 +191,45 @@ export async function paginateWithCursor<Req extends Record<string, any>, Res ex
   } while (cursor !== null && cursor !== undefined);
   return out as Res;
 }
+
+/// This function is a helper for paginating using a function wrapping an API using offset instead of start
+export async function paginateWithObfuscatedCursor<Req extends Record<string, any>, Res extends Array<{}>>(
+  options: GetAptosRequestOptions,
+): Promise<Res> {
+  const out: any[] = [];
+  let cursor: string | undefined;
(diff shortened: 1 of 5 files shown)


src/client/get.ts
```diff
@@ -159,3 +159,45 @@ export async function paginateWithCursor<Req extends Record<string, any>, Res ex
   } while (cursor !== null && cursor !== undefined);
   return out as Res;
 }
+
+/// This function is a helper for paginating using a function wrapping an API using offset instead of start
+export async function paginateWithObfuscatedCursor<Req extends Record<string, any>, Res extends Array<{}>>(
```
0xmaayan, line 164: iiuc, the only 2 differences is the `start vs offset` params and the `re-evaluate length calculation` - any reason not to simply use the existing `paginateWithCursor` but just provide a flag for the `re-evaluate length calculation` part? `requestParams` are already being passed as an argument to the `paginateWithCursor` function.

gregnazario: I wanted to be more explicit because the `start` vs `offset` is actually more of the problem that the one with the `obfuscated cursor` is not useful / a user shouldn't provide it directly.  You would have to get it from a previous function call.

Also, I think the cursor header isn't always set, so that whole logic probably needs to be checked.

Basically there are 2 code paths on the serverside for pagination, and I'm trying to separate them in teh event we have other errors with it.

0xmaayan (approved):

PR #583, opened 2024-11-18
Add secp256k1 verify and pubKey recovery function

heliuchuan:
### Description

With secp256k1 you can recover the publicKey from the signature and message.  This is common in EVM world.  This will ease migrations as it allows users to use familiar APIs, which may be missing the publicKey.

### Test Plan

e2e test added

### Checklist

### DIFF
--- a/src/api/account.ts
+++ b/src/api/account.ts
@@ -2,7 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { Account as AccountModule } from "../account";
-import { AccountAddress, PrivateKey, AccountAddressInput, createObjectAddress } from "../core";
+import {
+  AccountAddress,
+  PrivateKey,
+  AccountAddressInput,
+  createObjectAddress,
+  Secp256k1Signature,
+  AnyPublicKey,
+  AnySignature,
+} from "../core";
 import {
   AccountData,
   AnyNumber,
@@ -11,6 +19,7 @@ import {
   GetAccountOwnedTokensFromCollectionResponse,
   GetAccountOwnedTokensQueryResponse,
   GetObjectDataQueryResponse,
+  HexInput,
   LedgerVersionArg,
   MoveModuleBytecode,
   MoveResource,
@@ -39,6 +48,7 @@ import {
   getResources,
   getTransactions,
   lookupOriginalAccountAddress,
+  verifySecp256k1Account,
 } from "../internal/account";
 import { APTOS_COIN, APTOS_FA, ProcessorType } from "../utils/const";
 import { AptosConfig } from "./aptosConfig";
@@ -873,5 +883,45 @@ export class Account {
    */
   async deriveAccountFromPrivateKey(args: { privateKey: PrivateKey }): Promise<AccountModule> {
     return deriveAccountFromPrivateKey({ aptosConfig: this.config, ...args });
+  }
+
+  /**
+   * Verifies a Secp256k1 account by checking the signature against the message and account's authentication key.
+   *
+   * This function takes a message and signature, and attempts to recover the public key that created the signature.
+   * It then verifies that the recovered public key matches the authentication key stored on-chain for the provided account address.
+   *
+   * If a recovery bit is provided, it will only attempt verification with that specific recovery bit.
+   * Otherwise, it will try all possible recovery bits (0-3) until it finds a match.
+   *
+   * @param args - The arguments for verifying the Secp256k1 account
+   * @param args.message - The message that was signed
+   * @param args.signature - The signature to verify (either raw hex or Secp256k1Signature object)
+   * @param args.recoveryBit - Optional specific recovery bit to use for verification
+   * @param args.accountAddress - The address of the account to verify
+   * @returns The recovered public key if verification succeeds
+   * @throws Error if verification fails or no matching public key is found
+   *
+   * @example
+   * ```typescript
+   * import { Aptos, AptosConfig, Network } from "@aptos-labs/ts-sdk";
+   *
+   * const config = new AptosConfig({ network: Network.DEVNET });
+   * const aptos = new Aptos(config);
+   *
+   * const publicKey = await aptos.verifySecp256k1Account({
+   *   message: "0x1234...",
+   *   signature: "0x5678...",
+   *   accountAddress: "0x1"
+   * });
+   * ```
+   */
+  async verifySecp256k1Account(args: {
+    message: HexInput;
+    signature: HexInput | Secp256k1Signature | AnySignature;
+    recoveryBit?: number;
+    accountAddress: AccountAddressInput;
+  }): Promise<AnyPublicKey> {
+    return verifySecp256k1Account({ aptosConfig: this.config, ...args });
   }
 }
--- a/src/core/crypto/secp256k1.ts
+++ b/src/core/crypto/secp256k1.ts
@@ -4,14 +4,18 @@
 import { sha3_256 } from "@noble/hashes/sha3";
 import { secp256k1 } from "@noble/curves/secp256k1";
 import { HDKey } from "@scure/bip32";
+import { bytesToNumberBE, inRange } from "@noble/curves/abstract/utils";
 import { Serializable, Deserializer, Serializer } from "../../bcs";
 import { Hex } from "../hex";
 import { HexInput, PrivateKeyVariants } from "../../types";
 import { isValidBIP44Path, mnemonicToSeed } from "./hdKey";
 import { PrivateKey } from "./privateKey";
-import { PublicKey, VerifySignatureArgs } from "./publicKey";
+import { PublicKey } from "./publicKey";
 import { Signature } from "./signature";
 import { convertSigningMessage } from "./utils";
+
+const secp256k1P = BigInt("0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f");
+const secp256k1N = BigInt("0xfffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141");
 
 /**
  * Represents a Secp256k1 ECDSA public key.
@@ -53,7 +57,7 @@ export class Secp256k1PublicKey extends PublicKey {
    * @param args.message - The message that was signed.
    * @param args.signature - The signature to verify against the public key.
    */
-  verifySignature(args: VerifySignatureArgs): boolean {
+  verifySignature(args: { message: HexInput; signature: Secp256k1Signature }): boolean {
     const { message, signature } = args;
     const messageToVerify = convertSigningMessage(message);
     const messageBytes = Hex.fromHexInput(messageToVerify).toUint8Array();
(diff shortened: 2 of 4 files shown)


src/api/account.ts
```diff
@@ -874,4 +884,44 @@ export class Account {
...
+   * const publicKey = await aptos.verifySecp256k1Account({
+   *   message: "0x1234...",
+   *   signature: "0x5678...",
+   *   accountAddress: "0x1"
+   * });
+   * ```
+   */
+  async verifySecp256k1Account(args: {
+    message: HexInput;
+    signature: HexInput | Secp256k1Signature | AnySignature;
```
0xmaayan, line 921: why do we accept `AnySignature` if we only verify for `Secp256k1Account`? why is `Secp256k1Signature` is not part of `AnySignature`? Also, why do we accept `HexInput` as a signature? dont we want to make sure the function accepts a valid signature?

src/internal/account.ts
```diff
@@ -795,3 +803,72 @@ export async function isAccountExist(args: { aptosConfig: AptosConfig; authKey:
...
+export async function verifySecp256k1Account(args: {
+  aptosConfig: AptosConfig;
+  message: HexInput;
+  signature: HexInput | Secp256k1Signature | AnySignature;
+  recoveryBit?: number;
+  accountAddress: AccountAddressInput;
+}): Promise<AnyPublicKey> {
+  const { aptosConfig, message, recoveryBit, accountAddress } = args;
+  let signature: HexInput | Secp256k1Signature;
+  if (args.signature instanceof AnySignature) {
```
0xmaayan, line 816: `instanceOf` is known to be causing a lot of problem, specifically when using the TS sdk from different bundles
See: https://github.com/aptos-labs/aptos-ts-sdk/pull/462

tests/e2e/api/account.test.ts
```diff
@@ -301,6 +303,52 @@ describe("account api", () => {
       expect(tokens[0].current_token_data?.token_name).toBe("Test Token");
     });
 
+    test("it verifies a secp256k1 signature", async () => {
```
0xmaayan, line 306: lets separate the "succeed to verify" and "failed to verify" tests

src/core/crypto/secp256k1.ts
```diff
@@ -124,6 +128,33 @@ export class Secp256k1PublicKey extends PublicKey {
   static isInstance(publicKey: PublicKey): publicKey is Secp256k1PublicKey {
     return "key" in publicKey && (publicKey.key as any)?.data?.length === Secp256k1PublicKey.LENGTH;
   }
+
+  static fromSignatureAndMessage(args: {
```
0xmaayan, line 132: please add a doc string

src/api/account.ts
gregnazario: I agree, public key recovery should be in the signature to return.  I don't think it belongs at the account level given you can't recover the private key

gregnazario (commented):
Let's compartmentalize this properly to the Secp256k1 signatures since it applies to that, and then we won't need to do other parsing

src/api/account.ts
heliuchuan: Because Secp256k1Account does not exist.  SingleKeyAccount exists which returns AnySignature when it signs something.  You can get the inner Secp256k1Signature via signature.signature, but you would still have to type check it into Secp256k1Signature.  Having the function handle it seems better DevX

heliuchuan: @gregnazario not sure what you are suggesting

heliuchuan: Oh and HexInput is again for convenience.  The function will check signature validity.

src/internal/account.ts
heliuchuan: @hardsetting can you give guidance here?  `variant` is private on AnySignature and AnyPublicKey

src/core/crypto/secp256k1.ts
```diff
@@ -124,6 +128,33 @@ export class Secp256k1PublicKey extends PublicKey {
...
+    const { signature, message, recoveryBit } = args;
+    let signatureBytes: Uint8Array;
+    if (signature instanceof Secp256k1Signature) {
+      signatureBytes = signature.toUint8Array();
+    } else {
+      signatureBytes = Hex.fromHexInput(signature).toUint8Array();
+      if (signatureBytes.length !== Secp256k1Signature.LENGTH) {
+        throw new Error(`Signature length should be ${Secp256k1Signature.LENGTH}`);
+      }
+    }
```
hardsetting, line 146: I would maybe do the opposite so as to not repeat the length check (currently done here and in the `Secp256k1Signature` constructor)

```typescript
const signature = signatureInput instanceof Secp256k1Signature ? signatureInput : new Secp256k1Signature(signatureInput);
const signatureBytes = signature.toUint8Array();
```

src/internal/account.ts
hardsetting: We could make the `variant` into public readonly.

What is the use case for `verifySecp256k1Account`?
Depending on that, using `instanceof` should be fine.
For example, if it's expected to be used for accounts instantiated or deserialized by the same application (which is probably be the case), we should be good.

src/internal/account.ts
```diff
@@ -795,3 +803,104 @@ export async function isAccountExist(args: { aptosConfig: AptosConfig; authKey:
...
+ * It then verifies that the recovered public key matches the authentication key stored on-chain for the provided account address.
+ *
+ * If a recovery bit is provided, it will only attempt verification with that specific recovery bit.
+ * Otherwise, it will try all possible recovery bits (0-3) until it finds a match.
+ *
+ * @param args - The arguments for verifying the Secp256k1 account
+ * @param args.aptosConfig - The configuration for connecting to the Aptos blockchain.
+ * @param args.message - The message that was signed
+ * @param args.signature - The signature to verify (either raw hex or Secp256k1Signature object)
+ * @param args.recoveryBit - Optional specific recovery bit to use for verification
```
hardsetting, line 820: Can we ever expect the recoveryBit to be provided manually by the user?
How would I obtain it when signing a message using a Secp256k1PrivateKey

src/internal/account.ts
```diff
@@ -795,3 +803,104 @@ export async function isAccountExist(args: { aptosConfig: AptosConfig; authKey:
...
+ * });
+ * ```
+ */
+export async function verifySecp256k1Account(args: {
+  aptosConfig: AptosConfig;
+  message: HexInput;
+  signature: HexInput | Secp256k1Signature | AnySignature;
+  recoveryBit?: number;
+  accountAddress: AccountAddressInput;
+}): Promise<AnyPublicKey> {
```
hardsetting, line 845: This function seems to solve a very specific use case:
obtain a public key for an account, from a signed message and address, only for Secp256k1 accounts.

Why would I need it as a dapp developer?
Are you using this internally somewhere?

src/internal/account.ts
heliuchuan: It's unclear to me.  We dont provide in our APIs but other SDK may.  It is optional since there are a finite number of bit options and you can simply loop if not provided.

src/internal/account.ts
heliuchuan: I worked around this for now

src/core/crypto/singleKey.ts
```diff
@@ -199,7 +199,7 @@ export class AnySignature extends Signature {
   /**
    * Index of the underlying enum variant
    */
-  private readonly variant: AnySignatureVariant;
+  private readonly anySignatureVariant: AnySignatureVariant;
```
gregnazario, line 202: Let's not rename this in this PR.  It's an unnecessary refactor.  To make changes like this should be in a separate commit and labeled appropriately

src/core/crypto/secp256k1.ts
```diff
@@ -124,6 +129,33 @@ export class Secp256k1PublicKey extends PublicKey {
...
+    if (!inRange(r, BigInt(1), secp256k1P)) throw new Error("Invalid secp256k1 signature - r ≥ p");
+    const s = bytesToNumberBE(signatureBytes.subarray(32, 64)); // Let s = int(sig[32:64]); fail if s ≥ n.
+    if (!inRange(s, BigInt(1), secp256k1N)) throw new Error("Invalid secp256k1 signature - s ≥ n");
+    const sig = new secp256k1.Signature(r, s);
+    const messageToVerify = convertSigningMessage(message);
+    const messageBytes = Hex.fromHexInput(messageToVerify).toUint8Array();
+    const messageSha3Bytes = sha3_256(messageBytes);
+    const publicKeyBytes = sig.addRecoveryBit(recoveryBit).recoverPublicKey(messageSha3Bytes).toRawBytes(false);
+    return new Secp256k1PublicKey(publicKeyBytes);
+  }
```
gregnazario, line 158: Can we put whitespace in here?

src/core/crypto/secp256k1.ts
```diff
@@ -318,7 +350,7 @@ export class Secp256k1Signature extends Signature {
    * The signature bytes
    * @private
    */
-  private readonly data: Hex;
+  private readonly secp256k1SigData: Hex;
```
gregnazario, line 353: again, another refactor commit

src/core/crypto/singleKey.ts
```diff
@@ -264,6 +264,7 @@ export class AnySignature extends Signature {
 
   static isInstance(signature: Signature): signature is AnySignature {
     return (
+      "anySignatureVariant" in signature &&
```
gregnazario, line 267: is this why you're renaming it?

heliuchuan: yes, I fear that the isInstance methods are not specific enough because 'data' and 'variant' are too generic.

src/core/crypto/singleKey.ts
heliuchuan: I figure it's NBD because it is private.  I'd agree if it were not private, but it's completely internal as I understand.

gregnazario (dismissed):

src/core/crypto/singleKey.ts
gregnazario: I just hope someone's not doing some weird serialization of this, but for now it should be fine.

These kinds of changes often lead to unexpected breaking changes.

src/core/crypto/secp256k1.ts
```diff
@@ -53,7 +57,7 @@ export class Secp256k1PublicKey extends PublicKey {
    * @param args.message - The message that was signed.
    * @param args.signature - The signature to verify against the public key.
    */
-  verifySignature(args: VerifySignatureArgs): boolean {
+  verifySignature(args: { message: HexInput; signature: Secp256k1Signature }): boolean {
```
gregnazario, line 60: isn't this a breaking change?

src/internal/account.ts
```diff
@@ -795,3 +804,95 @@ export async function isAccountExist(args: { aptosConfig: AptosConfig; authKey:
...
+  }
+
+  for (let i = 0; i < 4; i += 1) {
+    try {
+      const publicKey = new AnyPublicKey(
+        Secp256k1PublicKey.fromSignatureAndMessage({
+          signature,
+          message,
+          recoveryBit: i,
+        }),
```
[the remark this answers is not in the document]
gregnazario: Nvm, recovery bit is a misnomer since there's an extremely low chance of it being 2 bits.

src/core/crypto/secp256k1.ts
heliuchuan: I guess this is a breaking change technically...  though its a break that alerts the dev to code issues

but I'll revert it and deal with in another PR, maybe.

src/api/account.ts
hardsetting: I also would expect this function to be in a different place than `Account`.

I think `Secp256k1PublicKey.fromMessageAndSignature` makes total sense (`fromSignedMessage` could be even more concise).

Ideally that's all the devs would need, and they could do the following:

```typescript
const innerPublicKey = new Secp256k1PublicKey.fromSignedMessage({ message, signature });
const publicKey = new AnyPublicKey({ publicKey: innerPublicKey });
const derivedAddress = publicKey.authKey().derivedAddress.toString();

// Simple case
const isValid = derivedAddress === accountAddress;

// Handle rotated auth keys
const { authentication_key } = await aptos.getAccountInfo({ accountAddress });
const isValid = derivedAddress === authentication_key;
```

now.. how can we make this even easier / more concise?
1) Going from `Secp256k1PublicKey` to `AnyPublicKey` is a bit annoying, maybe we can have the constructor there instead: `AnyPublicKey.fromSecp256k1SignedMessage({ message, signature })`
2) The action of verifying the publicKey is associated to an account from its address is agnostic to the signature scheme, so we could have `verifyAuthenticationKey({ authKey, accountAddress })` which could live in the `api` to avoid dependency cycles

hardsetting: I really hate the `AnyPublicKey` name ugh

src/core/crypto/secp256k1.ts
```diff
@@ -124,6 +128,36 @@ export class Secp256k1PublicKey extends PublicKey {
...
+   *
+   * @param args - The arguments for recovering the public key.
+   * @param args.signature - The signature to recover the public key from.
+   * @param args.message - The message that was signed.
+   * @param args.recoveryBit - The recovery bit to use for the public key.
+   */
+  static fromSignatureAndMessage(args: {
+    signature: HexInput | Secp256k1Signature;
+    message: HexInput;
+    recoveryBit: number;
```
hardsetting, line 143: I would make the recoveryBit optional at this level, and handle it with recursion.

```typescript
static fromSignedMessage({ signature, message, recoveryBit }) {
  if (recoveryBit === undefined) {
    try {
      for (let i = 0; i < 4; i += 1) {
        return Secp256k1PublicKey.fromSignedMessage({ signature, message recoveryBit: i });
      }
    } catch {
      // ...
    }
    throw new Error('Unable to derive public key');
  }
  
  // regular implementation
}
```

hardsetting: or have a private internal implementation I guess

heliuchuan: Recovery bit must be set here, because if you loopthrough all options you actually get two public keys.

src/api/account.ts
heliuchuan: So the problem is you need to look up the auth key with the address, and only via the auth key can you figure out which public key is correct.  If a recovery is provided 2 public keys would be returned.

I'm down for AnyPublicKey.fromSecp256k1SignedMessage({ message, signature }) though

gregnazario: Yeah I re-evaluated it and did similar here https://github.com/aptos-labs/aptos-go-sdk/pull/108

hardsetting: > So the problem is you need to look up the auth key with the address, and only via the auth key can you figure out which public key is correct. If a recovery is provided 2 public keys would be returned.

Oh I see, gotcha makes sense

gregnazario (approved):

PR #584, opened 2024-11-18
Add `gasProfile()` method to run Move gas profiling

0xmaayan:
### Checklist

### DIFF
--- a/src/cli/move.ts
+++ b/src/cli/move.ts
@@ -310,6 +310,22 @@ export class Move {
     return this.runCommand(cliArgs, showStdout);
   }
 
+  async gasProfile(args: {
+    network: string;
+    transactionId: string;
+    extraArguments?: Array<string>;
+    showStdout?: boolean;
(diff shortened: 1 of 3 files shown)


src/cli/move.ts
```diff
@@ -310,6 +310,22 @@ export class Move {
...
   }
 
+  async gasProfile(args: {
+    network: string;
+    transactionId: string;
+    extraArguments?: Array<string>;
+    showStdout?: boolean;
+  }): Promise<{ output: string; result?: any }> {
+    const { network, transactionId, extraArguments, showStdout } = args;
+    const cliArgs = ["aptos", "move", "replay", "--profile-gas", "--network", network, "--txn-id", transactionId];
```
gregnazario, line 320: Oh interesting didn't know it works on historical transactions.

Maybe rename it to gasProfileExistingTransaction or something like that?

Or just leave room in this function to take a different input that also allows gas profiling new transactions.

gregnazario (approved):

PR #585, opened 2024-11-19
Add example for batch transactions using Aptos SDK

Akk525:
### Description
This pull request adds a new example, `batchTransactions.js`, to the `examples/javascript` directory. The example demonstrates how to perform batch transactions using the Aptos SDK, showcasing the process of sending multiple transactions sequentially in a script. This example is designed to help developers optimize workflows that involve multiple on-chain operations.

### Test Plan
1. Run the script using the following command:
```bash
   node examples/javascript/batchTransactions.js
   
2. Verify the following:
   - The script successfully submits multiple transactions to the Aptos Devnet.
   - Transaction hashes are displayed in the console.
   - Each transaction hash can be validated on the [Aptos Devnet Explorer](https://explorer.devnet.aptos.dev/) by searching     for the transaction hash.

### Related Links
No related issues or pull requests.

### Checklist
  **Yes**, the code has been formatted using `pnpm fmt`.
  **No**, as this is an addition to the examples directory and does not modify core functionality.

### DIFF
--- a/examples/javascript/batchTransactions.js
+++ b/examples/javascript/batchTransactions.js
@@ -1,0 +1,78 @@
+// Import required modules and classes from the 'aptos' library
+import { AptosClient, AptosAccount, FaucetClient, TransactionBuilderEd25519 } from 'aptos';
+
+// Define the URL for the Aptos blockchain node (Devnet) and the Faucet service
+const NODE_URL = 'https://fullnode.devnet.aptoslabs.com';
+const FAUCET_URL = 'https://faucet.devnet.aptoslabs.com';
+
+// Create an Aptos client to interact with the blockchain node
+const client = new AptosClient(NODE_URL);
+
+// Create a Faucet client to fund accounts using the Devnet faucet
+const faucet = new FaucetClient(NODE_URL, FAUCET_URL);
+
+// Create a new Aptos account (this generates a private/public key pair and a new address)
+const account = new AptosAccount();
+
+// Fund the new account using the faucet, adding 1,000,000 micro-coins (1 Aptos coin)
+// Note: `account.address()` retrieves the address of the account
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: examples/javascript/batchTransactions.js [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+
+  // Combine the two transactions into an array for batch processing
+  const transactions = [transferPayload1, transferPayload2];
+
+  // Loop through each transaction payload, send the transaction, and track its hash
+  for (const txPayload of transactions) {
+    const txHash = await sendTransaction(txPayload); // Send the transaction and get its hash
+    transactionHashes.push(txHash); // Store the transaction hash in the array
+    console.log(`Transaction submitted with hash: ${txHash}`); // Log the transaction hash
+  }
```
gregnazario, line 71: If you're looking to send multiple transactions at once, should probably do these in parallel and handle the sequence number appropriately.

```suggestion
  // Loop through each transaction payload, send the transaction, and track its hash
  const promises = transactions.map((payload) => sendTransaction(payload)); // Handle errors here
  
  const transactionHashes = Promise.all(promises);
  console.log(`Transactions submitted with hashes: ${JSON.stringify(transactionHashes)}`); // Log the transaction hashes
  }
```

gregnazario (commented):
We generally have avoided the JavaScript examples, just because we already are building them in TypeScript, which tends to have better practices / showing how to manage types (where JavaScript you can omit them).

I'm curious have you seen [this example in TypeScript](https://github.com/aptos-labs/aptos-ts-sdk/blob/main/examples/typescript/batch_funds.ts)?

It uses a transaction worker thread to manage the batch send, which might not be exactly what you want.

I'd love feedback as well if JavaScript examples are important rather than just TypeScript.

cc @0xmaayan

PR #565, opened 2024-11-04
Implement a new api for building script payload

runtian-zhou:
### Checklist
  - [ ] Have you ran `pnpm fmt`?
  - [ ] Have you updated the `CHANGELOG.md`?

Issue #578, opened 2024-11-14
[Feature Request]  bundle size bloating


yeptos:
# 🚀 Feature Request Description

## Motivation

Currently, the `poseidon-lite` package takes up over 450k bytes after gzip compression. For us, this accounts to a significant portion from the total bundle size of our webapp. Since the `poseidon-lite` package is only used when deriving Keyless accounts, I think we need to explor ways to make this optional and load asynchronously when the module is needed.

PR #587, opened 2024-11-20
Runtianz/new script composer

runtian-zhou:
### Description
Reopen the script composer feature with updated dependency built into aptos-labs.

### Test Plan
CI

Tested that the package can work in web, nodejs and nextjs environment.

### Related Links

#565 

### Checklist


0xmaayan:
Please test and validate it 
1. Passes all SDK unit + examples tests
2. Does not break a client-side render app (react)
3. Does not break a server-side render app (nextjs)

0xmaayan:
> > Please test and validate it
> > 
> > 1. Passes all SDK unit + examples tests
> > 2. Does not break a client-side render app (react)
> > 3. Does not break a server-side render app (nextjs)
> 
> I have tested a React + Vite client-side render app and a Next.js server-side render app. Both work fine.
> 
> However, it's worth noting that in the current version, the initialization is done when the SDK loads, so the bundled JavaScript is a bit large—about 5 MB for the client-side.

The bundle size is a concern. I'd like to get consensus from the sdk team before approving it. @gregnazario

0xmaayan:
> @0xmaayan I personally think its too much 😞 since it's already 5x the bundle size. People are already complaining about 421kb from `poseidon-lite` #578 but if it's really needed in the SDK I'm not going to push back that much.
> 
> I'm curious if we can test it in a react-native environment as well?

@runtian-zhou 
1. Any reason we dont move the wasm initialization to the end user?
2. Thoughts on releasing it under a tagged version?

@GhostWalker562 I dont think the SDK officially supports react-native, but I assume you want it tested for Petra?

runtian-zhou:
> Thoughts on releasing it under a tagged version?

I think that would be amazing.

0xmaayan:
> > Thoughts on releasing it under a tagged version?
> 
> I think that would be amazing.

Cool. Then we dont need to merge it to `main`, but release a tagged version from this branch

runtian-zhou:
Let's release it first and get feedbacks from the community. I'm getting multiple asks for this feature so would be great to let them try it out first. In the meantime, we can optimize on the packaging situation.

lumos42:
eagerly waiting to use this feature in our app

itsmnthn:
> @0xmaayan I personally think its too much 😞 since it's already 5x the bundle size. People are already complaining about 421kb from `poseidon-lite` #578 but if it's really needed in the SDK I'm not going to push back that much.
> 
> I'm curious if we can test it in a react-native environment as well?

One workaround for the size is lazy load.
Script Composer can be different package and could by lazy loaded and uses ts-sdk to compose the script wen needed

wdyt?

0xmaayan:
This feature has been available for some time under a different version https://www.npmjs.com/package/@aptos-labs/ts-sdk/v/1.33.0-sc.1 and @runtian-zhou mentioned receiving positive feedback from users, along with excitement about integrating it officially into the `ts-sdk`.

@WGB5445 validated that the feature works in both web and Node.js environments.

@0xmaayan confirmed there is no increase in the bundle size

current `main` branch
```
npm notice name: @aptos-labs/ts-sdk
npm notice version: 1.33.1
npm notice filename: aptos-labs-ts-sdk-1.33.1.tgz
npm notice package size: 1.1 MB
npm notice unpacked size: 6.38 MB
```

script compose branch
```
npm notice name: @aptos-labs/ts-sdk
npm notice version: 1.33.0-sc.1
npm notice filename: aptos-labs-ts-sdk-1.33.0-sc.1.tgz
npm notice package size: 1.0 MB
npm notice unpacked size: 6.4 MB
```

I’m okay with moving forward and merging it, thoughts @gregnazario @GhostWalker562 ?

@runtian-zhou One thing we’re missing, which could be beneficial, is adding a complete example in the [examples folder](https://github.com/aptos-labs/aptos-ts-sdk/tree/main/examples/typescript), not a blocker tho

0xmaayan:
> @0xmaayan would like to see it tested in a react-native environment too but otherwise sounds good to me

@runtian-zhou

WGB5445:
This is the GitHub Pages deployment page for the Next.js and React versions, where you can view the corresponding source code:

- [[React version](https://wgb5445.github.io/react-aptos-sc/)](https://wgb5445.github.io/react-aptos-sc/)
- [[Next.js version](https://wgb5445.github.io/nextjs-aptos-sc/)](https://wgb5445.github.io/nextjs-aptos-sc/)

lumos42:
We've been using it in testnet and mainnet for some time, and there's a feature request which could really optimize the script composer - the ability to provide ABIs manually. Currently it fetches for ABIs every time which slows down the process.

@itsmnthn can provide more details on this.

itsmnthn:
> We've been using it in testnet and mainnet for some time, and there's a feature request which could really optimize the script composer - the ability to provide ABIs manually. Currently it fetches for ABIs every time which slows down the process.
> 
> 
> 
> @itsmnthn can provide more details on this.

Yes passing ABIs to the addBatchedCall for the given call and for standard module's ABIs as part of the script composer to avoid fetching it over the network multiple times to speed up execution time especially when connection is slow. Things add up when the script is complex and has more steps even when the network speed is gud.

I've written custom AptosScriptComposer class to use in our FE which add abi as a parameter to addBatchedCall to avoid fetching ABI. So it does not fetch for a module which being called but still fetches standard abis required for argument types etc (which is not efficient right now especially to use on frontend)

src/transactions/scriptComposer/callArgument.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+export class callArgument {
```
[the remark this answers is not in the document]
runtian-zhou: Sry should have removed this file...

src/api/transactionSubmission/build.ts
```diff
@@ -96,6 +103,82 @@ export class Build {
...
+    const composer = new AptosScriptComposer(this.config);
+    await composer.init();
+    const builder = await args.builder(composer);
+    const bytes = builder.build();
+    const rawTxn = await generateRawTransaction({
+      aptosConfig: this.config,
+      payload: TransactionPayloadScript.load(new Deserializer(bytes)),
+      ...args,
+    });
+    return new SimpleTransaction(rawTxn, args.withFeePayer === true ? AccountAddress.ZERO : undefined);
```
0xmaayan, line 179: If well follow the same format, this will be taken care with https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/internal/transactionSubmission.ts#L165

src/transactions/scriptComposer/index.ts
```diff
@@ -0,0 +1,86 @@
...
+
+import { ScriptComposerWasm } from "@aptos-labs/script-composer-pack";
+import { AptosApiType } from "../../utils";
+import { AptosConfig } from "../../api/aptosConfig";
+import { InputBatchedFunctionData } from "../types";
+import { fetchMoveFunctionAbi, getFunctionParts, standardizeTypeTags } from "../transactionBuilder";
+import { CallArgument } from "../../types";
+import { convertCallArgument } from "../transactionBuilder/remoteAbi";
+
+// A wrapper class around TransactionComposer, which is a WASM library compiled
```
0xmaayan, line 12: Can we keep the same comment format? it will show up nicer in the IDE
```
/**
*
**/
```

src/transactions/scriptComposer/index.ts
```diff
@@ -0,0 +1,86 @@
...
+  constructor(aptosConfig: AptosConfig) {
+    this.config = aptosConfig;
+    this.builder = undefined;
+  }
+
+  // Initializing the wasm needed for the script composer, must be called
+  // before using the composer.
+  async init() {
+    if (!AptosScriptComposer.transactionComposer) {
+      const module = await import("@aptos-labs/script-composer-pack");
```
0xmaayan, line 33: was this tested in node/web (react+next)/react native envs?

src/transactions/scriptComposer/index.ts
```diff
@@ -0,0 +1,86 @@
...
+      const { TransactionComposer, initSync } = module;
+      if (!ScriptComposerWasm.isInitialized) {
+        ScriptComposerWasm.init();
+      }
+      initSync({ module: ScriptComposerWasm.wasm });
+      AptosScriptComposer.transactionComposer = TransactionComposer;
+    }
+    this.builder = AptosScriptComposer.transactionComposer.single_signer();
```
0xmaayan, line 41: lets keep the same format
```suggestion
    this.builder = AptosScriptComposer.transactionComposer.singleSigner();
```

src/transactions/scriptComposer/index.ts
```diff
@@ -0,0 +1,86 @@
...
+// from aptos-core/aptos-move/script-composer.
+//
+// This class allows the SDK caller to build a transaction that invokes multiple Move functions
+// and allow for arguments to be passed around.
+export class AptosScriptComposer {
+  private config: AptosConfig;
+
+  private builder?: any;
+
+  private static transactionComposer?: any;
```
0xmaayan, line 22: why `any`?

src/transactions/scriptComposer/index.ts
```diff
@@ -0,0 +1,86 @@
...
+      if (!ScriptComposerWasm.isInitialized) {
+        ScriptComposerWasm.init();
+      }
+      initSync({ module: ScriptComposerWasm.wasm });
+      AptosScriptComposer.transactionComposer = TransactionComposer;
+    }
+    this.builder = AptosScriptComposer.transactionComposer.single_signer();
+  }
+
+  // Add a move function invocation to the TransactionComposer.
```
0xmaayan, line 44: use `/** */` comment format

src/transactions/scriptComposer/index.ts
```diff
@@ -0,0 +1,86 @@
...
+  //
+  // Similar to how to create an entry function, the difference is that input arguments could
+  // either be a `CallArgument` which represents an abstract value returned from a previous Move call
+  // or the regular entry function arguments.
+  //
+  // The function would also return a list of `CallArgument` that can be passed on to future calls.
+  async addBatchedCalls(input: InputBatchedFunctionData): Promise<CallArgument[]> {
+    const { moduleAddress, moduleName, functionName } = getFunctionParts(input.function);
```
0xmaayan, line 52: we we rewrite the same flow?
https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/transactions/transactionBuilder/transactionBuilder.ts#L180

src/transactions/scriptComposer/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    return this.builder.add_batched_call(
+      `${moduleAddress}::${moduleName}`,
+      functionName,
+      typeArguments.map((arg) => arg.toString()),
+      functionArguments,
+    );
+  }
+
+  build(): Uint8Array {
+    return this.builder.generate_batched_calls(true);
```
0xmaayan, line 84:
```suggestion
    return this.builder.generateBatchedCalls(true);
```

src/transactions/scriptComposer/index.ts
runtian-zhou: Yes.

src/transactions/scriptComposer/index.ts
runtian-zhou: This is to allow for a lazy initialization of the composer so we don't need to pull the types in rightaway.

src/api/transactionSubmission/build.ts
```diff
@@ -96,6 +103,82 @@ export class Build {
...
+   * }
+   * runExample().catch(console.error);
+   * ```
+   */
+  async scriptComposer(args: {
+    sender: AccountAddressInput;
+    builder: (builder: AptosScriptComposer) => Promise<AptosScriptComposer>;
+    options?: InputGenerateTransactionOptions;
+    withFeePayer?: boolean;
+  }): Promise<SimpleTransaction> {
```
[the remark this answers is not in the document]
runtian-zhou: This is because the `generateTransaction` cannot take the serialized blob of move script as input (input contains both bytecode and type arguments/inputs).

0xmaayan:
Overall LGTM, most important thing is that the change is marked as `Breaking` on CHANGELOG - not sure why, but if it is breaking we need to go with a major bump release cc @GhostWalker562

src/api/transactionSubmission/build.ts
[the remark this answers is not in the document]
0xmaayan: Ideally, we would have `generateBatchTransaction`

runtian-zhou:
Oh this shouldn't be a breaking cause we didn't remove existing apis, just a new feature. Lemme remove that.

src/transactions/scriptComposer/index.ts
runtian-zhou: It's a bit unfortunate the camel cases are in the packed wasm so we can't just change the calls here. Tho these apis aren't called by users directly so it should be safe?

src/transactions/scriptComposer/index.ts
0xmaayan: then ideally the sdk would hold an internal type for that

src/transactions/scriptComposer/index.ts
runtian-zhou: This is because we have some extra bookkeeping logics to maintain in the composer? Was wondering if you have some suggestions for how to refactor the codes here.

0xmaayan (approved):
🚀 ship it!

PR #593, opened 2024-11-25
[secp256k1] Support inputed compressed public keys

gregnazario:
### Description
Support compressed public keys for secp256k1 natively in the SDK.

Looks like originally it only supported uncompressed keys, which may not be given to users.

### Test Plan
Check out the new test for how we test compression.

### Checklist


pnpm-lock.yaml
0xmaayan, line 1: need to revert

0xmaayan (approved):

pnpm-lock.yaml
gregnazario: Yep, already fixed

PR #594, opened 2024-11-25
Permissioned Accounts

chiumax:
### Description
big thanks to @kaw2k for the help + stubs

Best Practice Questions:
- Currently types are living under `internal/Permissions.ts`, should they be moved to `api/...ts`?
  - Based on the context, types should likely move to api/ since they are part of the public interface
  - This matches the pattern seen in api/permissions.ts which exposes the Permission types

- Should we remove the factories for each permissions or keep both types and factories?
  - Factories don't provide much value beyond type construction
  - Tests show direct object construction is just as clear
  - Consider deprecating factories in favor of just types

Points of Interest:
- Issue with returning arrays in `grantPermission`:
  - NFT permissions can have multiple capabilities (transfer, mutate)
  - Current workaround is handling each capability separately
  - Need to investigate why batched array returns break Move code
  - Consider refactoring NFT permissions into separate permission objects per capability
  - Gas vs APT permissions?

Architecture Changes @kaw2k:
- Moved core permission functionality from tests to internal/Permissions.ts
- Updated some of the syntax in the test cases
- Added proper abstraction layers:
  - api/permissions.ts - Public interface
  - internal/permissions.ts - Implementation
  - tests - Integration testing

Permission Types:
1. `FungibleAssetPermission`
2. `Gas Permission`
3. `NFT Permission`
4. `NFT Collection Permission`

Core Features:
- Request Permission
  - `FungibleAsset` ✅ 
  - `NFT` 🟠 
      - `mutate` ❌ 
      - 'transfer` ✅ 
  - `Gas` ❌ 
  - `NFTCollection` ❌ 

- Revoke Permission  
  - `FungibleAsset` ✅ 
  - `NFT` ✅ 
   - `Gas` ❌ 
  - `NFTCollection` ❌ 

- View Permission
  - `FungibleAsset` ✅ 
  - `NFT` 🟠  
        - `mutate` ❌ 
        - `transfer` ✅ 
   - `Gas` ❌ 
  - `NFTCollection` ❌ 

TODO:
- Implement NFT mutate capability
- Investigate batched permission returns
- Add proper error handling
- Consider deprecating permission factories

### Checklist

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,178 +1,4552 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
 
 settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-dependencies:
-  '@aptos-labs/aptos-cli':
-    specifier: ^0.2.0
-    version: 0.2.0
-  '@aptos-labs/aptos-client':
-    specifier: ^0.1.1
-    version: 0.1.1
-  '@noble/curves':
-    specifier: ^1.4.0
-    version: 1.4.0
-  '@noble/hashes':
-    specifier: ^1.4.0
-    version: 1.4.0
-  '@scure/bip32':
-    specifier: ^1.4.0
-    version: 1.4.0
-  '@scure/bip39':
-    specifier: ^1.3.0
-    version: 1.3.0
-  '@wgb5445/aptos-intent-npm':
-    specifier: ^0.0.12
-    version: 0.0.12
-  aptos-intent:
-    specifier: ^0.1.1
-    version: 0.1.1
-  eventemitter3:
-    specifier: ^5.0.1
-    version: 5.0.1
-  form-data:
-    specifier: ^4.0.0
-    version: 4.0.0
-  js-base64:
-    specifier: ^3.7.7
-    version: 3.7.7
-  jwt-decode:
-    specifier: ^4.0.0
-    version: 4.0.0
-  nodemon:
-    specifier: ^3.1.7
-    version: 3.1.7
-  poseidon-lite:
-    specifier: ^0.2.0
-    version: 0.2.0
+importers:
 
-devDependencies:
-  '@babel/traverse':
-    specifier: ^7.25.7
-    version: 7.25.7
-  '@cucumber/cucumber':
-    specifier: ^11.0.1
-    version: 11.0.1
-  '@graphql-codegen/cli':
-    specifier: ^5.0.2
-    version: 5.0.2(@types/node@20.16.10)(graphql@16.9.0)(typescript@5.6.2)
-  '@graphql-codegen/import-types-preset':
-    specifier: ^3.0.0
-    version: 3.0.0(graphql@16.9.0)
-  '@graphql-codegen/typescript':
-    specifier: ^4.0.9
-    version: 4.0.9(graphql@16.9.0)
-  '@graphql-codegen/typescript-graphql-request':
-    specifier: ^6.2.0
-    version: 6.2.0(graphql-request@6.1.0)(graphql-tag@2.12.6)(graphql@16.9.0)
-  '@graphql-codegen/typescript-operations':
-    specifier: ^4.2.3
-    version: 4.2.3(graphql@16.9.0)
-  '@types/base-64':
-    specifier: ^1.0.2
-    version: 1.0.2
-  '@types/jest':
-    specifier: ^29.5.13
-    version: 29.5.13
-  '@types/jsonwebtoken':
-    specifier: ^9.0.7
-    version: 9.0.7
-  '@types/node':
-    specifier: ^20.16.10
-    version: 20.16.10
-  '@typescript-eslint/eslint-plugin':
-    specifier: ^6.21.0
-    version: 6.21.0(@typescript-eslint/parser@6.21.0)(eslint@8.57.1)(typescript@5.6.2)
-  '@typescript-eslint/parser':
-    specifier: ^6.21.0
-    version: 6.21.0(eslint@8.57.1)(typescript@5.6.2)
-  dotenv:
-    specifier: ^16.4.5
-    version: 16.4.5
-  eslint:
-    specifier: ^8.57.1
-    version: 8.57.1
-  eslint-config-airbnb-base:
-    specifier: ^15.0.0
-    version: 15.0.0(eslint-plugin-import@2.31.0)(eslint@8.57.1)
-  eslint-config-airbnb-typescript:
-    specifier: ^17.1.0
-    version: 17.1.0(@typescript-eslint/eslint-plugin@6.21.0)(@typescript-eslint/parser@6.21.0)(eslint-plugin-import@2.31.0)(eslint@8.57.1)
-  eslint-config-prettier:
-    specifier: ^9.1.0
-    version: 9.1.0(eslint@8.57.1)
-  eslint-plugin-import:
-    specifier: ^2.31.0
-    version: 2.31.0(@typescript-eslint/parser@6.21.0)(eslint@8.57.1)
-  graphql:
-    specifier: ^16.9.0
-    version: 16.9.0
-  graphql-request:
-    specifier: ^6.1.0
-    version: 6.1.0(graphql@16.9.0)
-  jest:
-    specifier: ^29.7.0
-    version: 29.7.0(@types/node@20.16.10)(ts-node@10.9.2)
-  jsonwebtoken:
-    specifier: ^9.0.2
-    version: 9.0.2
-  prettier:
-    specifier: ^3.3.3
-    version: 3.3.3
-  tree-kill:
-    specifier: ^1.2.2
-    version: 1.2.2
-  ts-jest:
-    specifier: ^29.2.5
-    version: 29.2.5(@babel/core@7.24.6)(esbuild@0.23.1)(jest@29.7.0)(typescript@5.6.2)
-  ts-loader:
-    specifier: ^9.5.1
-    version: 9.5.1(typescript@5.6.2)(webpack@5.94.0)
-  ts-node:
-    specifier: ^10.9.2
-    version: 10.9.2(@types/node@20.16.10)(typescript@5.6.2)
-  tsup:
-    specifier: ^8.3.0
-    version: 8.3.0(typescript@5.6.2)
-  typedoc:
-    specifier: ^0.26.8
-    version: 0.26.8(typescript@5.6.2)
-  typescript:
-    specifier: ^5.6.2
-    version: 5.6.2
+  .:
+    dependencies:
+      '@aptos-labs/aptos-cli':
+        specifier: ^0.2.0
+        version: 0.2.0
+      '@aptos-labs/aptos-client':
+        specifier: ^0.1.1
+        version: 0.1.1
+      '@noble/curves':
+        specifier: ^1.4.0
+        version: 1.4.0
+      '@noble/hashes':
+        specifier: ^1.4.0
+        version: 1.4.0
+      '@scure/bip32':
+        specifier: ^1.4.0
+        version: 1.4.0
+      '@scure/bip39':
+        specifier: ^1.3.0
+        version: 1.3.0
+      '@wgb5445/aptos-intent-npm':
+        specifier: ^0.0.12
+        version: 0.0.12
+      aptos-intent:
+        specifier: ^0.1.1
+        version: 0.1.1
+      eventemitter3:
+        specifier: ^5.0.1
+        version: 5.0.1
+      form-data:
+        specifier: ^4.0.0
+        version: 4.0.0
+      js-base64:
+        specifier: ^3.7.7
+        version: 3.7.7
+      jwt-decode:
+        specifier: ^4.0.0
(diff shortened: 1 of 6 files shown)


src/internal/permissions.ts
```diff
@@ -0,0 +1,429 @@
...
+        function: "0x1::fungible_asset::grant_permission",
+        functionArguments: [
+          BatchArgument.new_signer(0),
+          args.permissionedSigner[0].borrow(),
+          args.permission.asset, // do i need to convert this to AccountAddress? .... i guess not??
+          args.permission.balance,
+        ],
+        typeArguments: [],
+      });
+    case PermissionType.NFT: {
```
chiumax, line 326: returning an array here breaks

src/internal/permissions.ts
```diff
@@ -0,0 +1,429 @@
...
+  type: Exclude<string, PermissionType | `${PermissionType}`>;
+  [key: string]: string | Exclude<string, PermissionType | `${PermissionType}`>;
+}
+
+// revoke permission types
+export type RevokeFungibleAssetPermission = Pick<FungibleAssetPermission, "type" | "asset">;
+export type RevokeNFTPermission = Pick<NFTPermission, "type" | "assetAddress">;
+
+// factories -- on second thought, i think these are dumb
+export function FungibleAssetPermission(args: Omit<FungibleAssetPermission, "type">): FungibleAssetPermission {
```
chiumax, line 92: remove these?

src/internal/permissions.ts
```diff
@@ -0,0 +1,429 @@
...
+  return { type: PermissionType.NFTCollection, ...args };
+}
+
+// revoke factories
+export function RevokeFungibleAssetPermission(args: Omit<RevokeFungibleAssetPermission, "type">): RevokeFungibleAssetPermission {
+  return { type: PermissionType.FungibleAsset, ...args };
+}
+export function RevokeNFTPermission(args: Omit<RevokeNFTPermission, "type">): RevokeNFTPermission {
+  return { type: PermissionType.NFT, ...args };
+}
```
kaw2k, line 114: The main draw here is improved type inference. To get a feel for how it would work without the factories, try consuming `requestPermissions` without using factories. You open a new object and are left without guidance on what the first step is. A brand new consumer of the API might not find it readily apparent that all the properties are defined based on what `type` is specified.

Might be good to brainstorm other possible DX around this.

src/internal/permissions.ts
```diff
@@ -0,0 +1,429 @@
...
+  filter
+}: {
+  aptosConfig: AptosConfig;
+  primaryAccount: SingleKeyAccount;
+  subAccount: AbstractedEd25519Account;
+  filter?: T;
+}): Promise<FilteredPermissions<T>> {
+  const handle = await getHandleAddress({ aptosConfig, primaryAccount, subAccount });
+
+  const res = await fetch(`http://127.0.0.1:8080/v1/accounts/${handle}/resources`);
```
kaw2k, line 135: Eventually we will want to use the passed in node for this fetch.

src/internal/permissions.ts
```diff
@@ -0,0 +1,429 @@
...
+        }>;
+      };
+    };
+  }>;
+
+  const data = (await res.json()) as NodeDataResponse;
+
+  const permissions = data[0].data.perms.data.map((d) => {
+    switch (d.key.type_name) {
+      case "0x1::fungible_asset::WithdrawPermission": // can this be better? i dont rly like this
```
kaw2k, line 153: Replace with the capability enum. What don't you like about it? The hardcoded string? The potential for a large `switch`?

src/internal/permissions.ts
```diff
@@ -0,0 +1,429 @@
...
+        return NFTPermission({
+          assetAddress: d.key.data,
+          capabilities: { transfer: true, mutate: false }
+        });
+      default:
+        return GenericPermission({
+          type: d.key.type_name,
+          data: d.key.data,
+          value: d.value
+        });
```
kaw2k, line 168: We shouldn't need the generic case after we have all the capabilities outlined.

src/internal/permissions.ts
```diff
@@ -0,0 +1,429 @@
...
+}
+export async function getPermissions<T extends PermissionType>({
+  aptosConfig,
+  primaryAccount,
+  subAccount,
+  filter
+}: {
+  aptosConfig: AptosConfig;
+  primaryAccount: SingleKeyAccount;
+  subAccount: AbstractedEd25519Account;
```
kaw2k, line 130: Thinking about this for a moment, instead of `subAccount` we would want the public key of the sub account. Dapps will use this SDK as well as wallets, and wallets most likely won't be saving the private keys of the sub account.

src/internal/permissions.ts
```diff
@@ -0,0 +1,418 @@
...
+      });
+    case PermissionType.NFT: {
+      let txn: Promise<BatchArgument[]>[] = [];
+      const nftPermission = args.permission as NFTPermission; // bruh why do i need this
+      if (nftPermission.capabilities.transfer) {
+        // console.log("grant transfer");
+        // txn.push(builder.add_batched_calls({
+        //   function: "0x1::object::grant_permission",
+        //   functionArguments: [
+        //     BatchArgument.new_signer(0),
```
kaw2k, line 328: bruh

kaw2k: I _think_ the types are getting messed up here due to `GenericPermission`. The `type` field of `GenericPermission` is `string` which also matches at the type level.

src/internal/permissions.ts
```diff
@@ -0,0 +1,418 @@
...
+      return builder;
+    },
+  });
+}
+
+export async function revokePermissions(args: {
+  aptosConfig: AptosConfig;
+  primaryAccount: SingleKeyAccount;
+  subAccount: AbstractedEd25519Account;
+  permissions: RevokePermission[];
```
kaw2k, line 226: Without putting any thought into this, something feels off here. Need to dig a little.

src/internal/permissions.ts
```diff
@@ -0,0 +1,418 @@
...
+}): Promise<SimpleTransaction> {
+  const { aptosConfig, primaryAccount, subAccount, permissions } = args;
+
+  const aptos = new Aptos(aptosConfig);
+  const transaction = await aptos.transaction.build.batched_intents({
```
kaw2k, line 231: We need to change this to be the public key as well.

src/internal/permissions.ts
```diff
@@ -0,0 +1,418 @@
...
+export async function getHandleAddress({
+  aptosConfig,
+  primaryAccount,
+  subAccount,
+}: {
+  aptosConfig: AptosConfig;
+  primaryAccount: SingleKeyAccount;
+  subAccount: AbstractedEd25519Account;
+}) {
+  const aptos = new Aptos(aptosConfig);
```
kaw2k, line 405: And here.

src/internal/permissions.ts
```diff
@@ -0,0 +1,418 @@
...
+}
+
+export enum MoveVMPermissionType {
+  FungibleAsset = "0x1::fungible_asset::WithdrawPermission",
+  
+}
+
+export interface FungibleAssetPermission {
+  type: PermissionType.FungibleAsset;
+  asset: string;
```
kaw2k, line 58: Let's discuss about the two types of fungible permissions (primary store vs fixed amount)

src/internal/permissions.ts
chiumax: should factories be moved to api file?

chiumax: but good point actually

src/internal/permissions.ts
chiumax: yea feels hardcoded but NBD if no way around it

src/internal/permissions.ts
kaw2k: Since they share the same name as the interface, I think having them both be here is fine.

src/internal/permissions.ts
```diff
@@ -0,0 +1,429 @@
...
+  FungibleAsset = "FungibleAsset",
+  Gas = "Gas",
+  NFT = "NFT",
+  NFTCollection = "Collection",
+}
+
+export enum MoveVMPermissionType {
+  FungibleAsset = "0x1::fungible_asset::WithdrawPermission",
+  
+}
```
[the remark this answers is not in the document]
chiumax: rename the type, use the enum

src/internal/permissions.ts
chiumax: @hardsetting to give some input here

src/internal/permissions.ts
chiumax: could be a little odd having union of `Permission` but thought process is just passing in the whole object since it has props needed. Depends on how dApps manage this information.

src/internal/permissions.ts
chiumax: with regards to subaccount

src/internal/permissions.ts
chiumax: decision is to leave them here for types

src/internal/permissions.ts
hardsetting: I'm surprised that typing `{ type: 'permission1', ... }` would not provide autocompletion strictly for `Permission1`.
That's the point of discriminated unions. Could it be the IDE not working correctly? lemme try a couple ones

hardsetting: This is with VSCode

Screenshot 2024-12-04 at 12 19 16 PM

hardsetting: But if you want to use factories go for it, I would maybe use some different naming like
```typescript
export function makeFungibleAssetPermission(...);
export function makeGasPermission(...);
```

they're still functions after all

hardsetting: what's the point of the "revoke" factories?

hardsetting: also try to split the code in different files / folders.
types and factories are a good candidate for different files

src/internal/permissions.ts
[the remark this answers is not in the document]
hardsetting: This doesn't seem to be used at the moment.
What is it for? Can we removed it if not needed?

src/internal/permissions.ts
```diff
@@ -0,0 +1,429 @@
...
+
+export enum NFTCollectionCapability {
+  transfer = "transfer",
+  mutate = "mutate",
+}
+
+export interface GenericPermission {
+  type: Exclude<string, PermissionType | `${PermissionType}`>;
+  [key: string]: string | Exclude<string, PermissionType | `${PermissionType}`>;
+}
```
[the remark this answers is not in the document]
hardsetting: Agreed, is this needed? Maybe we could remove

src/internal/permissions.ts
```diff
@@ -0,0 +1,429 @@
...
+  return { type: PermissionType.FungibleAsset, ...args };
+}
+export function RevokeNFTPermission(args: Omit<RevokeNFTPermission, "type">): RevokeNFTPermission {
+  return { type: PermissionType.NFT, ...args };
+}
+
+// mostly for testing
+export function GenericPermission(args: GenericPermission): GenericPermission {
+  return {...args}
+}
```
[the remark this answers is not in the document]
hardsetting: remove it if "generic" is never used

src/internal/permissions.ts
chiumax: fixed by parsing out complicated nft permission

chiumax: todo: send example of this breaking

src/internal/permissions.ts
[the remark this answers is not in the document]
chiumax: used in `getPermissions` move level permissions have their own strings outside of "FungibleAsset" or "NFT"

src/internal/permissions.ts
chiumax: information required for revoking is a subset of what's needed for granting permission so it's just a builder for a smaller object

pnpm-lock.yaml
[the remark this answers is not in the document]
chiumax: ok will look into this

src/internal/permissions.ts
```diff
@@ -0,0 +1,418 @@
...
+    type: string;
+    data: {
+      perms: {
+        data: Array<{
+          key: { data: string; type_name: string };
+          value: string;
+        }>;
+      };
+    };
+  }>;
```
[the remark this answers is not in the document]
chiumax: you're right, move resource is a type: https://gitkraken.dev/link/dnNjb2RlOi8vZWFtb2Rpby5naXRsZW5zL2xpbmsvci8zNDIzNGQ5MDZiMzYyZjExYWYxOGIyYWQ3Mjc4YmYwNGFjM2E5NzdkL2Yvc3JjL3R5cGVzL2luZGV4LnRzP3VybD1naXQlNDBnaXRodWIuY29tJTNBYXB0b3MtbGFicyUyRmFwdG9zLXRzLXNkay5naXQmbGluZXM9MzMw?origin=gitlens

kaw2k (approved):
Lets merge, we can address lingering concerns in a future PR. The shape of things should be stable.

PR #600, opened 2024-12-05
Update fa_coin and your_fungible_asset

jmintuitive:
### Description
Updates the example code used in the "Your First Fungible Asset" tutorial in order to make it customizable. That way during the updated tutorial, you can change the name of the asset and see what the result is on-chain. 

### Test Plan

I tested locally to verify that the fa_coin still deployed and executed transactions correctly, but now the updated `your_fungible_asset` script actually looks up the `ASSET_NAME`. 

### Checklist
  - [x] Have you updated the `CHANGELOG.md`? (Not needed)

### DIFF
--- a/examples/typescript/your_fungible_asset.ts
+++ b/examples/typescript/your_fungible_asset.ts
@@ -10,6 +10,7 @@ import {
   InputViewFunctionData,
   Network,
   NetworkToNetworkName,
+  MoveValue
 } from "@aptos-labs/ts-sdk";
 import { compilePackage, getPackageBytesToPublish } from "./utils";
 /**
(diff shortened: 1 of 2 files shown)


examples/typescript/your_fungible_asset.ts
```diff
@@ -80,6 +81,34 @@ async function burnCoin(admin: Account, fromAddress: AccountAddress, amount: Any
...
 
+/**
+ * Fetches the name of the fungible asset created by the given account.
+ * This function interacts with the `get_name` view function in the Move module.
+ *
+ * @param aptos - An instance of the Aptos client.
+ * @param admin - The account that deployed the fungible asset module.
+ * @returns The name of the token as a string.
+ */
+async function fetchTokenName(aptos: Aptos, admin: Account): Promise<string> {
```
gregnazario, line 92: These examples use a "global" aptos object, so it doesn't have to be passed here

examples/typescript/your_fungible_asset.ts
```diff
@@ -175,28 +209,29 @@ async function main() {
...
+    `Charlie's updated ${tokenName} primary fungible store balance: ${await getFaBalance(charlie, metadataAddress)}.`,
   );
 
   console.log("Alice freezes Bob's account.");
   const freezeTransactionHash = await freeze(alice, bob.accountAddress);
   await aptos.waitForTransaction({ transactionHash: freezeTransactionHash });
 
   console.log(
-    "Alice as the admin forcefully transfers the newly minted coins of Charlie to Bob ignoring that Bob's account is frozen.",
+    `Alice as the admin forcefully transfers the newly minted coins of Charlie to Bob ignoring that Bob's account is frozen.`,
```
gregnazario, line 230: please replace with `"`

gregnazario (approved):
LGTM, just clean up those two issues

PR #596, opened 2024-11-27
Sub accounts

chiumax:
### Checklist
  - [ ] Have you ran `pnpm fmt`?
  - [ ] Have you updated the `CHANGELOG.md`?

PR #604, opened 2024-12-12
Converts permissions to a serializable class

kaw2k:
### Description
There will need to be more work here, for example I didn't convert the revoke permissions to be a serializable class. Looking to get feedback. 

### Test Plan
Tests pass 

### Related Links
Base PR #596

### DIFF
--- a/src/internal/permissions.ts
+++ b/src/internal/permissions.ts
@@ -20,8 +20,8 @@ import {
   MoveVMPermissionType,
   RevokePermission,
   Permission,
-  buildFungibleAssetPermission,
-  buildNFTPermission,
+  FungibleAssetPermission,
+  NFTPermission,
 } from "../types/permissions";
 import { Transaction } from "../api/transaction";
 import { view } from "./view";
@@ -59,13 +59,13 @@ export async function getPermissions<T extends PermissionType>({
   const permissions = data[0].data.perms.data.map((d) => {
     switch (d.key.type_name) {
       case MoveVMPermissionType.FungibleAsset: // can this be better? i dont rly like this
-        return buildFungibleAssetPermission({
-          asset: d.key.data,
+        return FungibleAssetPermission.from({
+          asset: AccountAddress.fromString(d.key.data),
           balance: d.value,
         });
       case MoveVMPermissionType.TransferPermission:
-        return buildNFTPermission({
-          assetAddress: d.key.data,
+        return NFTPermission.from({
+          assetAddress: AccountAddress.fromString(d.key.data),
           capabilities: { transfer: true, mutate: false },
         });
       default:
@@ -108,21 +108,24 @@ export async function requestPermission(args: {
 
       // if nft permission has multiple capabilities, we need to add multiple txns
       // For NFT permissions with multiple capabilities, split into separate transactions
-
       const expandedPermissions = permissions.flatMap((permission) => {
         if (permission.type === PermissionType.NFT && permission.capabilities) {
-          const expanded = [];
+          const expanded: Permission[] = [];
           if (permission.capabilities.transfer) {
-            expanded.push({
-              ...permission,
-              capabilities: { transfer: true, mutate: false },
-            });
+            expanded.push(
+              NFTPermission.from({
+                assetAddress: permission.assetAddress,
+                capabilities: { transfer: true, mutate: false },
+              }),
+            );
           }
           if (permission.capabilities.mutate) {
-            expanded.push({
-              ...permission,
-              capabilities: { transfer: false, mutate: true },
-            });
+            expanded.push(
+              NFTPermission.from({
+                assetAddress: permission.assetAddress,
+                capabilities: { transfer: false, mutate: true },
+              }),
+            );
           }
           return expanded;
         }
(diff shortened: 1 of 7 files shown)


src/types/permissions.ts
```diff
@@ -42,33 +50,151 @@ export enum PermissionType {
...
+
+  constructor(asset: AccountAddress, balance: string) {
+    super();
+    this.asset = asset;
+    this.balance = balance;
+  }
+
+  static from({ asset, balance }: { asset: AccountAddress; balance: string }): FungibleAssetPermission {
+    return new FungibleAssetPermission(asset, balance);
+  }
```
kaw2k, line 68: I prefer the single object as a parameter version, which do you all prefer?

src/types/permissions.ts
```diff
@@ -42,33 +50,151 @@ export enum PermissionType {
...
+  serialize(serializer: Serializer): void {
+    serializer.serializeStr(this.type);
+    this.assetAddress.serialize(serializer);
+
+    const [capabilityKeys, capabilityValues] = Object.entries(this.capabilities).reduce(
+      ([keys, values], [key, value]) => [keys.concat(key), values.concat(value)],
+      [[] as string[], [] as boolean[]],
+    );
+    serializer.serializeStr(JSON.stringify(capabilityKeys));
+    serializer.serializeStr(JSON.stringify(capabilityValues));
```
kaw2k, line 138: Serializing the capabilities is a bit messy. The base type is:

```
capabilities: {
mutate: true,
transfer: false
}
``` 
where each property is optional. Let me know of a better way.

src/types/permissions.ts
```diff
@@ -42,33 +50,151 @@ export enum PermissionType {
...
+  constructor(amount: number) {
+    super();
+    this.amount = amount;
+  }
+
+  static from = ({ amount }: { amount: number }): GasPermission => new GasPermission(amount);
+
+  serialize(serializer: Serializer): void {
+    serializer.serializeStr(this.type);
+    serializer.serializeU16(this.amount);
```
kaw2k, line 98: What is the appropriate type for `amount`?

src/types/permissions.ts
```diff
@@ -42,33 +51,151 @@ export enum PermissionType {
...
+
+  constructor(asset: AccountAddress, balance: string) {
+    super();
+    this.asset = asset;
+    this.balance = balance;
+  }
+
+  static from({ asset, balance }: { asset: AccountAddress; balance: string }): FungibleAssetPermission {
+    return new FungibleAssetPermission(asset, balance);
+  }
```
hardsetting: Why not just using the constructor?

src/types/permissions.ts
hardsetting: Yeah agreed, but it's a bit annoying with classes cause you have to write more code

src/types/permissions.ts
```diff
@@ -42,33 +51,151 @@ export enum PermissionType {
...
+
+  serialize(serializer: Serializer): void {
+    serializer.serializeStr(this.type);
+    this.asset.serialize(serializer);
+    serializer.serializeStr(this.balance);
+  }
+
+  static deserialize(deserializer: Deserializer): FungibleAssetPermission {
+    deserializer.deserializeStr() as PermissionType.FungibleAsset; // type, is this needed?;
```
hardsetting, line 78: this is not getting assigned to anything

hardsetting (commented):
What we've been doing so far, to mimic rust enums, is the following pattern:

```typescript
export abstract class Permission {
  static deserialize(deserializer: Deserializer) {
    const type = deserializer.deserializeStr() as PermissionType;
    switch (type) {
      case PermissionType.FungibleAsset: return FungibleAssetPermission.load(deserializer);
      ...
      default: throw new Error('Unsupported permission type');
    }
  }
}

export class FungibleAssetPermission extends Permission {
  serialize(serializer: Serializer) { ... }
  load(deserializer: Deserializer) { ... }
}
```

src/types/permissions.ts
kaw2k: I was mirroring other classes in the codebase that have `from` as a static constructor.

kaw2k:
These types don't mirror the rust types. The classes have helped clean up the code in a few ways, a little complications in other. Pushing another update and calling out some other changes.

src/types/permissions.ts
```diff
@@ -42,70 +54,138 @@ export enum PermissionType {
...
+    super();
+    this.collectionAddress = collectionAddress;
+    this.capabilities = capabilities;
+  }
 
-/**
- * Revoke permission types
- */
-export type RevokeFungibleAssetPermission = Pick<FungibleAssetPermission, "type" | "asset">;
-export type RevokeNFTPermission = Pick<NFTPermission, "type" | "assetAddress">;
```
kaw2k, line 78: Note: I removed the revoking types here in favor of just using the base class. Might add these back in but modified after using the new style in connect.

src/internal/permissions.ts
```diff
@@ -36,8 +28,8 @@ export async function getPermissions<T extends PermissionType>({
   aptosConfig: AptosConfig;
   primaryAccountAddress: AccountAddress;
   subAccountPublicKey: Ed25519PublicKey;
-  filter?: T;
-}): Promise<FilteredPermissions<T>> {
+  filter?: new (...a: any) => T;
```
kaw2k, line 31: Trying something different with the filtering. Eventually we will hit the indexer and we can move away from calling the node directly. 

This takes the class itself, so for example `FungibleAssetPermission` and uses `instanceof` for filtering.

src/types/permissions.ts
chiumax: number works
gas implementation pr: https://github.com/aptos-labs/aptos-core/pull/15232/files

chiumax (approved):
🚢

PR #615, opened 2025-01-09
fix links

wmjae:


README.md
```diff
@@ -89,7 +89,7 @@ const account = await Account.fromPrivateKey({ privateKey });
 
 // Also, can use this function that resolves the provided private key type and derives the public key from it
 // to support key rotation and differentiation between Legacy Ed25519 and Unified authentications
-// Read more https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/api/account.ts#L364
+// Read more https://github.com/aptos-labs/aptos-ts-sdk/blob/v0.0.1/src/api/account.ts#L377
```
gregnazario, line 92: Hi, please make sure that these links point to `main` or just remove the link altogether, as `v0.0.1` is the oldest version.

wmjae: OK, I will remove the link later.
Btw, pointing to the main will cause the same issue again in the future. As  the main branch code will be changed again then the line of this link will incorrect again.

gregnazario: Yeah, that's why I figured removing the link is probably the best bet, it'll always be wrong

wmjae: Done, I have removed the link. :)

gregnazario (approved):
thanks for your contribution!

PR #619, opened 2025-01-17
fix keyless end-to-end mainnet test to actually wait for funded account

alinush:
### Description

The _old_ code did not properly wait for the testing account to be funded when the account did not exist. (Not sure if it correctly handled the account existing but with balance 0.)

The **new** code should now properly handle both cases: i.e., the `getAccountAPTAmount ` function returning zero for both a non-existent account and an existing account.

We now wait in a `while` loop until the account has enough funds to pay the gas and cover the transferred amount (even though paying to self).

(Related to https://github.com/aptos-labs/aptos-ts-sdk/pull/618 which properly documents the behavior of the `getAccountAPTAmount` function.)

### Test Plan

This is an end-to-end test. So I manually-tested the test, lol.

### Related Links

None.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
 - Add `AccountUtils` class to help with account serialization and deserialization
 - Add `SingleKeySigner` interface which adds the ability to get the `AnyPublicKey` from a `SingleKeyAccount`
 - We now throw an error earlier when you try to use the faucet with testnet or mainnet, rather than letting the call happen and then fail later.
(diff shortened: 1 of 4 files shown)


alinush:
The lints are failing:

```
Error:   80:22  error  Unexpected `await` inside a loop  no-await-in-loop
```

@GhostWalker562, not a TypeScript expert, but why can't I await in a loop? How can I implement something equivalent otherwise?

alinush:
Converted to draft, since I am trying to fix the keyless unit tests as well... since https://github.com/aptos-labs/aptos-core/pull/15704 broke them by removing the default VK.

heliuchuan (approved):

PR #621, opened 2025-01-24
Half the default max gas amount to 100000

heliuchuan:
### Description

There have been many users that report insufficient funds for txns.  Users report seemingly having enough APT, usually some amount like .15 APT.  By default, you need to have minimum .2 APT to send **any** txn to cover the potential max gas charge.  This is certainly too high for 99% of txns.

With this change, by default you need .1 APT.

We should consider half-ing this again to 50000 (or lower to 25000) to by default require .05 APT.

### Checklist


gregnazario:
As I mentioned earlier, this should be configurable

PR #624, opened 2025-02-03
[view] Add ability to get BCS output from view functions

gregnazario:
### Description
This is experimental for now, and will be moved into the main SDK at a later time.

### Related Links
Dependent on https://github.com/aptos-labs/aptos-client/pull/2

### Checklist

### DIFF
--- a/src/client/core.ts
+++ b/src/client/core.ts
@@ -15,6 +15,7 @@ import { AptosApiError } from "../errors";
  * @param options.method - The HTTP method to use for the request.
  * @param options.body - The body of the request.
  * @param options.contentType - The content type of the request.
+ * @param options.acceptType - The accept type of the request.
  * @param options.params - The query parameters to include in the request.
  * @param options.overrides - Additional overrides for the request.
  * @param options.overrides.HEADERS - Custom headers to include in the request.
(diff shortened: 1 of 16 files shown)


src/internal/view.ts
```diff
@@ -27,7 +27,7 @@ export async function view<T extends Array<MoveValue> = Array<MoveValue>>(args:
   viewFunctionPayload.serialize(serializer);
   const bytes = serializer.toUint8Array();
 
-  const { data } = await postAptosFullNode<Uint8Array, MoveValue[]>({
+  const { data } = await postAptosFullNode<T, any>({
```
0xmaayan, line 30: hmmmm we should really not use any type here

src/api/general.ts
```diff
@@ -200,11 +200,36 @@ export class General {
    */
   async view<T extends Array<MoveValue>>(args: {
     payload: InputViewFunctionData;
-    options?: LedgerVersionArg;
+    options?: LedgerVersionArg & { bcsOutput?: boolean };
```
0xmaayan, line 203: either we remove it here, or we need to handle it in the `view` function - I think we should just remove it from here

0xmaayan (commented):
If it is an experimental version, we should publish it under a tagged version

src/api/general.ts
gregnazario: Yeah, let me go back and fix these

src/client/core.ts
```diff
@@ -57,6 +58,62 @@ export async function request<Req, Res>(options: ClientRequest<Req>, client: Cli
...
+ *
+ * @returns The response from the request.
+ * @group Implementation
+ * @category Client
+ */
+export async function binaryRequest<Req>(options: ClientRequest<Req>, client: Client): Promise<ClientResponse<Buffer>> {
+  if (client.binaryProvider === undefined) {
+    throw new Error("Binary provider is not provided");
+  }
+  const { url, method, body, contentType, acceptType, params, overrides, originMethod } = options;
```
0xmaayan, line 86: can make a global `getRequestConfig()` or something like that,
then in the `binaryRequest` function, we just need to add the `"accept": acceptType ?? MimeType.BCS,` header

src/client/core.ts
```diff
@@ -118,3 +175,66 @@ export async function aptosRequest<Req extends {}, Res extends {}>(
...
+ *
+ * @param aptosRequestOpts - Options for the Aptos request, including the URL and path.
+ * @param aptosConfig - The configuration information for the SDK client instance.
+ * @param apiType - The type of API being accessed, which determines how the response is handled.
+ * @returns The response from the API request or throws an AptosApiError if the request fails.
+ * @group Implementation
+ * @category Client
+ * @experimental
+ */
+export async function aptosBinaryRequest<Req extends {}>(
```
0xmaayan, line 190: a repeated code

src/api/general.ts
0xmaayan: I dont think we do something with the `bcsOutput`?

gregnazario: woops, thought I removed this

0xmaayan (approved):

PR #627, opened 2025-02-06
Add MultiEd25519 account support

heliuchuan:
### Test Plan

Added unit and e2e tests
### Checklist


src/core/crypto/multiKey.ts
```diff
@@ -23,6 +23,79 @@ function bitCount(byte: number) {
 }
 /* eslint-enable no-bitwise */
 
+export abstract class AbstractMultiKey extends AccountPublicKey {
```
hardsetting, line 26: Nice!

src/account/MultiEd25519Account.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  readonly signingScheme = SigningScheme.MultiEd25519;
+
+  /**
+   * The signers used to sign messages.  These signers should correspond to public keys in the
+   * MultiEd25519Account.  The number of signers should be equal to this.publicKey.threshold.
+   * @group Implementation
+   * @category Account (On-Chain Model)
+   */
+  readonly signers: Ed25519PrivateKey[];
```
hardsetting, line 41: just wondering, why "signers" as opposed to "privateKeys"?

tests/unit/account.test.ts
```diff
@@ -165,7 +167,7 @@ describe("Account", () => {
     });
   });
 
-  describe("sign and verify", () => {
+  describe.only("sign and verify", () => {
```
hardsetting, line 170: Why is this needed?

hardsetting (approved):
Looks great!

tests/unit/account.test.ts
heliuchuan: oops

src/account/MultiEd25519Account.ts
heliuchuan: I think its better to describe functionality.  In theory I might want to change the type to something that signs rather than privateKey.

PR #649, opened 2025-02-26
[enums] Add partial support for enum types

GhostWalker562:
### Description
Add some support for skipping types when parsing structs. This sets up support for enums with the limited information given.

If the struct is unknown and `options.allowUnknownStructs` is `true`, we will just let the struct be `FixedBytes`
If the struct is known through `moduleAbi` and is fieldless, we will consider it an enum and the struct will be `FixedBytes`
If the struct is a framework know enum (like `DelegationKey` or `RateLimiter`), the struct will be `FixedBytes`

### Test Plan
- [x] Tested locally to ensure that Enum function arguments were passed and parsed properly.

### Related Links
N/A

### Checklist


gregnazario (approved):
Seems fine, but keep in mind, you could have structs that are not defined in the current module, since you are using it for any public function.

This will probably need to be reworked to be able to make outbound calls if the struct type doesn't match the module

PR #664, opened 2025-03-21
[ts-sdk] add a helper function to return the paired FA address

lightmark:
### Description
add a helper function `pairedFaMetadataAddress` to return the paired fa address for even non-standard coin type string.

### Test Plan
unit test.

### Checklist

### DIFF
--- a/src/core/accountAddress.ts
+++ b/src/core/accountAddress.ts
@@ -184,6 +184,31 @@ export class AccountAddress extends Serializable implements TransactionArgument 
   }
 
   /**
+   * Convert the account address to a string in SHORT format, which is 0x followed by the shortest
+   * possible representation (no leading zeros).
+   *
+   * @returns AccountAddress as a string in SHORT form.
+   * @group Implementation
+   * @category Serialization
+   */
+  toStringShort(): `0x${string}` {
+    return `0x${this.toStringShortWithoutPrefix()}`;
+  }
+
+  /**
+   * Returns a lossless short string representation of the address by trimming leading zeros.
+   * If the address consists of all zeros, returns "0".
+   *
+   * @returns A string representation of the address without leading zeros
+   * @group Implementation
+   * @category Serialization
+   */
+  toStringShortWithoutPrefix(): string {
+    const hex = bytesToHex(this.data).replace(/^0+/, "");
+    return hex === "" ? "0" : hex;
+  }
+
+  /**
    * Get the inner data as a Uint8Array.
    * The inner data is already a Uint8Array, so no conversion takes place.
    *
--- a/src/utils/helpers.ts
+++ b/src/utils/helpers.ts
@@ -4,6 +4,7 @@
 import { decode } from "js-base64";
 import { MoveFunctionId, MoveStructId } from "../types";
 import { AccountAddress } from "../core/accountAddress";
+import { createObjectAddress } from "../core/account/utils/address";
 
 /**
  * Sleep for the specified amount of time in milliseconds.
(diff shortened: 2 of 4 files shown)


tests/unit/helpers.test.ts
```diff
@@ -0,0 +1,79 @@
+// @ts-ignore - Importing internal function for testing
```
0xmaayan, line 1: not sure why `ts-ignore`, should be able to just do
```
import { ... } from "../../src";
```

tests/unit/helpers.test.ts
```diff
@@ -0,0 +1,79 @@
...
+import { paired_fa_metadata_address, standardizeMoveTypeString } from "../../src/utils/helpers";
+
+describe("paired_fa_metadata_address", () => {
+    test("matches Rust implementation test cases", () => {
```
0xmaayan, line 5: what does it mean? can we be specific with what the test is doing

src/core/accountAddress.ts
```diff
@@ -496,4 +496,17 @@ Received ${input}`,
...
+   * If the address consists of all zeros, returns "0".
+   *
+   * @returns A string representation of the address without leading zeros
+   * @group Implementation
+   * @category Serialization
+   */
+  shortStrLossless(): string {
+    const hex = bytesToHex(this.data).replace(/^0+/, "");
+    return hex === "" ? "0" : hex;
+  }
```
gregnazario, line 511: There's already a function for this?

`toString()`

lightmark: I don't think ToString will trim the leading 0s

lightmark: * Return the AccountAddress as a string as per AIP-40.
   * https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md.
   * This representation returns special addresses in SHORT form (0xf)
   * and other addresses in LONG form (0x + 64 characters).

src/utils/helpers.ts
```diff
@@ -235,3 +237,57 @@ export function isValidFunctionInfo(functionInfo: string): boolean {
...
+ * pairedFaMetadataAddress("0x1::aptos_coin::AptosCoin")  // simple form
+ * pairedFaMetadataAddress("0x0000000000000000000000000000000000000000000000000000000000000001::aptos_coin::AptosCoin")  // long form
+ * pairedFaMetadataAddress("1::aptos_coin::AptosCoin")  // without 0x prefix
+ * pairedFaMetadataAddress("0x00001::aptos_coin::AptosCoin")  // with leading zeros
+ * pairedFaMetadataAddress("0x1::coin::Coin<0x1412::a::struct<0x0001::aptos_coin::AptosCoin>>")  // nested type parameters
+ *
+ * @param coinType - The coin type string in any of these formats:
+ *   - Short form address: "0x1::aptos_coin::AptosCoin"
+ *   - Long form address: "0x0000000000000000000000000000000000000000000000000000000000000001::aptos_coin::AptosCoin"
+ *   - Without 0x prefix: "1::aptos_coin::AptosCoin"
```
hardsetting, line 278: I have never seen the `1::aptos_coin::AptosCoin` case.
We should probably discourage users from using this format, I don't think we should support it

src/utils/helpers.ts
```diff
@@ -235,3 +237,57 @@ export function isValidFunctionInfo(functionInfo: string): boolean {
...
+ *   - With leading zeros: "0x00001::aptos_coin::AptosCoin"
+ *   - With nested types: "0x1::coin::Coin<0x1412::a::struct<0x0001::aptos_coin::AptosCoin>>"
+ * @returns The calculated metadata address as a hex string with "0x" prefix
+ */
+export function pairedFaMetadataAddress(coinType: string): string {
```
hardsetting, line 283: Thoughts on using a stricter type like `0x${string}::${string}::${string}`?

src/utils/helpers.ts
```diff
@@ -235,3 +237,57 @@ export function isValidFunctionInfo(functionInfo: string): boolean {
...
+
+/**
+ * Constants for metadata address calculation
+ */
+const APTOS_COIN_TYPE_STR = "0x1::aptos_coin::AptosCoin";
+const APT_METADATA_ADDRESS_HEX = AccountAddress.A.toStringLong();
+
+/**
+ * Helper function to standardize Move type string by converting all addresses to short form,
+ * including addresses within nested type parameters
```
hardsetting, line 249: Why are we using short form for FAs? I thought we agreed on using the same format across products?

src/utils/helpers.ts
```diff
@@ -4,6 +4,8 @@
 import { decode } from "js-base64";
 import { MoveFunctionId, MoveStructId } from "../types";
 import { AccountAddress } from "../core/accountAddress";
+import { createObjectAddress } from "../core/account/utils/address";
+import { sha3_256 } from "@noble/hashes/sha3";
```
hardsetting, line 8: where is this used?

src/core/accountAddress.ts
```diff
@@ -496,4 +496,17 @@ Received ${input}`,
...
+
+  /**
+   * Returns a lossless short string representation of the address by trimming leading zeros.
+   * If the address consists of all zeros, returns "0".
+   *
+   * @returns A string representation of the address without leading zeros
+   * @group Implementation
+   * @category Serialization
+   */
+  toStringShort(): string {
```
hardsetting, line 508: This doesn't seem to be including the prefix 0x, should we call it toStringShortWithoutPrefix?

src/utils/helpers.ts
```diff
@@ -235,3 +237,57 @@ export function isValidFunctionInfo(functionInfo: string): boolean {
...
+ * including addresses within nested type parameters
+ */
+function standardizeMoveTypeString(input: string): string {
+  // Regular expression to match addresses in the type string, including those within type parameters
+  // This regex matches "0x" followed by hex digits, handling both standalone addresses and those within <>
+  const addressRegex = /0x[0-9a-fA-F]+/g;
+
+  return input.replace(addressRegex, (match) => {
+    // Use AccountAddress to handle the address
+    return `0x${AccountAddress.from(match, { maxMissingChars: 63 }).toStringShort()}`;
```
hardsetting, line 258: Let's make a toStringShort() that returns the value with the prefix

src/utils/helpers.ts
gregnazario: This is what was hashed on-chain, so it's stuck like that.  The move to string does short form :(

src/utils/helpers.ts
gregnazario: I want to get rid of it again... I don't know how it got in.

If it's shorter than 64 characters, it should reject without a leading 0x.

src/utils/helpers.ts
gregnazario: These stricter types are actually really annoying, but for consistency, we should have that type.

src/utils/helpers.ts
```diff
@@ -235,3 +237,57 @@ export function isValidFunctionInfo(functionInfo: string): boolean {
...
+ *   - Short form address: "0x1::aptos_coin::AptosCoin"
+ *   - Long form address: "0x0000000000000000000000000000000000000000000000000000000000000001::aptos_coin::AptosCoin"
+ *   - Without 0x prefix: "1::aptos_coin::AptosCoin"
+ *   - With leading zeros: "0x00001::aptos_coin::AptosCoin"
+ *   - With nested types: "0x1::coin::Coin<0x1412::a::struct<0x0001::aptos_coin::AptosCoin>>"
+ * @returns The calculated metadata address as a hex string with "0x" prefix
+ */
+export function pairedFaMetadataAddress(coinType: string): string {
+  // Add 0x prefix if missing
+  const prefixedCoinType = coinType.startsWith("0x") ? coinType : `0x${coinType}`;
```
gregnazario, line 285: Struct types require a 0x in front, we can remove this.

src/utils/helpers.ts
```diff
@@ -235,3 +237,57 @@ export function isValidFunctionInfo(functionInfo: string): boolean {
...
+
+  // Standardize the coin type string to handle any address format
+  const standardizedMoveTypeName = standardizeMoveTypeString(prefixedCoinType);
+
+  return standardizedMoveTypeName === APTOS_COIN_TYPE_STR
+    ? AccountAddress.A.toStringLong()
+    : createObjectAddress(AccountAddress.A, standardizedMoveTypeName).toStringLong();
```
gregnazario, line 292: I thought types are usually short strings?  Why doesn't this function return `AccountAddress`?

src/utils/helpers.ts
lightmark: This is for offline computation of the paired metadata address. if you ask me try the algorithm is like this, it is due to the type_name implementation of David.

gregnazario (commented):
Okay, so this is what should exist:
1. toShortString -> gives you the short string with the leading 0x
2. all functions that take in coin types should be typed as `0x{string}::${string}::${string}`
3. All outputs of functions should be AccountAddress, unless it's explicitly to a string.  So, you should return the AccountAddress as the paired address, then call `.toShortString()`

src/utils/helpers.ts
lightmark: the Coin type can best nested as 0xaaa::a::a<0x111::b::b>

src/utils/helpers.ts
lightmark: you are right.

gregnazario (approved):

PR #679, opened 2025-04-14
Gabriele/txn worker improvements

hardsetting:
### Description
Proposal for new transaction worker design, to overcome some promise resolution issues with the current design.

### Test Plan
Testing using the gas station's load tester
Will add CI later

### Checklist


src/transactions/management/transactionWorker2.ts
```diff
@@ -0,0 +1,430 @@
...
+   * Number of currently pending submissions i.e. submission requests that have been sent,
+   * but have not been responded to yet.
+   * For context, a submission request is expected to return a pending response that indicates
+   * the transaction was accepted into the mempool for later execution.
+   */
+  private pendingSubmissions = 0;
+
+  /**
+   * A set of pending responses returned by submission requests.
+   * When the set is not empty, the worker will regularly check for the account'scurrent sequence number
```
banool, line 151: nit
```suggestion
   * When the set is not empty, the worker will regularly check for the account's current sequence number
```

src/transactions/management/transactionWorker2.ts
```diff
@@ -0,0 +1,430 @@
...
+ */
+interface Deferred<T> {
+  resolve: (value: T) => void;
+  reject: (reason: unknown) => void;
+}
+
+interface TransactionWorkerInput {
+  data: InputGenerateTransactionPayloadData;
+  options?: Omit<InputGenerateTransactionOptions, "accountSequenceNumber">;
+  deferred: Deferred<string>;
```
banool, line 47: It'd be good to have a comment saying what the string in this Deferred is for.

src/transactions/management/transactionWorker2.ts
```diff
@@ -0,0 +1,430 @@
...
+   * which includes both pending submissions and pending responses.
+   */
+  private get totalPendingResponses() {
+    return this.pendingResponses.size + this.pendingSubmissions;
+  }
+
+  /**
+   * Simple utility function for getting the current account's sequence number
+   */
+  private async getCurrSequenceNumber() {
```
banool, line 213: nit
```suggestion
  private async getCurrrentSequenceNumber() {
```

src/transactions/management/transactionWorker2.ts
```diff
@@ -0,0 +1,430 @@
...
+
+      if (!this.isRunning && this.totalPendingResponses === 0) {
+        break;
+      }
+
+      const sequenceNumber = await this.getCurrSequenceNumber();
+
+      for (const response of this.pendingResponses) {
+        if (sequenceNumber > response.sequenceNumber) {
+          response.deferred.resolve(response.hash);
```
banool, line 357: Could you add a comment explaining why we pass the txn hash in here?

src/transactions/management/transactionWorker2.ts
```diff
@@ -0,0 +1,430 @@
...
+   * Starts the transaction management process. You can await the Promise returned by
+   * this method if you want to handle any errors that may occur with the worker's
+   * internal tasks.
+   *
+   * @throws {Error} Throws an error if the worker has already started.
+   * @group Implementation
+   * @category Transactions
+   */
+  async start() {
+    if (this.isRunning) {
```
banool, line 332: Do we need both `isRunning` and `stopSignal`?

banool (commented):
This looks great, much much simpler, and actually more powerful thanks to the failed seqnum handling behavior (was the previous version correctly handling gaps due to rejected txns? I don't think so).

I suppose next step is to add events / signals / etc that the caller can subscribe to?

src/transactions/management/transactionWorker2.ts
```diff
@@ -0,0 +1,430 @@
...
+  /**
+   * Get the next sequence number to be used to build a transaction.
+   * The first time this is called, it will return the account's current sequence number.
+   * Subsequent calls will return the previously returned value, incremented by 1.
+   *
+   * In case `failedSequenceNumbers` is not empty because of submission errors,
+   * the first element will be popped and reclaimed to fill the sequence number gap and resume
+   * normal operation.
+   */
+  private async getNextSequenceNumber() {
```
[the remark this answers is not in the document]
hardsetting: You don't need to have any `await` in the body of an `async` function in javascript, although often times that defeats the purpose.

From the outside though, the function will always return a promise.
It's good to "simulate" async behavior if needed.

In this particular context, we're actually returning a promise, so it still makes sense.

```typescript
return await promise;
```
and 
```typescript
return promise;
```
are mostly equivalent and the latter is generally preferred.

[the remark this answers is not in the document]
hardsetting: In this case, if I dropped the `async` modifier, the function would have been the same (because it returns a promise), but I think it's better to let the "reader" know this is async.

src/transactions/management/transactionWorker2.ts
```diff
@@ -0,0 +1,430 @@
...
+    if (this.isRunning) {
+      throw new Error("Worker is already running");
+    }
+    this.isRunning = true;
+
+    // main loop
+    while (true) {
+      // sleep until there are pending responses or the worker is stopped
+      while (this.pendingResponses.size === 0 && (this.isRunning || this.pendingSubmissions > 0)) {
+        await Promise.race([this.newPendingResponseSignal.wait(), this.stopSignal.wait()]);
```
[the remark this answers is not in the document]
hardsetting: Very neat stuff

src/transactions/management/transactionWorker2.ts
```diff
@@ -0,0 +1,430 @@
...
+
+      for (const response of this.pendingResponses) {
+        if (sequenceNumber > response.sequenceNumber) {
+          response.deferred.resolve(response.hash);
+          this.pendingResponses.delete(response);
+        } else if (Math.floor(Date.now() / 1000) > response.expirationTimestamp) {
+          response.deferred.reject(new Error("transaction expired"));
+          this.pendingResponses.delete(response);
+          // Add sequence number to pool of failed sequence numbers to be reused
+          console.log(
```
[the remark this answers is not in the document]
hardsetting: Ideally you can use what's returned by the API to log.
(assuming the internals are working as expected)

PR #691, opened 2025-05-02
Fix incorrect assumption about withFeePayer

banool:
### Description
The SDK makes some invalid assumptions around fee payers, it is valid to set `withFeePayer` but not set a `feePayerAddress`.

### Test Plan
CI. I also tested manually as part of https://github.com/aptos-labs/confidential-payments-example.

### Checklist

### DIFF
--- a/confidential-assets/src/confidentialAsset.ts
+++ b/confidential-assets/src/confidentialAsset.ts
@@ -92,12 +92,13 @@ export class ConfidentialAsset {
     args: {
       tokenAddress: string;
       publicKey: HexInput | TwistedEd25519PublicKey;
-    } & Omit<InputGenerateSingleSignerRawTransactionArgs, "payload" | "aptosConfig">,
+      withFeePayer?: boolean;
+    } & Omit<InputGenerateSingleSignerRawTransactionArgs, "payload" | "aptosConfig" | "feePayerAddress">,
   ): Promise<SimpleTransaction> {
     const pkU8 = publicKeyToU8(args.publicKey);
     return this.client.transaction.build.simple({
       ...args,
-      withFeePayer: Boolean(args.feePayerAddress),
+      withFeePayer: args.withFeePayer,
       data: {
         function: `${this.confidentialAssetModuleAddress}::${MODULE_NAME}::register`,
         functionArguments: [args.tokenAddress, pkU8],
@@ -109,12 +110,14 @@ export class ConfidentialAsset {
     args: {
       tokenAddress: string;
       amount: AnyNumber;
+      /** If not set we will use the sender's address. */
       to?: AccountAddress;
-    } & Omit<InputGenerateSingleSignerRawTransactionArgs, "payload" | "aptosConfig">,
+      withFeePayer?: boolean;
+    } & Omit<InputGenerateSingleSignerRawTransactionArgs, "payload" | "aptosConfig" | "feePayerAddress">,
   ): Promise<SimpleTransaction> {
     return this.client.transaction.build.simple({
       ...args,
-      withFeePayer: Boolean(args.feePayerAddress),
+      withFeePayer: args.withFeePayer,
       sender: args.sender,
       data: {
         function: `${this.confidentialAssetModuleAddress}::${MODULE_NAME}::deposit_to`,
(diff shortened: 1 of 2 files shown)


confidential-assets/src/confidentialAsset.ts
```diff
@@ -110,11 +112,13 @@ export class ConfidentialAsset {
...
       amount: AnyNumber;
       to?: AccountAddress;
+      /** This value will be ignored if feePayerAddress is provided. */
+      withFeePayer?: boolean;
     } & Omit<InputGenerateSingleSignerRawTransactionArgs, "payload" | "aptosConfig">,
   ): Promise<SimpleTransaction> {
     return this.client.transaction.build.simple({
       ...args,
-      withFeePayer: Boolean(args.feePayerAddress),
+      withFeePayer: args.feePayerAddress !== undefined ? true : args.withFeePayer,
```
0xmaayan, line 121: The arguments it expects here is a bit weird, in regards to the `withFeePayer` - we dont really need them to provide the fee payer address. In fact, the sdk was built in a way that we dont require them to provide it because a lot of the times they might just not know it. This is why we use the default fee payer address
https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/internal/transactionSubmission.ts#L164

It is much cleaner to assign the fee payer address when the fee payer actually signs the transaction
https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/internal/transactionSubmission.ts#L254

I dont see why these functions even need the `InputGenerateSingleSignerRawTransactionArgs` args, all they really need is the `build.simple` expected args excluding the [data](https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/api/transactionSubmission/build.ts#L99) which this function builds internally 
```
    sender: AccountAddressInput;
    options?: InputGenerateTransactionOptions;
    withFeePayer?: boolean;
```

confidential-assets/src/confidentialAsset.ts
```diff
@@ -146,8 +152,10 @@ export class ConfidentialAsset {
 
   async withdraw(
     args: CreateConfidentialWithdrawOpArgs & {
-      to?: AccountAddressInput;
       tokenAddress: string;
+      to?: AccountAddressInput;
```
0xmaayan, line 156: we should either make it a required field, or mention that if it is not set, it refers to the sender

confidential-assets/src/confidentialAsset.ts
banool: There is still a use case for letting people set it right, e.g. if (for whatever reason) they only want a certain fee payer to sign it. I agree it's niche though, but presumably that's why the regular SDK lets you set it.

0xmaayan: actually, the sdk does not let you set it when you build the transaction (using `build.simple`). It only lets you define whether it is a sponsor transaction or not. Then, if it is, it sets a [default address](https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/internal/transactionSubmission.ts#L164), and only later when the sponsor signs it, the sdk assigns the fee payer address.

It is not really relevant to set the sponsor address when building the transaction, because either way the sponsor would need to sign it later.

banool: I see, thanks for the info! I'll not let them set the `feePayerAddress` at all then.

confidential-assets/src/confidentialAsset.ts
```diff
@@ -92,12 +92,13 @@ export class ConfidentialAsset {
     args: {
       tokenAddress: string;
       publicKey: HexInput | TwistedEd25519PublicKey;
-    } & Omit<InputGenerateSingleSignerRawTransactionArgs, "payload" | "aptosConfig">,
+      withFeePayer?: boolean;
+    } & Omit<InputGenerateSingleSignerRawTransactionArgs, "payload" | "aptosConfig" | "feePayerAddress">,
```
0xmaayan, line 96: nit: I would just define `options?: InputGenerateTransactionOptions, sender: AccountAddressInput` as args instead of the `Omit` =)

https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/transactions/types.ts#L322

banool: Oops automerged, I'll think about this on the next PR.

PR #710, opened 2025-05-21
Fix encrypted balances where wrong randomness was used

heliuchuan:
### Checklist

### DIFF
--- a/confidential-assets/src/confidentialKeyRotation.ts
+++ b/confidential-assets/src/confidentialKeyRotation.ts
@@ -41,6 +41,8 @@ export class ConfidentialKeyRotation {
 
   newConfidentialAmount: ConfidentialAmount;
 
+  newEncryptedBalance: TwistedElGamalCiphertext[];
+
   constructor(args: {
     currDecryptionKey: TwistedEd25519PrivateKey;
     newDecryptionKey: TwistedEd25519PrivateKey;
@@ -55,21 +57,31 @@ export class ConfidentialKeyRotation {
     this.currEncryptedBalance = args.currEncryptedBalance;
     this.currConfidentialAmount = args.currConfidentialAmount;
     this.newConfidentialAmount = args.newConfidentialAmount;
+
+    this.newEncryptedBalance = this.newConfidentialAmount.getAmountEncrypted(
+      this.newDecryptionKey.publicKey(),
+      this.randomness,
+    );
   }
 
   static FIAT_SHAMIR_SIGMA_DST = "AptosConfidentialAsset/RotationProofFiatShamir";
 
   static async create(args: CreateConfidentialKeyRotationOpArgs) {
-    const randomness = args.randomness ?? ed25519GenListOfRandom(ConfidentialAmount.CHUNKS_COUNT);
+    const {
+      randomness = ed25519GenListOfRandom(ConfidentialAmount.CHUNKS_COUNT),
+      currEncryptedBalance,
+      currDecryptionKey,
+      newDecryptionKey,
+    } = args;
 
-    const currentBalance = await ConfidentialAmount.fromEncrypted(args.currEncryptedBalance, args.currDecryptionKey);
+    const currentBalance = await ConfidentialAmount.fromEncrypted(currEncryptedBalance, currDecryptionKey);
 
     const newBalance = ConfidentialAmount.fromAmount(currentBalance.amount);
 
     return new ConfidentialKeyRotation({
-      currDecryptionKey: args.currDecryptionKey,
-      newDecryptionKey: args.newDecryptionKey,
-      currEncryptedBalance: args.currEncryptedBalance,
+      currDecryptionKey,
+      newDecryptionKey,
+      currEncryptedBalance,
       randomness,
       currConfidentialAmount: currentBalance,
       newConfidentialAmount: newBalance,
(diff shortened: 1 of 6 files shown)


confidential-assets/src/confidentialKeyRotation.ts
```diff
@@ -55,21 +57,31 @@ export class ConfidentialKeyRotation {
...
+      this.randomness,
+    );
   }
 
   static FIAT_SHAMIR_SIGMA_DST = "AptosConfidentialAsset/RotationProofFiatShamir";
 
   static async create(args: CreateConfidentialKeyRotationOpArgs) {
-    const randomness = args.randomness ?? ed25519GenListOfRandom(ConfidentialAmount.CHUNKS_COUNT);
+    const {
+      randomness = ed25519GenListOfRandom(ConfidentialAmount.CHUNKS_COUNT),
```
banool, line 71: What about the provided randomness, we're just not using it now right? In that case `CreateConfidentialKeyRotationOpArgs` shouldn't accept it as an arg at all.

confidential-assets/src/confidentialNormalization.ts
```diff
@@ -48,8 +50,12 @@ export class ConfidentialNormalization {
     this.decryptionKey = args.decryptionKey;
     this.unnormalizedEncryptedBalance = args.unnormalizedEncryptedBalance;
     this.balanceAmount = args.balanceAmount;
-    this.normalizedConfidentialAmount = args.normalizedConfidentialAmount;
+    this.normalizedConfidentialBalance = args.normalizedConfidentialAmount;
```
banool, line 53: Let's change the name of the arg too then?

confidential-assets/src/confidentialNormalization.ts
```diff
@@ -48,8 +50,12 @@ export class ConfidentialNormalization {
...
     this.randomness = args.randomness;
+    this.normalizedEncryptedBalance = this.normalizedConfidentialBalance.getAmountEncrypted(
```
banool, line 55: I'm assuming the user will always want this, so it's good to compute it now, rather than doing it lazily later?

confidential-assets/src/confidentialTransfer.ts
```diff
@@ -58,7 +58,11 @@ export class ConfidentialTransfer {
 
   confidentialAmountToTransfer: ConfidentialAmount;
 
-  confidentialAmountAfterTransfer: ConfidentialAmount;
+  encryptedAmountToTransfer: TwistedElGamalCiphertext[];
+
+  confidentialBalanceAfterTransfer: ConfidentialAmount;
+
+  encryptedBalanceAfterTransfer: TwistedElGamalCiphertext[];
```
banool, line 65: Can we get some comments on these? I think I get it but someone new probably wouldn't know what is different between confidential x and encrypted x.

confidential-assets/src/confidentialKeyRotation.ts
heliuchuan: It is being used in the constructor.  Line 63.

confidential-assets/src/confidentialNormalization.ts
heliuchuan: Not sure if you are asking a question.

Am I lazily doing it later in this PR?  I'm doing it 'now' in the constructor right?

confidential-assets/src/confidentialTransfer.ts
heliuchuan: Will soon be refactoring this into a EncryptedAmount class that will combine ConfidentialAmount and TwistedElGamalCiphertext[] fyi.  

Plan to get rid of the use of ConfidentialAmount as attribute throughout SDK.

confidential-assets/src/confidentialTransfer.ts
```diff
@@ -86,55 +112,75 @@ export class ConfidentialTransfer {
...
     const newBalanceRandomness = ed25519GenListOfRandom(ConfidentialAmount.CHUNKS_COUNT);
-    const recipientPublicKeyU8 = publicKeyToU8(args.recipientEncryptionKey);
+    const recipientPublicKeyU8 = publicKeyToU8(recipientEncryptionKey);
 
-    const confidentialAmountToTransfer = ConfidentialAmount.fromAmount(args.amountToTransfer, {
+    const confidentialAmountToTransfer = ConfidentialAmount.fromAmount(amountToTransfer, {
       chunksCount: ConfidentialAmount.CHUNKS_COUNT / 2,
     });
 
     const encryptedAmountByRecipient = confidentialAmountToTransfer.amountChunks.map((chunk, i) =>
```
heliuchuan, line 150: note this is a very unfortunate artifact of the way confidentialAmount does encrypting.  can be fixed after - https://github.com/aptos-labs/aptos-ts-sdk/pull/711

confidential-assets/src/confidentialNormalization.ts
banool: I see that. Let me rephrase, why are we computing this now vs doing it when the user actually wants it.

heliuchuan: Yes, because the user will always want this if they are going to normalize

banool: okay cool cool

banool (approved):

PR #713, opened 2025-05-26
Simplify the confidential asset API

heliuchuan:
### Description

This renames many of the functions, function parameters, and gets rid of unneeded params that are derivable or retrievable on chain.

Notable changes
- Instead of actual and pending balance, use available and pending.  This is inline with standard banking nomenclature.
- Fetch the sender encrypted available balance from the chain instead of needing to pass it in.
- Fetch the recipient encryption key from the chain instead of passing it in when transferring.
- Get rid of 'safe' versions of  normalize/rollover/rotate key.  The chain will prevent 'unsafe' operations
- Check chain relevant chain state for normalize/rollover/rotate key and throw error when appropriate
- Updates getBalance for convenience
- Adds func docs
- Condenses deposit functions into one
- Refactors the e2e tests and adds a few cases
- Does some input validation on amount
- Preload the tables when client is created
- Adds EncryptedAmount to deprecate use of ConfidentialAmount

### Test Plan

Uses existing tests and adds new tests

### Checklist


confidential-assets/src/chunkedAmount.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  chunkBits = CHUNK_BITS;
+
+  static CHUNKS_COUNT = CHUNKS_COUNT;
+
+  static CHUNK_BITS = CHUNK_BITS;
+
+  static CHUNK_BITS_BIG_INT = BigInt(ChunkedAmount.CHUNK_BITS);
+
+  static CHUNKS_COUNT_HALF = ChunkedAmount.CHUNKS_COUNT / 2;
```
banool, line 28: This feels awkward, what if the dev sets one of these values. Even we don't get it right right, like if they pass in `chunksCount` to the constructor we only update `chunksCount`, not  `CHUNKS_COUNT`.

PR #716, opened 2025-06-04
[txns] Add support for orderless transactions

gregnazario:
### Description
This enables orderless transactions to be submitted.  Instead of submitting account sequence number, users would submit the replayProtectionNonce

### Checklist

### DIFF
--- a/examples/typescript/simple_turbo_transfer.ts
+++ b/examples/typescript/simple_turbo_transfer.ts
@@ -93,7 +93,7 @@ const example = async () => {
       functionArguments: [bob.accountAddress, TRANSFER_AMOUNT],
     },
     options: {
-      replayNonce: 12345,
+      replayProtectionNonce: 12345,
     },
   });
 
--- a/examples/typescript/simple_turbo_transfer.ts
+++ b/examples/typescript/simple_turbo_transfer.ts
@@ -1,0 +1,119 @@
+/* eslint-disable no-console */
+
+/**
+ * This example shows how to use the Aptos client to create accounts, fund them, and transfer between them.
+ */
+
+import {
+  Account,
+  AccountAddress,
+  Aptos,
+  AptosConfig,
+  InputViewFunctionJsonData,
+  Network,
+  NetworkToNetworkName,
+} from "@aptos-labs/ts-sdk";
+import dotenv from "dotenv";
+dotenv.config();
+
+const APTOS_COIN = "0x1::aptos_coin::AptosCoin";
+const ALICE_INITIAL_BALANCE = 100_000_000;
+const BOB_INITIAL_BALANCE = 100;
+const TRANSFER_AMOUNT = 100;
+
+// Default to devnet, but allow for overriding
+const APTOS_NETWORK: Network = NetworkToNetworkName[process.env.APTOS_NETWORK ?? Network.DEVNET];
+
+/**
+ * Prints the balance of an account
+ * @param aptos
+ * @param name
+ * @param address
+ * @returns {Promise<*>}
+ *
+ */
+const balance = async (aptos: Aptos, name: string, address: AccountAddress): Promise<any> => {
+  const payload: InputViewFunctionJsonData = {
+    function: "0x1::coin::balance",
+    typeArguments: ["0x1::aptos_coin::AptosCoin"],
+    functionArguments: [address.toString()],
+  };
+  const [balance] = await aptos.viewJson<[number]>({ payload: payload });
+
+  console.log(`${name}'s balance is: ${balance}`);
+  return Number(balance);
+};
+
+const example = async () => {
+  console.log("This example will create two accounts (Alice and Bob), fund them, and transfer between them.");
+
+  // Set up the client
+  const config = new AptosConfig({ network: APTOS_NETWORK });
+  const aptos = new Aptos(config);
+
+  // Create two accounts
+  const alice = Account.generate();
+  const bob = Account.generate();
+
+  console.log("=== Addresses ===\n");
+  console.log(`Alice's address is: ${alice.accountAddress}`);
+  console.log(`Bob's address is: ${bob.accountAddress}`);
+
+  // Fund the accounts
+  console.log("\n=== Funding accounts ===\n");
+
+  const aliceFundTxn = await aptos.fundAccount({
+    accountAddress: alice.accountAddress,
+    amount: ALICE_INITIAL_BALANCE,
+  });
+  console.log("Alice's fund transaction: ", aliceFundTxn);
+
+  const bobFundTxn = await aptos.fundAccount({
+    accountAddress: bob.accountAddress,
+    amount: BOB_INITIAL_BALANCE,
+  });
+  console.log("Bob's fund transaction: ", bobFundTxn);
+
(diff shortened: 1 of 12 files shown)


src/api/transactionSubmission/build.ts
```diff
@@ -3,7 +3,11 @@
 
 import { AccountAddressInput } from "../../core";
 import { generateTransaction } from "../../internal/transactionSubmission";
-import { InputGenerateTransactionPayloadData, InputGenerateTransactionOptions } from "../../transactions";
+import {
+  InputGenerateTransactionPayloadData,
+  InputGenerateTransactionOptions,
+  InputGenerateOrderlessTransactionOptions,
```
0xmaayan, line 9: Does not seem like it is being used
```suggestion
  InputGenerateOrderlessTransactionOptions,
```

src/transactions/instances/transactionPayload.ts
```diff
@@ -511,3 +513,187 @@ export class MultiSigTransactionPayload extends Serializable {
     return new MultiSigTransactionPayload(EntryFunction.deserialize(deserializer));
   }
 }
+
+export enum TransactionInnerPayloadVariants {
```
0xmaayan, line 517: Does this match the rust enum? if so lets add a link to it.
Also, lets add a doc string for this enum

src/transactions/instances/transactionPayload.ts
```diff
@@ -511,3 +513,187 @@ export class MultiSigTransactionPayload extends Serializable {
...
+  }
+
+  static load(deserializer: Deserializer): TransactionInnerPayloadV1 {
+    let executable = TransactionExecutable.deserialize(deserializer);
+    let extra_config = TransactionExtraConfig.deserialize(deserializer);
+    return new TransactionInnerPayloadV1(executable, extra_config);
+  }
+}
+
+export enum TransactionExecutableVariants {
```
0xmaayan, line 584: same comment

src/transactions/instances/transactionPayload.ts
0xmaayan: Also, can this be places in the `src/types/types.ts` file like all the others https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/types/types.ts#L72

src/transactions/instances/transactionPayload.ts
```diff
@@ -511,3 +513,187 @@ export class MultiSigTransactionPayload extends Serializable {
...
+export enum TransactionInnerPayloadVariants {
+  V1 = 0,
+}
+
+/**
+ * Represents any transaction payload that can be submitted to the Aptos chain for execution.
+ *
+ * This is specifically used for orderless transactions.
+ */
+export class TransactionPayloadPayload extends TransactionPayload {
```
0xmaayan, line 526: thoughts? also aligns with the enum name
```suggestion
export class TransactionInnerPayload extends TransactionPayload {
```

src/transactions/instances/transactionPayload.ts
```diff
@@ -511,3 +513,187 @@ export class MultiSigTransactionPayload extends Serializable {
...
+  serialize(serializer: Serializer): void {
+    serializer.serializeU32AsUleb128(TransactionExecutableVariants.Empty);
+  }
+
+  static load(_: Deserializer): TransactionExecutableEmpty {
+    return new TransactionExecutableEmpty();
+  }
+}
+
+export enum TransactionExtraConfigVariants {
```
0xmaayan, line 657: same comments

src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -409,24 +427,72 @@ export async function generateRawTransaction(args: {
...
-    payload,
+    txnPayload,
     BigInt(maxGasAmount),
     BigInt(gasUnitPrice),
     BigInt(expireTimestamp),
     new ChainId(chainId),
   );
 }
 
+export function convertPayloadToPayloadPayload(
```
0xmaayan, line 455:
```suggestion
export function convertPayloadToInnerPayload(
```

src/transactions/types.ts
```diff
@@ -113,13 +114,24 @@ export type AnyRawTransactionInstance = RawTransaction | MultiAgentRawTransactio
  * @group Implementation
  * @category Transactions
  */
-export type InputGenerateTransactionOptions = {
+export type InputGenerateTransactionOptions =
+  | InputGenerateSequenceNumberTransactionOptions
+  | InputGenerateOrderlessTransactionOptions;
+
+export type InputGenerateSequenceNumberTransactionOptions = {
```
0xmaayan, line 121: this is a breaking change, we need to release a major version

src/transactions/types.ts
```diff
@@ -113,13 +114,24 @@ export type AnyRawTransactionInstance = RawTransaction | MultiAgentRawTransactio
...
   maxGasAmount?: number;
   gasUnitPrice?: number;
   expireTimestamp?: number;
   accountSequenceNumber?: AnyNumber;
```
0xmaayan, line 125: you can add the `replayNonce: undefined` here and then [here](https://github.com/aptos-labs/aptos-ts-sdk/pull/716/files) simply do `if(options?.replayNonce)`

src/api/transactionSubmission/build.ts
gregnazario: Oh hm... I thought that was removed, some weird merge issue.  will fix

src/transactions/instances/transactionPayload.ts
gregnazario: Yep, will do.  I got caught up in serialization issues, that I didn't move them to the appropriate locations.

0xmaayan (approved):
LGTM. need to fix the tests, and remove the unused `InputGenerateTurboTransactionOptions` from the `build.ts` file

vusirikala:
I'm assuming there must be some part of code where the accountSequenceNumber is incremented upon submitting transactions. I think we should also avoid incrementing the local accountSequenceNumber of an account when submitting turbo transactions.

vusirikala:
In the aptos-core side, we've also added `replay_protection_nonce` field in the output of APIs that retrieve transactions. We might have to add the `replay_protection_nonce` field in the read path of the SDK as well.

examples/typescript/simple_turbo_transfer.ts
```diff
@@ -0,0 +1,119 @@
...
+  // Transfer between users
+  const txn = await aptos.transaction.build.simple({
+    sender: alice.accountAddress,
+    data: {
+      function: "0x1::coin::transfer",
+      typeArguments: [APTOS_COIN],
+      functionArguments: [bob.accountAddress, TRANSFER_AMOUNT],
+    },
+    options: {
+      replayNonce: 12345,
```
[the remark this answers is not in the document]
gregnazario: I cut it down cause it was long but 🤷🏻‍♂️

gregnazario:
> In the aptos-core side, we've also added `replay_protection_nonce` field in the output of APIs that retrieve transactions. We might have to add the `replay_protection_nonce` field in the read path of the SDK as well.

Will have to look

PR #720, opened 2025-06-08
Make default gas and expiry time configurable at client construction

heliuchuan:
### Description

This allows for the dev to change the default without setting it in every call

### Checklist

### DIFF
--- a/src/api/aptosConfig.ts
+++ b/src/api/aptosConfig.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import aptosClient from "@aptos-labs/aptos-client";
-import { AptosSettings, ClientConfig, Client, FullNodeConfig, IndexerConfig, FaucetConfig } from "../types";
+import { AptosSettings, ClientConfig, Client, FullNodeConfig, IndexerConfig, FaucetConfig, TransactionGenerationConfig } from "../types";
 import {
   NetworkToNodeAPI,
   NetworkToFaucetAPI,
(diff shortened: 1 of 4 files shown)


src/api/aptosConfig.ts
```diff
@@ -154,6 +168,10 @@ export class AptosConfig {
     this.fullnodeConfig = settings?.fullnodeConfig ?? {};
     this.indexerConfig = settings?.indexerConfig ?? {};
     this.faucetConfig = settings?.faucetConfig ?? {};
+    this.transactionGenerationConfig = settings?.transactionGenerationConfig ?? {
+      DEFAULT_MAX_GAS_AMOUNT: 200000,
+      DEFAULT_TXN_EXPIRY_SEC_FROM_NOW: 20,
```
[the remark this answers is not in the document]
gregnazario: code constants are in caps, but these are not constants, they're configuration, so they should be camel case

We additionally fix all other inputs should match language native constructs.  If you take a look at all the inputs they're camelCase and not snake_case.

src/api/aptosConfig.ts
```diff
@@ -154,6 +168,10 @@ export class AptosConfig {
     this.fullnodeConfig = settings?.fullnodeConfig ?? {};
     this.indexerConfig = settings?.indexerConfig ?? {};
     this.faucetConfig = settings?.faucetConfig ?? {};
+    this.transactionGenerationConfig = settings?.transactionGenerationConfig ?? {
```
0xmaayan, line 171: what is the reasoning in adding it in the global config vs adding it in the transaction option prop? https://github.com/aptos-labs/aptos-ts-sdk/blob/main/src/transactions/types.ts#L116

heliuchuan: Because it has to set on every call, which is annoying. Theres a default, but it's not configurable.  This makes the defaul val configurable, while allowing for override

0xmaayan (approved):

gregnazario (approved):

PR #737, opened 2025-07-01
Support `uint8array` for parsing `AccountAddress` with remote ABI

0xmaayan:
### Description
Add support for Uint8Array coming from external sources when parsing the function arguments using the remoteAbi.

Usually, dapps will be getting the account address as an object of `{data:Uint8Array}` from the Wallet (following
the aptos wallet standard), and use it to construct a transaction and pass it to the wallet adapter.

From there the wallet adapter either passes it along to the wallet (which works, might have some kind of internal normalization logic) or directly to the ts-sdk. If the later, the process fails with `type mismatch` as the sdk does not handle `uint8array` types

### Test Plan
tested with local build against the wallet adapter

### Checklist

### DIFF
--- a/src/transactions/transactionBuilder/remoteAbi.ts
+++ b/src/transactions/transactionBuilder/remoteAbi.ts
@@ -369,10 +369,15 @@ function parseArg(
      */
     throwTypeMismatch("boolean", position);
   }
-  // TODO: support uint8array?
   if (param.isAddress()) {
     if (isString(arg)) {
       return AccountAddress.fromString(arg);
+    }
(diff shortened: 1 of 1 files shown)


src/transactions/transactionBuilder/remoteAbi.ts
```diff
@@ -369,11 +369,16 @@ function parseArg(
      */
     throwTypeMismatch("boolean", position);
   }
-  // TODO: support uint8array?
```
gregnazario, line 372: yay!

gregnazario (approved):

PR #748, opened 2025-08-04
Add transaction manager

heliuchuan:
### Description

This adds a `TransactionManager` to make interacting with the Aptos blockchain much easier.

Before

```ts
const config = new AptosConfig({ network: Network.DEVNET });
const aptos = new Aptos(config);
await aptos.fundAccount({
  accountAddress: alice.accountAddress,
  amount: ALICE_INITIAL_BALANCE,
});
const transaction = await aptos.transferCoinTransaction({
  sender: alice.accountAddress,
  recipient: bob.accountAddress,
  amount: TRANSFER_AMOUNT,
});
const pendingTxn = await aptos.signAndSubmitTransaction({ signer: alice, transaction });
const response = await aptos.waitForTransaction({ transactionHash: pendingTxn.hash });

```

After

```ts
const config = new AptosConfig({ network: Network.DEVNET });
const txnManager = TransactionManager.new(config).defaultSender(alice);
await txnManager.fundAccount().submit();
const response = await txnManager
  .transferCoinTransaction({
    recipient: bob.accountAddress,
    amount: TRANSFER_AMOUNT,
  })
  .submit();
```

See updated typescript examples in this PR for examples.

### Checklist

### DIFF
--- a/src/api/manager.ts
+++ b/src/api/manager.ts
@@ -399,6 +399,20 @@ export class TransactionContext extends SenderlessTransactionContext implements 
     return this.aptos;
   }
 
+  async simulate(): Promise<UserTransactionResponse> {
+    const transaction = await this.generateTransaction();
+    const [userTxn] = await this.aptos.transaction.simulate.simple({
+      feePayerPublicKey: this.feePayerAccount?.publicKey,
+      options: {
+        estimateGasUnitPrice: this.gasUnitPrice === undefined,
+        estimateMaxGasAmount: this.maxGasAmount === undefined,
+      },
+      signerPublicKey: this.sender.publicKey,
+      transaction,
+    });
+    return userTxn;
+  }
+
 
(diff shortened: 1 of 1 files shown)


0xmaayan (commented):
- Looks like there is no option to simulate a transaction? 
- This flow only really supports signAndSubmit as a one step, what if I only want to generate a transaction and then sign it as two separate steps?
- Similarly, Wallets use the sdk with the generate->simulate->submit flow - I dont think this flow supports it?
- `submit()` by default waits for transaction - what if I dont want to wait for it?
- I am not completely sure what this flow tries to solve? it looks like it is more of a personal preference of code writing.
- On this ^ topic, things like `registerName()`, `renewDomain()` feel a bit confusing to me, I think having `aptos.ans.registerName()` is more explanatory and meaningful for external devs

PR #752, opened 2025-08-06
ledger_version as new param of keyless verifySignatureAsync

zjma:
### Description

In some dapp, users need to prove to a 3rd-party that "(a) it owns a blockchain account and (b) for that account, some condition holds true`. As an example, a distributed content platform dapp may hold content encrypted with a 3rd party's public key, and users will purchase the content (reflected in some on-chain dapp state), provide such a proof to the 3rd party to convince it to help decrypt the data.

To implement such a 3rd party, (a) can be done with `verifySignatureAsync()` APIs in this SDK, and (b) can be done by calling a view function provided by the dapp. But to check if both are true at the same time, (a) and (b) have to be done against a same recent ledger version.

This PR enables that by augmenting `verifySignatureAsync()`s.

### Test Plan
manual testing

### Checklist

### DIFF
--- a/src/core/crypto/keyless.ts
+++ b/src/core/crypto/keyless.ts
@@ -201,6 +201,7 @@ export class KeylessPublicKey extends AccountPublicKey {
     message: HexInput;
     signature: Signature;
     options?: { throwErrorWithReason?: boolean };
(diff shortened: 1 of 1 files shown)


zjma:
no longer needed.

PR #767, opened 2025-09-16
[account] Add Get Balance Function

WGB5445:
### Checklist

### DIFF
--- a/src/api/account.ts
+++ b/src/api/account.ts
@@ -908,7 +908,10 @@ export class Account {
    * ```
    * @group Account
    */
-  async getBalance(args: { accountAddress: AccountAddressInput; asset: MoveStructId | AccountAddressInput }): Promise<number> {
+  async getBalance(args: {
+    accountAddress: AccountAddressInput;
+    asset: MoveStructId | AccountAddressInput;
+  }): Promise<number> {
     return getBalance({ aptosConfig: this.config, ...args });
   }
 
--- a/src/api/account.ts
+++ b/src/api/account.ts
@@ -904,7 +904,7 @@ export class Account {
    * // APT coin by type
    * const apt = await aptos.getBalance({ accountAddress: "0x1", asset: "0x1::aptos_coin::AptosCoin" });
    * // Some FA by metadata address
-   * const fa = await aptos.getBalance({ accountAddress: "0x1", asset: "0xA" });
+   * const fa = await aptos.getBalance({ accountAddress: "0x1", asset: "0xa" });
    * ```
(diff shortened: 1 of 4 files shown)


src/api/account.ts
```diff
@@ -890,6 +890,31 @@ export class Account {
...
+   * @param args.asset The asset to query: Move struct ID (e.g., `0x1::aptos_coin::AptosCoin`) or FA metadata address.
+   * @returns The balance as a number.
+   *
+   * @example
+   * ```ts
+   * const aptos = new Aptos(new AptosConfig());
+   * // APT coin by type
+   * const apt = await aptos.getBalance({ accountAddress: "0x1", asset: "0x1::aptos_coin::AptosCoin" });
+   * // Some FA by metadata address
+   * const fa = await aptos.getBalance({ accountAddress: "0x1", asset: "0xA" });
```
banool, line 907:
```suggestion
   * const fa = await aptos.getBalance({ accountAddress: "0x1", asset: "0xa" });
```
Nit, we tend to represent these in lower case.

banool (approved):

src/internal/account.ts
```diff
@@ -745,6 +745,39 @@ export async function getAccountCoinsCount(args: {
...
+export async function getBalance(args: {
+  aptosConfig: AptosConfig;
+  accountAddress: AccountAddressInput;
+  asset: MoveStructId | AccountAddressInput;
+}): Promise<number> {
+  const { aptosConfig, accountAddress, asset } = args;
+
+  const response = await getAptosFullNode<{}, number>({
+    aptosConfig,
+    originMethod: "GET",
```
gregnazario, line 770: This is actually not a method, it should be `getBalance`

src/internal/account.ts
```diff
@@ -745,6 +745,39 @@ export async function getAccountCoinsCount(args: {
...
+    path: `accounts/${accountAddress}/balance/${asset}`,
+    params: {
+      accountAddress: accountAddress.toString(),
+      asset: asset instanceof Uint8Array ? AccountAddress.from(asset).toString() : asset.toString(),
+    },
+  });
+
+  return parseInt(response.data.toString(), 10);
```
gregnazario, line 778: This balance could be over the maximum int... I don't know what we want to do here though.  This could throw an error I think.

WGB5445: Yes, I noticed the others are also using this approach… I think switching to bigint would be better. Let’s update it to bigint in the next PR

gregnazario (approved):

PR #871, opened 2026-03-27
[encrypted-txns] add encrypted transaction payload support

kaw2k:
## Description                                                                                                                                                  
   
  Adds `options.encrypted: true` to transaction build, which encrypts the executable payload (entry function / script / multisig inner payload) before submission.
                                                            
  **Protocol flow:** The SDK builds a `DecryptedPlaintext` (executable + random 16-byte `decryption_nonce`), authenticates it with `PayloadAssociatedData` (sender address + all signer `(address, authenticationKey)` pairs) as AEAD, and encrypts it with the per-epoch batch encryption key fetched from the fullnode's ledger info. `RawTransaction` then carries `TransactionPayload::EncryptedPayload`: ciphertext, cleartext `extra_config` (e.g. orderless replay nonce), `payload_hash` (domain-separated BCS hash of the plaintext), and the `encryptionEpoch` for validator key routing. Validators decrypt after batching; the client never decrypts.

  **`ClaimedEntryFunction`:** An optional cleartext field (`claimed_entry_fun`) that exposes the module (and optionally function name) without decrypting, used so fee payers and multisig co-signers can identify the call without seeing the full payload.

  **Required build options when `encrypted: true`:**
  - `options.authenticationKey`: sender's auth key. Accepts an `AccountPublicKey` (auth key derived automatically via `.authKey()`) or a raw 32-byte hex string.
  Must match the signing authenticator.
  - `options.secondarySignerAuthenticationKeys`: one per secondary signer, in address order. Same `AccountPublicKey | HexInput` union.
  - `options.feePayerAuthenticationKey`: required when a non-zero fee payer is present. Same union type.

  **Crypto primitives** (`src/core/crypto/encryption/`) are ported from `aptos-core/crates/aptos-batch-encryption/ts-batch-encrypt` and verified against the same
  BCS layouts and constants (`ID_HASH_DST`, G2 hash-to-curve). The canonical implementation lives in `aptos-core`; this copy will move exclusively to the TS SDK
  once it stabilizes.

  **Example usage:**
  ```typescript
  const transaction = await aptos.transaction.build.simple({
    sender: sender.accountAddress,
    data: {
      function: "0x1::aptos_account::transfer",
      functionArguments: [receiver.accountAddress, amount],
    },
    options: {
      encrypted: true,
      authenticationKey: sender.publicKey,
    },
  });
  await aptos.signAndSubmitTransaction({ signer: sender, transaction });
  ```

  **Key changes:**
  - `options.encrypted` flag on transaction build triggers encryption
  - `options.authenticationKey` accepts `AccountPublicKey` or raw hex — auth key derived automatically from public key when possible
  - `fetchAndCacheEncryptionKey` — fetches and epoch-caches the fullnode's batch encryption key from ledger info
  - Simulation is rejected with a clear error for encrypted payloads (not yet supported)
  - Orderless (replay-protection nonce in cleartext `extra_config`, `u64::MAX` sequence) + encrypted supported together

 **TODO:**
  - Transaction simulation (requires protocol-level support; currently rejected with a clear error)
  - Encrypted multisig is structurally handled (multisig address preserved in cleartext `extra_config`, `claimed_entry_fun` populated) but untested end-to-end, pending fullnode support

  ## Test plan

  - `pnpm test` (localnet): encrypted e2e suite is skipped when `APTOS_NETWORK` is unset or `local` (localnet lacks encrypted-tx support); rest of suite unchanged
  - Unit (`tests/unit/core/crypto/encryption/`, `tests/unit/transactions/`): BCS round-trips for `DecryptedPlaintext`, `PayloadAssociatedData`,
  `TransactionPayloadEncryptedPayload`, `TransactionExtraConfigV1`; batch-encryption constants cross-checked against `aptos-core`
  - E2E (`APTOS_NETWORK=devnet`, `pnpm test-encrypted`): ledger `encryption_key` field, `fetchAndCacheEncryptionKey`, encrypted build produces correct payload
  type, simulate rejects with clear error, sign+submit succeeds, plain and encrypted orderless (including BCS round-trip on raw txn)

  ## Checklist


src/utils/const.ts
```diff
@@ -37,6 +37,14 @@ export const DEFAULT_MAX_GAS_AMOUNT = 2000000;
  */
 export const MIN_MAX_GAS_AMOUNT = 2000;
 
+/**
+ * Minimum gas unit price (Octas per gas unit) for encrypted transactions.
+ * Encrypted transactions require 2× the network base minimum (100) because validators
+ * bear the decryption compute cost (aptos-core `encrypted_txn_min_price_per_gas_unit`,
+ * RELEASE_V1_45+). The node rejects encrypted submissions below this price.
+ */
+export const MIN_ENCRYPTED_TXN_GAS_UNIT_PRICE = 200;
```
gregnazario, line 46: Interesting did not know encrypted is minimum 2x

src/core/crypto/encryption/symmetric.ts
```diff
@@ -0,0 +1,143 @@
...
+    const derived = hmacKdf(otpSource);
+    return new OneTimePad(derived.slice(0, 16));
+  }
+
+  padKey(value: SymmetricKey): SymmetricKey {
+    const paddedKey = new Uint8Array(16);
+    for (let i = 0; i < 16; i++) {
+      paddedKey[i] = value.key[i] ^ this.otp[i];
+    }
+    return new SymmetricKey(paddedKey);
```
gregnazario, line 49: Lots of `16` maybe make it a constant

src/core/crypto/encryption/symmetric.ts
```diff
@@ -0,0 +1,143 @@
...
+  constructor(nonce: Uint8Array, ctBody: Uint8Array) {
+    super();
+    if (nonce.length !== 12) {
+      throw new Error("Nonce must be 12 bytes");
+    }
+    this.nonce = nonce;
+    this.ctBody = ctBody;
+  }
+
+  serialize(serializer: Serializer): void {
```
gregnazario, line 74: similarly with 12

src/core/authenticationKey.ts
```diff
@@ -59,28 +59,33 @@ export class AuthenticationKey extends Serializable {
...
-   *
-   * @param serializer - The serializer instance used to perform the serialization.
-   * @group Implementation
-   * @category Serialization
+   * BCS wire format: ULEB128 length `0x20` then 32 bytes. Matches Rust `AuthenticationKey`'s
+   * `serde_bytes`-derived `Serialize`. Used by `PayloadAssociatedData::V1.signer_auth_keys`.
    */
   serialize(serializer: Serializer): void {
-    serializer.serializeFixedBytes(this.data.toUint8Array());
+    serializer.serializeBytes(this.data.toUint8Array());
```
gregnazario, line 69: oh what?

AuthenticationKey is supposed to be 32 opaque bytes.  It's the hash of a public key with some info afaik.

gregnazario (commented):
Other  than the auth key, seems fine.  though there's a lot of crypto code

src/core/authenticationKey.ts
kaw2k: This part might be off, but the references claude was digging up seemed appropriate. Will revert if you would like. For reference, here were the findings:

## Claude

It's still 32 opaque bytes semantically, but the BCS wire format is length-prefixed. `AuthenticationKey` derives [`SerializeKey`](https://github.com/aptos-labs/aptos-core/blob/cbebe66379dc1b2dec95748a0f6340c634a7ac8d/types/src/transaction/authenticator.rs#L939-L953), whose macro emits `serialize_newtype_struct(name, serde_bytes::Bytes::new(&bytes))` for non-human-readable serializers ([`aptos-crypto-derive`](https://github.com/aptos-labs/aptos-core/blob/cbebe66379dc1b2dec95748a0f6340c634a7ac8d/crates/aptos-crypto-derive/src/lib.rs#L200-L206)):

```rust
serializer.serialize_newtype_struct(
    #name_string,
    serde_bytes::Bytes::new(&ValidCryptoMaterial::to_bytes(self).as_slice()),
)
```
In BCS, `newtype_struct` is transparent and `serde_bytes::Bytes → serialize_bytes → ULEB128(len) || bytes`, so on the wire it's 0x20 + 32 bytes.

This is what we need to match for `signer_auth_keys: Vec<(AccountAddress, AuthenticationKey)>` in [`PayloadAssociatedData`](https://github.com/aptos-labs/aptos-core/blob/cbebe66379dc1b2dec95748a0f6340c634a7ac8d/types/src/transaction/encrypted_payload.rs#L49-L55). The previous serializeFixedBytes was off by the 1-byte length prefix.

src/utils/const.ts
kaw2k: Yep! [reference](https://github.com/aptos-labs/aptos-core/blob/cbebe66379dc1b2dec95748a0f6340c634a7ac8d/aptos-move/aptos-gas-schedule/src/gas_schedule/transaction.rs#L296-L300)

src/core/authenticationKey.ts
gregnazario: So, this is  mixing up stuff afaik, it's not really a key... but I can look at it agin.

There should be no prefix.

kaw2k: Updated

src/internal/encryptionKey.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @param args.aptosConfig - Aptos configuration.
+ */
+export async function fetchAndCacheEncryptionKey(args: {
+  aptosConfig: AptosConfig;
+}): Promise<{ key: EncryptionKey; epoch: bigint } | null> {
+  const { aptosConfig } = args;
+  const cacheKey = `encryption-key-${aptosConfig.fullnode ?? aptosConfig.network}`;
+
+  const ledgerInfo = await getLedgerInfo({ aptosConfig });
```
ibalajiarun, line 35: Is getLedgerInfo calling remote API endpoint? Do we want to ping API every time we want to do an epoch check?

ibalajiarun (commented):
beautiful code

src/internal/encryptionKey.ts
kaw2k: It does call the remote api, however we memoize the ledger info with a 10s ttl. If that data is stale we fetch it up before hand.

gregnazario (approved):

gregnazario:
@cursoragent please add an ignore on the staking e2e test, if it's dependent with on-chain state, it's likely that state changed

PR #885, opened 2026-04-13
feat!: v7.0.0 — ESM-only, tree-shakeable SDK with sub-path exports

gregnazario:
## Summary

Major modernization of the SDK build, module structure, and developer experience. **Shipping as v7.0.0.**

### Build
- **ESM-only** — removed CJS output, requires Node.js 22+
- **Plain `tsc` build** — replaced tsup/esbuild with plain TypeScript compiler. Unbundled, unminified 1:1 file mapping. `nodenext` module resolution with `.js` extensions
- **Removed dependencies**: `js-base64` (native `atob`/`btoa`), `tsup`, `esbuild`
- **Removed stale `sucrase` pnpm override**
- **Zero audit vulnerabilities** — overrides for lodash, picomatch, brace-expansion, yaml; bumped vite + @vitest/coverage-v8

### Tree-shaking & Sub-path Exports
- **Three usage tiers**:
  1. **`Aptos` class** — convenience, all-in-one (not tree-shakeable)
  2. **Namespace classes from sub-paths** — `import { General, AptosConfig } from "@aptos-labs/ts-sdk/general"` (tree-shakeable + autocomplete)
  3. **Standalone functions** — `import { getLedgerInfo } from "@aptos-labs/ts-sdk/general"` (maximum tree-shaking)
- **Poseidon isolation** — keyless/poseidon code excluded from barrel to keep main bundle small
- **`sideEffects: false`** for bundler tree-shaking

### Runtime Compatibility
- Supports **browsers, React Native, Node.js 22+, Bun, and Deno**
- No Node-only APIs in `src/` (no `Buffer`, no `node:` imports)
- Documented in CLAUDE.md and AGENTS.md to prevent regressions

### Dependencies
- **@noble/curves + @noble/hashes** upgraded to 2.x
- **@scure/bip32 + @scure/bip39** upgraded to 2.x

### Breaking Changes

See `upgrade-guides/UPGRADE_GUIDE_7.0.0.md` for migration details:
- ESM-only (no CJS `require()`)
- Node.js 22+ required
- Keyless, poseidon, HD key imports moved out of main barrel
- `generateSignedTransactionForSimulation` is now async
- Standalone functions only available from sub-path imports
- `AccountSequenceNumber.lastUncommintedNumber` renamed to `lastUncommittedNumber` (typo fix)

### Version note

`package.json` currently reads `6.3.1` to keep the review diff small. It will be bumped to `7.0.0` via `pnpm update-version` right before merge — that script keeps `package.json`, `src/version.ts`, and `docs/` in sync, and CI's `pnpm check-version` enforces the three stay aligned.

## Test Plan

- [x] `pnpm build` — clean tsc build
- [x] `pnpm test` — unit tests passing (incl. new `base64Url*` regression coverage)
- [x] `pnpm check` — biome lint + format clean
- [x] `pnpm audit` — zero vulnerabilities
- [x] `node tests/bundle-size/check-tree-shaking.mjs` — poseidon isolation + entry points verified
- [ ] CI: web tests (Playwright), Bun tests, Deno tests
- [ ] Manual: verify sub-path imports in a consumer project


README.md
```diff
@@ -33,47 +33,68 @@ The SDK is compatible with the [Bun](https://bun.sh/) runtime. Install the SDK u
...
+const config = new AptosConfig({ network: Network.TESTNET });
+const ledger = await getLedgerInfo({ aptosConfig: config });
+```
+
+#### For Bun users
+
+Bun's HTTP/2 support is not fully mature yet. Disable HTTP/2:
+
+```ts
+const config = new AptosConfig({ network: Network.TESTNET, clientConfig: { http2: false } });
```
gregnazario: Note, we can remove this note.  Aptos-client 4.0 uses the appropriate `fetch` module, and this no longer applies

package.json
```diff
@@ -51,46 +110,35 @@
...
-  "pnpm": {
-    "overrides": {
-      "sucrase": "3.35.1",
-      "lodash": "4.17.23",
-      "js-yaml": "4.1.1",
-      "diff": "4.0.4",
-      "immutable": "^4.3.8"
-    }
-  }
+  "version": "7.0.0"
```
gregnazario: Can we change the PR description to match this appropriately?

CLAUDE.md
```diff
@@ -95,7 +113,7 @@ When releasing a new version with breaking changes:
 
 ## Bun Compatibility
 
-When using with Bun, disable HTTP/2:
+When using with Bun, users must disable HTTP/2:
 ```typescript
 const aptos = new Aptos(new AptosConfig({ network: Network.TESTNET, clientConfig: { http2: false } }));
```
gregnazario: This doesn't apply because of the new version of Aptos client 4.0.0

upgrade-guides/UPGRADE_GUIDE_8.0.0.md
```diff
@@ -0,0 +1,130 @@
+# Upgrade Guide: v7.x → v8.0.0
+
+## Overview
+
+v8.0.0 makes the SDK tree-shakeable and modernizes the build output. The SDK now uses plain `tsc` (no bundler), `nodenext` module resolution, and provides three tiers of usage for different bundle-size needs.
```
gregnazario: We should only be doing 6 -> 7

tests/unit/secp256r1.test.ts
```diff
@@ -32,8 +32,8 @@ describe("Secp256r1PublicKey", () => {
     const uncompressedPublicKey = Hex.fromHexInput(singleSignerSecp256r1.publicKey);
     expect(uncompressedPublicKey.toUint8Array().length).toEqual(65);
 
-    const point = p256.ProjectivePoint.fromHex(uncompressedPublicKey.toUint8Array());
-    const compressedPublicKey = point.toRawBytes(true);
+    const point = p256.Point.fromHex(uncompressedPublicKey.toString().replace("0x", ""));
```
gregnazario: there should be a function in Hex that lets you output without the 0x

tests/unit/secp256k1.test.ts
```diff
@@ -33,8 +33,8 @@ describe("Secp256k1PublicKey", () => {
     const uncompressedPublicKey = Hex.fromHexInput(secp256k1TestObject.publicKey);
     expect(uncompressedPublicKey.toUint8Array().length).toEqual(65);
 
-    const point = secp256k1.ProjectivePoint.fromHex(uncompressedPublicKey.toUint8Array());
-    const compressedPublicKey = point.toRawBytes(true);
+    const point = secp256k1.Point.fromHex(uncompressedPublicKey.toString().replace("0x", ""));
```
gregnazario: similarly

tests/e2e/transaction/transactionSubmission.test.ts
```diff
@@ -1082,8 +1082,7 @@ describe("transaction submission", () => {
       toBeSigned.set(clientHash, authenticatorData.length);
       const webauthnDigest = sha256(toBeSigned);
       const privBytes = Hex.fromHexInput(privateKey.toHexString()).toUint8Array();
-      const sig = p256.sign(webauthnDigest, privBytes).normalizeS();
-      const signatureBytes = sig.toCompactRawBytes();
+      const signatureBytes = p256.sign(webauthnDigest, privBytes, { prehash: false });
```
gregnazario, line 1085: why is prehash false?

tests/e2e/transaction/helper.ts
```diff
@@ -394,4 +393,8 @@ export async function publishHelloWorldAAPackage(aptos: Aptos, senderAccount: Ac
   );
 }
 
-export const b64urlEncode = (u8: Uint8Array) => Base64.fromUint8Array(u8, true);
+export const b64urlEncode = (u8: Uint8Array) =>
+  btoa(String.fromCharCode(...u8))
+    .replace(/\+/g, "-")
+    .replace(/\//g, "_")
+    .replace(/=+$/, "");
```
gregnazario: Does this work in node though?

I want to make sure that this encoding works in node, browser, bun, react-native, deno

src/utils/helpers.ts
```diff
@@ -106,23 +105,19 @@ export function floorToWholeHour(timestampInSeconds: number): number {
...
-    base64 += "=";
+  const str = base64UrlDecode(base64Url);
+  const bytes = new Uint8Array(str.length);
+  for (let i = 0; i < str.length; i++) {
+    bytes[i] = str.charCodeAt(i);
   }
-  // Use js-base64 to convert base64 to Uint8Array
-  return Base64.toUint8Array(base64);
+  return bytes;
 }
```
gregnazario, line 166: Can we consolidate base64 conversions, and ensure they work on all platforms node, browser, react-native, deno, bun

src/internal/keyless.ts
```diff
@@ -278,7 +281,7 @@ export async function updateFederatedKeylessJwkSetTransaction(args: {
     });
   }
 
-  const jwks: JWKS = await response.json();
+  const jwks = (await response.json()) as JWKS;
```
gregnazario, line 284: why is this a cast now?

src/internal/ans.ts
```diff
@@ -135,10 +135,11 @@ export function getANSExpirationStatus({
 }
 
 export const LOCAL_ANS_ACCOUNT_PK =
-  process.env.ANS_TEST_ACCOUNT_PRIVATE_KEY ??
+  (typeof process !== "undefined" && process.env?.ANS_TEST_ACCOUNT_PRIVATE_KEY) ||
   "ed25519-priv-0x37368b46ce665362562c6d1d4ec01a08c8644c488690df5a17e13ba163e20221";
 export const LOCAL_ANS_ACCOUNT_ADDRESS =
-  process.env.ANS_TEST_ACCOUNT_ADDRESS ?? "0x585fc9f0f0c54183b039ffc770ca282ebd87307916c215a3e692f2f8e4305e82";
+  (typeof process !== "undefined" && process.env?.ANS_TEST_ACCOUNT_ADDRESS) ||
+  "0x585fc9f0f0c54183b039ffc770ca282ebd87307916c215a3e692f2f8e4305e82";
```
gregnazario: Seems excessive... is this necessary?

src/core/crypto/hdKey.ts
```diff
@@ -85,8 +85,12 @@ export function isValidHardenedPath(path: string): boolean {
  * @group Implementation
  * @category Serialization
  */
+const textEncoder = new TextEncoder();
+const toBytes = (input: Uint8Array | string): Uint8Array =>
+  typeof input === "string" ? textEncoder.encode(input) : input;
+
```
gregnazario: Don't redefine this, use `Hex.from().toUint8Array()`

src/core/crypto/keyless.ts
```diff
@@ -493,7 +492,8 @@ function getPublicInputsHash(args: {
     fields.push(1n);
     fields.push(hashStrToField(proof.extraField, keylessConfig.maxExtraFieldBytes));
   }
-  fields.push(hashStrToField(`${encode(signature.jwtHeader, true)}.`, keylessConfig.maxJwtHeaderB64Bytes));
+  const jwtHeaderB64Url = btoa(signature.jwtHeader).replace(/\+/g, "-").replace(/\//g, "_").replace(/=+$/, "");
+  fields.push(hashStrToField(`${jwtHeaderB64Url}.`, keylessConfig.maxJwtHeaderB64Bytes));
```
gregnazario: Once again, centralize with the other base 64, ensure that it runs on all platforms

src/core/crypto/secp256k1.ts
```diff
@@ -79,7 +79,7 @@ export class Secp256k1PublicKey extends PublicKey {
     const messageBytes = Hex.fromHexInput(messageToVerify).toUint8Array();
     const messageSha3Bytes = sha3_256(messageBytes);
     const signatureBytes = signature.toUint8Array();
-    return secp256k1.verify(signatureBytes, messageSha3Bytes, this.key.toUint8Array(), { lowS: true });
+    return secp256k1.verify(signatureBytes, messageSha3Bytes, this.key.toUint8Array(), { lowS: true, prehash: false });
```
gregnazario, line 82: interesting, why is prehash false?

src/core/crypto/singleKey.ts
gregnazario, line 79: Do these load the keyless code if necessary?

src/transactions/transactionBuilder/transactionBuilder.ts
```diff
@@ -795,10 +815,12 @@ export function generateSignedTransaction(args: InputSubmitTransactionData): Uin
  * @group Implementation
  * @category Transactions
  */
+const textEncoder = new TextEncoder();
+
 export function hashValues(input: (Uint8Array | string)[]): Uint8Array {
   const hash = sha3Hash.create();
   for (const item of input) {
-    hash.update(item);
+    hash.update(typeof item === "string" ? textEncoder.encode(item) : item);
```
gregnazario: Let's have one textEncoder in utils for converting, so we don't have multiple.

tests/e2e/transaction/transactionSubmission.test.ts
gregnazario: The SDK pre-hashes the message with `sha3_256` before calling noble's `verify`/`sign` — see e.g. `secp256k1.ts:80` where `sha3_256(messageToVerify)` runs first. `@noble/curves` v2 changed the default so `prehash` is now `true`, which would double-hash. Passing `{ prehash: false }` tells noble to treat the input as the already-computed digest.

src/core/crypto/secp256k1.ts
gregnazario: `@noble/curves` v2 changed the default so `verify`/`sign` run their own hash over the input (`prehash: true`). The SDK already pre-hashes on line 80 (`sha3_256(messageToVerify)`) to produce the signing digest, so we need `{ prehash: false }` to stop noble from double-hashing the digest. Same pattern applies in secp256r1 and the WebAuthn test.

tests/e2e/transaction/helper.ts
gregnazario: Addressed in a follow-up: removed the duplicate and centralized as `base64UrlEncode(input: string | Uint8Array)` in `src/utils/helpers.ts`. It uses `btoa` + `String.fromCharCode` (universal in Node ≥ 16, browsers, Bun, Deno, RN 0.74+) and a `TEXT_ENCODER`-based UTF-8 path for strings. These tests now import `base64UrlEncode` directly from the SDK.

src/core/crypto/singleKey.ts
gregnazario: Yes. `AnyPublicKey`/`AnySignature` use a runtime variant registry (`src/core/crypto/anyKeyRegistry.ts`). The keyless module self-registers at module load time — so `import ... from "@aptos-labs/ts-sdk/keyless"` triggers registration, and from that point on `singleKey.ts:79` can decode the keyless variants. Consumers who never touch keyless don't pay the `poseidon-lite` cost because the registry stays empty of those variants.

upgrade-guides/UPGRADE_GUIDE_7.0.0.md
```diff
@@ -0,0 +1,162 @@
...
+
+```typescript
+import { getLedgerInfo } from "@aptos-labs/ts-sdk/general";
+import { getAccountInfo } from "@aptos-labs/ts-sdk/account";
+import { AptosConfig, Network } from "@aptos-labs/ts-sdk";
+
+const config = new AptosConfig({ network: Network.TESTNET });
+await getLedgerInfo({ aptosConfig: config });
+await getAccountInfo({ aptosConfig: config, accountAddress: "0x1" });
+```
```
gregnazario, line 131: Does this allow for sharing of an aptos client?  Or is it recreated on every standalone function?

Should it cache those per endpoint / per config?

gregnazario: The HTTP client lives on `AptosConfig` — instantiated once in its constructor (`src/api/aptosConfig.ts:187`). Every standalone function takes an `{ aptosConfig }` arg and `aptosRequest` reads `aptosConfig.client` directly, so as long as callers reuse the same config, the client (and its connection pool) is shared.

No per-endpoint/per-config cache today — reuse is at the `AptosConfig` level. If we wanted to centralize further (e.g. a process-wide config→client map), we could add that, but today the pattern is the same as the class-based API where one `Aptos` wraps one `AptosConfig`.

confidential-asset/tests/helpers/index.ts
```diff
@@ -1,6 +1,10 @@
 import path from "path";
 import fs from "fs";
 import { execSync } from "child_process";
+import { fileURLToPath } from "url";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
 import {
   Network,
   NetworkToNetworkName,
```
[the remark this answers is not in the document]
gregnazario: Fixed in 79144fad9 — moved `import { Network, ... }` before the `const __filename/__dirname` lines.

docs/@aptos-labs/ts-sdk-7.0.0/classes/AccountAuthenticator.html
```diff
@@ -0,0 +1,36 @@
...
+</div><h4 class="tsd-returns-title">Returns <span class="tsd-signature-type">string</span></h4><p>the hex formatas a string prefixed by <code>0x</code>.</p>
```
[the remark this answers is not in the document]
gregnazario: Fixed at source in c354aa5e8 — `src/bcs/serializer.ts:73` JSDoc now reads "hex format as a string". The generated HTML will match on next `pnpm doc`.

docs/@aptos-labs/ts-sdk-7.0.0/classes/AccountSequenceNumber.html
```diff
@@ -0,0 +1,47 @@
...
+<a href="#update" class="tsd-index-link"><svg class="tsd-kind-icon" viewBox="0 0 24 24" aria-label="Method"><use href="../assets/icons.svg#icon-2048"></use></svg><span>update</span></a>
+</div></section><section class="tsd-index-section"><h3 class="tsd-index-heading">Properties</h3><div class="tsd-index-list"><a href="#account" class="tsd-index-link"><svg class="tsd-kind-icon" viewBox="0 0 24 24" aria-label="Property"><use href="../assets/icons.svg#icon-1024"></use></svg><span>account</span></a>
+<a href="#aptosconfig" class="tsd-index-link"><svg class="tsd-kind-icon" viewBox="0 0 24 24" aria-label="Property"><use href="../assets/icons.svg#icon-1024"></use></svg><span>aptos<wbr/>Config</span></a>
+<a href="#currentnumber" class="tsd-index-link"><svg class="tsd-kind-icon" viewBox="0 0 24 24" aria-label="Property"><use href="../assets/icons.svg#icon-1024"></use></svg><span>current<wbr/>Number</span></a>
+<a href="#lastuncommintednumber" class="tsd-index-link"><svg class="tsd-kind-icon" viewBox="0 0 24 24" aria-label="Property"><use href="../assets/icons.svg#icon-1024"></use></svg><span>last<wbr/>Uncomminted<wbr/>Number</span></a>
```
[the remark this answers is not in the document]
gregnazario: Renamed in a0420fa31 as a breaking change for v7.0.0 (the existing `// TODO: Change to Uncommitted` comment confirmed the intent). No deprecation alias — major version bump is the clean break. Documented in CHANGELOG Breaking section and `upgrade-guides/UPGRADE_GUIDE_7.0.0.md` §6.

docs/@aptos-labs/ts-sdk-7.0.0/assets/icons.js
```diff
@@ -0,0 +1,18 @@
...
+        svg.style.display = "none";
+        if (location.protocol === "file:") updateUseElements();
+    }
+
+    function updateUseElements() {
+        document.querySelectorAll("use").forEach(el => {
+            if (el.getAttribute("href").includes("#icon-")) {
+                el.setAttribute("href", el.getAttribute("href").replace(/.*#/, "#"));
```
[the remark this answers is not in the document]
gregnazario: `docs/@aptos-labs/ts-sdk-7.0.0/assets/icons.js` is a TypeDoc-generated asset — regenerated by `pnpm doc` from upstream TypeDoc, so we can't modify it here without the fix landing upstream. Won't-fix in this PR.

CHANGELOG.md
```diff
@@ -7,15 +7,37 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
...
+- **Tree-shakeable standalone function API** — all SDK operations available as standalone functions (`getLedgerInfo`, `getBalance`, `transferCoinTransaction`, etc.) that accept `{ aptosConfig, ...args }`.
+- **Sub-path exports** — import from `@aptos-labs/ts-sdk/account`, `@aptos-labs/ts-sdk/transaction`, `@aptos-labs/ts-sdk/keyless`, etc. for minimal bundle sizes.
+- **`sideEffects: false`** in package.json for bundler tree-shaking.
+- **Variant registry pattern** for `AnyPublicKey`/`AnySignature` — keyless variants register at runtime, removing compile-time poseidon dependency from core crypto.
+
+## Breaking
+
+- ESM-only output; removed CommonJS `"require"` exports. Node.js 22+ required.
+  - Migration: Update imports from `require()` to `import` syntax
+  - See: `upgrade-guides/UPGRADE_GUIDE_7.0.0.md`
```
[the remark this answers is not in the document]
gregnazario: CHANGELOG already points to `UPGRADE_GUIDE_7.0.0.md`; I'll align the PR description/title to v7.0.0 to match.

kaw2k (commented):
Some curated feedback from claude 4.7 that would be worth investigating:

### `Aptos` sub-modules break the mixin `applyMixin` contract for non-method properties

```ts
// src/api/aptos.ts:184-196
applyMixin(Aptos, Account, "account");
applyMixin(Aptos, AccountAbstraction, "abstraction");
...
```

`applyMixin` iterates `Object.getOwnPropertyNames(baseClass.prototype)` which includes `"constructor"` and any accessors/getters. Each is redefined on `Aptos.prototype` as a plain data property forwarding to `this[baseClassProp][name](...args)`. Consequences:

- `Aptos.prototype.constructor` gets replaced with a delegating function (bug, but pre-existing).
- If any of the 13 namespace classes introduce a getter/setter, it becomes a broken function call.
- `AccountAbstraction` is mixed in (`applyMixin(Aptos, AccountAbstraction, "abstraction")`), but the `Aptos` interface declaration on line 151 does **not** extend `AccountAbstraction`. The runtime has the methods, types don't — so `aptos.enableAccountAbstraction(...)` compiles into an error for users who relied on it pre-refactor. Either drop the mixin line or add `AccountAbstraction` to the `extends` list.

### `/account` and other sub-paths still transitively import `poseidon-lite`

`functions/account.ts` re-exports from `internal/account.ts`. That file uses `KeylessAccount` and `FederatedKeylessAccount` as runtime values:

```ts
// src/internal/account.ts
if (signer instanceof KeylessAccount || signer instanceof FederatedKeylessAccount) { ... }
...
FederatedKeylessAccount.create({ ... })
...
KeylessAccount.create({ ... })
```

Any consumer importing `getAccountInfo` from `@aptos-labs/ts-sdk/account` pulls `internal/account.ts`, which pulls both keyless account modules, which pull `AbstractKeylessAccount`, which pulls `poseidon-lite`. The `sideEffects: false` hint does not help: these are value references, not side-effects.

The tree-shaking test at `tests/bundle-size/check-tree-shaking.mjs` only inspects `functions/account.js` (a file that is literally just re-export statements) for the string "poseidon" — which is a false-positive check. It does not follow transitive imports.

Suggested remediation: move the keyless branches in `internal/account.ts` behind dynamic `import()` (same pattern used in `internal/transactionSubmission.ts`) or split the keyless-specific helpers into `internal/keylessAccount.ts`.

### Runtime TS cast in `AptosConfig` masks a real API mismatch

```ts
// src/api/aptosConfig.ts:185-187
this.client = settings?.client ?? { provider: aptosClient as unknown as Client["provider"] };
```

`@aptos-labs/aptos-client` was bumped from `^2.1.0` to `^4.0.0` — a two-major version jump. The comment acknowledges types "don't resolve correctly under nodenext module resolution", but the cast suppresses a genuine signature change. If `aptosClient`'s runtime call shape no longer matches the `Client["provider"]` interface, every HTTP call goes through a silently-broken adapter. This should be verified against the v4 client (arg order, return shape), and the types either aligned or the adapter updated — the cast should not remain.

### `tsconfig.build.json` inherits `types: ["node", "vitest/globals"]`

```json
// tsconfig.json
"types": ["node", "vitest/globals"]
```

`tsconfig.build.json` extends `tsconfig.json` without overriding this. TypeScript will include these ambient types during production build. The PR already had to special-case `TEXT_ENCODER` (commit `e1be60527`) to avoid leaking `import("util").TextEncoder`. Other globals (`process` in `helpers.ts`) can still cause emitted `.d.ts` files to carry `/// <reference types="node" />`, forcing downstream consumers to install `@types/node`. For a runtime-agnostic SDK this is a regression.

Fix: in `tsconfig.build.json` override `"types": []` (or `["node"]` only where needed for `src/cli/`) and compile CLI separately if necessary.

README.md
```diff
@@ -33,47 +33,68 @@ The SDK is compatible with the [Bun](https://bun.sh/) runtime. Install the SDK u
...
-Bun's HTTP/2 support is not fully mature yet. You need to disable HTTP/2 to ensure the SDK works properly:
+Import only the namespaces you need for smaller bundles with full autocomplete:
 
 ```ts
-import { Aptos, AptosConfig, Network } from "@aptos-labs/ts-sdk";
+import { General, AptosConfig } from "@aptos-labs/ts-sdk/general";
+import { Faucet } from "@aptos-labs/ts-sdk/faucet";
+
+const config = new AptosConfig({ network: Network.TESTNET });
+const general = new General(config);
```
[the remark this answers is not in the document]
gregnazario: After the README refresh just pushed in this iteration (commit on the way), the Option 2 and Option 3 snippets now explicitly `import { Network } from "@aptos-labs/ts-sdk"`, so the examples do compile as written. Leaving the rest of the Option 3 copy edited to explain that mixing barrel imports still tree-shakes (package is `sideEffects: false`). Thanks for flagging.

src/utils/helpers.ts
```diff
@@ -106,23 +105,19 @@ export function floorToWholeHour(timestampInSeconds: number): number {
...
-  // Replace base64url-specific characters
+  // Convert base64url to standard base64, then decode with universal atob
   const base64 = base64Url.replace(/-/g, "+").replace(/_/g, "/");
-  // Pad the string with '=' characters if needed
-  const paddedBase64 = base64 + "==".substring(0, (3 - (base64.length % 3)) % 3);
-  const decodedString = decode(paddedBase64);
-  return decodedString;
+  const padded = base64 + "==".substring(0, (3 - (base64.length % 3)) % 3);
+  return atob(padded);
 }
```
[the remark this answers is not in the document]
gregnazario: Already fixed in this PR — `base64UrlDecode` now decodes via `base64UrlToBytes` + `new TextDecoder("utf-8").decode(...)` so UTF-8 JWT payloads survive the round-trip. Byte-level callers continue to use `base64UrlToBytes`. Added a regression test in `tests/unit/helpers.test.ts` that round-trips multi-byte UTF-8 through `base64UrlEncode`/`base64UrlDecode`.

src/index.ts
```diff
@@ -1,13 +1,36 @@
...
-export * from "./client";
-export * from "./core";
-export * from "./errors";
-export * from "./transactions";
-export * from "./transactions/management";
-export * from "./types";
-export * from "./utils";
+// NOTE: Standalone functions are NOT re-exported from the barrel to avoid circular
+// dependency issues. Import from '@aptos-labs/ts-sdk/functions' or sub-path exports.
+
```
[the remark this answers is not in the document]
gregnazario: Fixed in the push I am sending now — the comment previously said `@aptos-labs/ts-sdk/functions`, which is not exported. Updated to point at the real per-namespace sub-paths (`@aptos-labs/ts-sdk/general`, `/account`, `/keyless`, etc.) and explicitly reference `package.json#exports` as the source of truth.

upgrade-guides/UPGRADE_GUIDE_8.0.0.md
```diff
@@ -0,0 +1,130 @@
...
+// Keyless functions from the keyless sub-path
+import { deriveKeylessAccount } from "@aptos-labs/ts-sdk/keyless";
+
+// Keyless account classes directly
+import { KeylessAccount } from "@aptos-labs/ts-sdk/dist/account/KeylessAccount.js";
+import { EphemeralKeyPair } from "@aptos-labs/ts-sdk/dist/account/EphemeralKeyPair.js";
+
+// Poseidon utilities directly
+import { poseidonHash } from "@aptos-labs/ts-sdk/dist/core/crypto/poseidon.js";
+```
```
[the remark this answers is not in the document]
gregnazario: `upgrade-guides/UPGRADE_GUIDE_8.0.0.md` has been deleted in this PR — we decided to release this as v7.0.0 (see changelog + PR title). The v7 guide only documents the sub-path exports that are actually in `package.json#exports`, so the `./dist/*` deep-import concern no longer applies.

package.json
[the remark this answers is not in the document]
gregnazario: Version decision: we are shipping this as **v7.0.0**, not v8.0.0 (package.json currently reads 6.3.1 for review diff size; it will be bumped to 7.0.0 via `pnpm update-version` right before merge, which also regenerates `src/version.ts` and `docs/` — `pnpm check-version` enforces consistency in CI). The CHANGELOG, PR title, and `upgrade-guides/UPGRADE_GUIDE_7.0.0.md` all already point to 7.0.0.

README.md
gregnazario: Verified — the old "Bun: disable http2" note no longer exists anywhere in `README.md`; `rg -n "http2|aptos-client" README.md` returns nothing. Also removed the same obsolete note from `AGENTS.md` in this iteration.

CLAUDE.md
gregnazario: Verified — `CLAUDE.md` no longer contains the http2 / aptos-client 4.0 workaround note; `rg -n "http2|aptos-client" CLAUDE.md` returns nothing. Also cleaned up the same stale note from `AGENTS.md` in this iteration for consistency.

package.json
gregnazario: Will update the PR title/description to v7.0.0 as part of this pass (done alongside this reply — see updated title and body).

upgrade-guides/UPGRADE_GUIDE_8.0.0.md
gregnazario: Done earlier in this PR — `upgrade-guides/UPGRADE_GUIDE_8.0.0.md` has been deleted and only `UPGRADE_GUIDE_7.0.0.md` remains. CHANGELOG line 19 already references `UPGRADE_GUIDE_7.0.0.md`.

tests/unit/secp256r1.test.ts
gregnazario: Done — switched every `singleSignerSecp256r1.publicKey.replace("0x", "")` / `signatureHex.replace("0x", "")` / `expectedHex.replace("0x", "")` in `tests/unit/secp256r1.test.ts` over to `Hex.fromHexInput(...).toStringWithoutPrefix()`. Pushed in the follow-up commit in this iteration.

tests/unit/secp256k1.test.ts
gregnazario: Checked `tests/unit/secp256k1.test.ts` — it already uses `Hex.fromHexInput(...).toStringWithoutPrefix()` (line 36) and no `.replace("0x", "")` patterns remain (`rg -n "replace..0x" tests/unit/secp256k1.test.ts` returns nothing). No change needed; the parallel fix landed in `secp256r1.test.ts` where the pattern actually existed.

src/utils/helpers.ts
gregnazario: Done earlier in this PR: `src/utils/helpers.ts` now exposes a single pair — `base64UrlEncode(input: string | Uint8Array)` + `base64UrlToBytes(base64Url)` + `base64UrlDecode(base64Url)` — all built on universal primitives (`atob` / `btoa` / `String.fromCharCode` / `TextEncoder` / `TextDecoder`) that work in Node 22+, browsers, Bun, Deno, and React Native 0.74+. The e2e `helper.ts` duplicate and the inline implementation inside `core/crypto/keyless.ts` were deleted and now import from the shared helper.

Follow-up in this iteration: fixed a latent padding bug where `base64UrlToBytes` was using `(3 - (len % 3)) % 3` instead of the correct `(4 - (len % 4)) % 4`. The old formula under-padded 2-byte-remainder inputs, corrupting things like RSA JWK `n` values (length % 4 == 2). Added regression tests in `tests/unit/helpers.test.ts` covering every padding length and a realistic JWT RSA modulus.

src/internal/keyless.ts
gregnazario: `response.json()` from the fetch API is typed as `Promise<any>`, so the cast narrows `jwks` to the local `JWKS` interface (defined just above at line 241) for type-safe access to `jwks.keys[...].{kid, alg, e, n}`. There is no runtime validation here — callers already trust the JWKS endpoint URL (hard-coded for Firebase, derived from `iss` otherwise), and if the shape is wrong the subsequent `MoveVector.MoveString(jwks.keys.map(...))` will throw. Happy to add runtime Zod-style validation if you want, but not blocking on it here.

src/internal/ans.ts
gregnazario: The `getEnvVar` helper is the `process.env` guard — it was added in this PR because `src/` has to stay runtime-agnostic (no raw `process.env` access — breaks Deno default perms, and strictly, browsers too). `src/internal/ans.ts` previously had inline `typeof process !== "undefined" && process.env?.ANS_TEST_ACCOUNT_*` which was the same pattern duplicated twice. If it still feels heavy given its one call-site here, I can inline back to `typeof process !== "undefined" ? process.env?.["ANS_TEST_ACCOUNT_PRIVATE_KEY"] : undefined`, but I think the helper is the cleaner API now that we enforce the guard SDK-wide.

src/core/crypto/hdKey.ts
gregnazario: Looked at this — the `toBytes` helper in `hdKey.ts` is not parsing hex. It converts a UTF-8 label (e.g. `KeyType.ED25519 = "ed25519 seed"` — the SLIP-0010 seed label) into raw UTF-8 bytes for HMAC-SHA512 input. `Hex.from("ed25519 seed")` would throw with "Hex characters are invalid" because that string is not hex. So using `TEXT_ENCODER.encode(input)` here is the correct semantics; switching to `Hex.from(...).toUint8Array()` would break key derivation. Happy to rename the local helper to `toUtf8Bytes` to make intent clearer if you want.

src/core/crypto/keyless.ts
gregnazario: Done — `src/core/crypto/keyless.ts` no longer defines its own base64 encoder. It now imports the shared `base64UrlEncode` / `base64UrlToBytes` from `src/utils/helpers.ts` (see `import { base64UrlEncode, base64UrlToBytes, nowInSeconds } from "../../utils/index.js";` at line 34). Same single universal implementation is used by `keyless.ts`, e2e tests, and any future callers.

src/transactions/transactionBuilder/transactionBuilder.ts
gregnazario: Done earlier in this PR — `src/utils/const.ts` exports a single shared `TEXT_ENCODER: { encode(input: string): Uint8Array } = new TextEncoder()`. Every prior `new TextEncoder()` in `src/` now imports that constant (BCS serializer, `signingMessage`, `transactionBuilder`, `hashValues`, `hdKey`, `poseidon`, `moveStructs`, `account/utils/address`, `remoteAbi`, `crypto/utils`, `helpers`, etc.). The explicit structural type also stops `@types/node`'s `import("util").TextEncoder` from leaking into the emitted `.d.ts`, which used to force every SDK consumer to have `@types/node` installed.

README.md
```diff
@@ -33,47 +33,61 @@ The SDK is compatible with the [Bun](https://bun.sh/) runtime. Install the SDK u
...
+const faucet = new Faucet(config);
+```
+
+### Option 3: Standalone functions (maximum tree-shaking)
+
+For the smallest possible bundles (e.g., wallet adapters):
```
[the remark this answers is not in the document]
gregnazario: Thanks — adopted (with minor edits). The Option 3 section now says: "`AptosConfig` / `Network` are also available from their sub-paths (`@aptos-labs/ts-sdk/general`), but importing them from the main barrel is fine — the package is marked `sideEffects: false`, so modern bundlers still tree-shake unused exports." The snippet imports `getLedgerInfo`/`AptosConfig` from `/general` and `Network` from the barrel.

CHANGELOG.md
```diff
@@ -7,15 +7,40 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
...
+- **Namespace classes available from sub-paths** — `import { General, AptosConfig } from "@aptos-labs/ts-sdk/general"` for tree-shakeable usage with autocomplete. The `Aptos` class remains available but is not tree-shakeable.
+- **Plain `tsc` build** — replaced tsup bundler with plain TypeScript compiler. Output is unbundled, unminified ESM with 1:1 file mapping. `nodenext` module resolution.
+- **Removed `js-base64` dependency** — replaced with native `atob`/`btoa` (universal across all runtimes).
+- **Keyless imports moved** — `KeylessAccount`, `FederatedKeylessAccount`, `EphemeralKeyPair`, poseidon utilities, and keyless crypto types are no longer exported from the main entry. Import from sub-paths or direct file paths.
+- **HD Key and deserialization utils moved** — no longer in the crypto barrel. Import directly from their files.
+- **`generateSignedTransactionForSimulation` is now async** — callers must `await` it.
+  - See: `upgrade-guides/UPGRADE_GUIDE_7.0.0.md`
+- Rename `AccountSequenceNumber.lastUncommintedNumber` → `lastUncommittedNumber` (typo fix).
```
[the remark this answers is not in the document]
gregnazario: The source rename (`AccountSequenceNumber.lastUncommintedNumber` → `lastUncommittedNumber`) and the JSDoc typo fix ("formatas" → "format as") are both in this PR at the source level (see CHANGELOG Breaking entry and `src/bcs/serializer.ts` JSDoc). The generated HTML pages under `docs/@aptos-labs/ts-sdk-7.0.0/` will be regenerated via `pnpm update-version`/`pnpm doc` right before merge (that step is the same one that bumps `package.json` + `src/version.ts` to 7.0.0). Keeping the stale docs in-tree for now keeps the review diff smaller; they'll be clean on the version-bump commit.

CHANGELOG.md
```diff
@@ -7,15 +7,40 @@ All notable changes to the Aptos TypeScript SDK will be captured in this file. T
...
+- Remove `dotenv` usage from all TypeScript/JavaScript examples. Node 22+ users can rely on the built-in `node --env-file=.env` flag (or `tsx --env-file=.env`) when a `.env` file is needed; the examples default to devnet and don't require one.
 
 ## Fixed
 
 - Fix `fullnodeConfig.HEADERS` not being forwarded in paginated fullnode requests (`getAccountModules`, `getAccountResources`). Pagination helpers now route through `getAptosFullNode` so custom headers (e.g. `Authorization`) are included consistently. (#872)
+- Fix `base64UrlDecode` to decode bytes as UTF-8 via `TextDecoder`. Previously returned a Latin-1 binary string from `atob`, corrupting non-ASCII characters in JWT headers/payloads.
+- Add `base64UrlEncode` helper and consolidate duplicated base64url encoders in `keyless.ts` and e2e tests.
+- Add `getEnvVar` helper for runtime-agnostic `process.env` access; simplifies `LOCAL_ANS_ACCOUNT_*` guards.
+- Consolidate `TextEncoder` instantiation to a shared `TEXT_ENCODER` constant across BCS, poseidon, signing, transaction builder, and HD key paths.
+- Fix "hex formatas a string" typo in `Serializable.toString()` JSDoc.
```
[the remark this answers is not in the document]
gregnazario: Same as the sibling comment on line 27: source rename + typo fix are both in this PR. The `docs/@aptos-labs/ts-sdk-7.0.0/` HTML will be regenerated by `pnpm update-version`/`pnpm doc` as part of the version-bump commit before merge. The rename and typo fix in the source will carry through to the generated pages at that point.

docs/@aptos-labs/ts-sdk-7.0.0/classes/AccountAuthenticator.html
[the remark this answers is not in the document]
gregnazario: Source already fixed in `src/bcs/serializer.ts` JSDoc (`@returns the hex format as a string prefixed by …`). The generated `docs/` HTML will be regenerated by `pnpm update-version` + `pnpm doc` right before merge as part of the 7.0.0 bump — intentionally deferred to keep the review diff smaller. Won't-fix in this patch; fixed on version-bump.

docs/@aptos-labs/ts-sdk-7.0.0/classes/AccountSequenceNumber.html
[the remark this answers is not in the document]
gregnazario: Source already renamed to `lastUncommittedNumber` in `src/transactions/management/accountSequenceNumber.ts` (Breaking change, see CHANGELOG line 27). The HTML under `docs/@aptos-labs/ts-sdk-7.0.0/` will regenerate on `pnpm update-version`/`pnpm doc` as part of the version-bump commit before merge. Won't-fix the generated HTML directly here.

src/utils/helpers.ts
```diff
@@ -106,23 +117,39 @@ export function floorToWholeHour(timestampInSeconds: number): number {
...
 export function base64UrlToBytes(base64Url: string): Uint8Array {
-  // Convert Base64Url to Base64
-  let base64 = base64Url.replace(/-/g, "+").replace(/_/g, "/");
-  // Add padding if needed
-  while (base64.length % 4 !== 0) {
-    base64 += "=";
+  // Convert base64url to standard base64, then decode with universal atob
+  const base64 = base64Url.replace(/-/g, "+").replace(/_/g, "/");
+  const padded = base64 + "==".substring(0, (3 - (base64.length % 3)) % 3);
+  const binary = atob(padded);
```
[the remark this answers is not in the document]
gregnazario: Fixed in this push: `base64UrlToBytes` now uses `(4 - (base64.length % 4)) % 4` with a `"==="` pad template (padding is always based on `length % 4`, not `% 3`). Regression coverage added in `tests/unit/helpers.test.ts`: round-trips every remainder class (0, 1, 2, 3) and a realistic 342-char Google RSA JWK `n` value (length % 4 == 2) which the old formula decoded to the wrong length.

src/core/crypto/secp256r1.ts
```diff
@@ -300,8 +300,8 @@ export class Secp256r1PrivateKey extends PrivateKey {
   sign(message: HexInput): Secp256r1Signature {
     const msgHex = Hex.fromHexInput(message);
     const sha3Message = sha3_256(msgHex.toUint8Array());
-    const signature = p256.sign(sha3Message, this.key.toUint8Array());
-    return new Secp256r1Signature(signature.toCompactRawBytes());
+    const signature = p256.sign(sha3Message, this.key.toUint8Array(), { prehash: false });
+    return new Secp256r1Signature(signature);
   }
```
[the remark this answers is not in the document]
gregnazario: Intentional as a consequence of the `@noble/curves` v2 migration, but worth calling out.

In noble v2 the ECDSA signing API no longer low-S normalizes by default. Nothing on-chain enforces low-S for secp256r1 signatures in Aptos (secp256r1 in Aptos is primarily consumed via AnyPublicKey/AnySignature + passkey/WebAuthn flows, which relay whatever signature the authenticator produced). Re-introducing low-S normalization at the SDK layer would cause SDK-produced signatures to stop byte-matching signatures produced by WebAuthn authenticators / the equivalent Rust SDK, which would break interop.

Happy to add an explicit `normalizeS: true` option on `sign()` for callers that want canonical form, but making it the default is not the right call here. Tracking as a follow-up — not blocking this PR.

src/core/crypto/secp256r1.ts
```diff
@@ -442,8 +442,8 @@ export class Secp256r1Signature extends Signature {
     if (signatureLength !== Secp256r1Signature.LENGTH) {
       throw new Error(`Signature length should be ${Secp256r1Signature.LENGTH}, received ${signatureLength}`);
     }
-    const signature = p256.Signature.fromCompact(hex.toUint8Array()).normalizeS().toCompactRawBytes();
-    this.data = Hex.fromHexInput(signature);
+    const signature = p256.Signature.fromBytes(hex.toUint8Array());
+    this.data = Hex.fromHexInput(signature.toBytes());
   }
```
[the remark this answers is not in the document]
gregnazario: Same rationale as the sibling comment on `sign()` (line 305): noble v2 dropped the implicit low-S normalization, and for secp256r1 we want the SDK to round-trip whatever signature bytes it was handed (especially for passkey/WebAuthn ingest). Normalizing at construction would mutate the caller's bytes and break byte-for-byte verification against signatures produced by external authenticators. We still validate the `r`/`s` field ranges via `p256.Signature.fromBytes`, so malformed signatures still fail fast. If we want canonical form, it belongs on an explicit `normalize()` method, not in the constructor.

examples/javascript/simple_transfer.js
```diff
@@ -1,14 +1,12 @@
...
-const dotenv = require("dotenv");
-dotenv.config();
-const { Account, Aptos, AptosConfig, NetworkToNetworkName, Network } = require("@aptos-labs/ts-sdk");
+import { Account, Aptos, AptosConfig, NetworkToNetworkName, Network } from "@aptos-labs/ts-sdk";
 
 const ALICE_INITIAL_BALANCE = 1_000_000_000;
 const BOB_INITIAL_BALANCE = 100;
 const TRANSFER_AMOUNT = 100;
-const APTOS_NETWORK = NetworkToNetworkName[process.env.APTOS_NETWORK] || Network.DEVNET;
+const APTOS_NETWORK = NetworkToNetworkName[process.env.APTOS_NETWORK ?? Network.DEVNET];
```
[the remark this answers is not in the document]
gregnazario: Done — applied the exact suggestion. `examples/javascript/simple_transfer.js` now reads `const APTOS_NETWORK = NetworkToNetworkName[process.env.APTOS_NETWORK ?? Network.DEVNET] ?? Network.DEVNET;` so misspelled/unknown `APTOS_NETWORK` values fall back to `Network.DEVNET` instead of `undefined`.

examples/javascript/simple_sponsored_transaction.js
```diff
@@ -2,16 +2,14 @@
...
-const { Account, Aptos, AptosConfig, NetworkToNetworkName, Network } = require("@aptos-labs/ts-sdk");
+import { Account, Aptos, AptosConfig, NetworkToNetworkName, Network } from "@aptos-labs/ts-sdk";
 
 const ALICE_INITIAL_BALANCE = 1_000_000_000;
 const SPONSOR_INITIAL_BALANCE = 1_000_000_000;
 const BOB_INITIAL_BALANCE = 0;
 const TRANSFER_AMOUNT = 10;
 // Default to devnet, but allow for overriding
-const APTOS_NETWORK = NetworkToNetworkName[process.env.APTOS_NETWORK] || Network.DEVNET;
+const APTOS_NETWORK = NetworkToNetworkName[process.env.APTOS_NETWORK ?? Network.DEVNET];
```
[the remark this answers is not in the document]
gregnazario: Done — same fix as in `simple_transfer.js`. `examples/javascript/simple_sponsored_transaction.js` now falls back to `Network.DEVNET` on both the env-var lookup miss and the `NetworkToNetworkName` lookup miss.

tests/e2e/client/aptosRequest.test.ts
```diff
@@ -55,7 +55,9 @@
             config,
             AptosApiType.FULLNODE,
           );
-          expect(response.config.headers).toHaveProperty("x-aptos-client", `aptos-typescript-sdk/${VERSION}`);
+          expect(response.config.headers["x-aptos-client"]).toMatch(
+            new RegExp(`^aptos-typescript-sdk/${VERSION.replace(/\./g, "\\.")}; platform=\\w+(-\\w+)?$`),
```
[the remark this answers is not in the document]
gregnazario: Fixed in 31f315ded. Replaced the dynamic `new RegExp(\"...${VERSION}...\")` with a `startsWith` prefix check for the version-bearing part and a static regex literal for the platform-tag tail. Same assertion semantics, no dynamic RegExp, no unescaped-backslash risk. `VERSION` is the SDK's own `src/version.ts` constant (trusted), but this also makes CodeQL happy.

src/internal/transaction.ts
```diff
@@ -459,7 +459,7 @@ async function fillBlockTransactions(args: {
     const lastVersion = BigInt(block.last_version);
 
     // Convert the transaction to the type
-    const curVersion: string | undefined = (lastTxn as any)?.version;
+    const curVersion: string | undefined = "version" in lastTxn ? lastTxn.version : undefined;
     let latestVersion;
```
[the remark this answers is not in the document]
gregnazario: Fixed in 31f315ded. `fillBlockTransactions` now guards with `lastTxn && "version" in lastTxn`, so an empty-block response returns a safe `curVersion = undefined` path (which was always the intended behavior — see the `"we don't have any transactions"` branch just below) instead of throwing `TypeError: Cannot use 'in' operator to search for 'version' in undefined`.

examples/deno-test/simple_transfer.ts
```diff
@@ -21,7 +21,7 @@ const BOB_INITIAL_BALANCE = 100;
 const TRANSFER_AMOUNT = 100;
 
 // Default to devnet, but allow for overriding
-const APTOS_NETWORK: Network = NetworkToNetworkName[Deno.env.get("APTOS_NETWORK") ?? ""] ?? Network.DEVNET;
+const APTOS_NETWORK: Network = NetworkToNetworkName[Deno.env.get("APTOS_NETWORK") ?? "devnet"];
```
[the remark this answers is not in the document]
gregnazario: Fixed in 31f315ded. Added the outer `?? Network.DEVNET` fallback so an unknown `APTOS_NETWORK` value falls back to devnet instead of passing `undefined` into `AptosConfig`. Applied to both `simple_transfer.ts` and `simple_sponsored_transaction.ts` in the Deno example.

examples/bun-test/simple_transfer.ts
```diff
@@ -23,7 +23,7 @@ const BOB_INITIAL_BALANCE = 100;
 const TRANSFER_AMOUNT = 100;
 
 // Default to devnet, but allow for overriding
-const APTOS_NETWORK: Network = NetworkToNetworkName[Bun.env.APTOS_NETWORK ?? ""] ?? Network.DEVNET;
+const APTOS_NETWORK: Network = NetworkToNetworkName[Bun.env.APTOS_NETWORK ?? Network.DEVNET];
```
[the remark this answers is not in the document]
gregnazario: Fixed in 31f315ded. Same treatment as the Deno example — outer `?? Network.DEVNET` on both `simple_transfer.ts` and `simple_sponsored_transaction.ts` so misspelled `APTOS_NETWORK` values fall back to devnet instead of crashing `AptosConfig`.

tests/e2e/client/aptosRequest.test.ts
```diff
@@ -55,7 +55,9 @@
             config,
             AptosApiType.FULLNODE,
           );
-          expect(response.config.headers).toHaveProperty("x-aptos-client", `aptos-typescript-sdk/${VERSION}`);
+          expect(response.config.headers["x-aptos-client"]).toMatch(
+            new RegExp(`^aptos-typescript-sdk/${VERSION.replace(/\./g, "\\.")}; platform=\\w+(-\\w+)?(/[\\d.]+)?$`),
```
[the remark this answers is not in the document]
gregnazario: Same fix as #46 above — addressed in 31f315ded by removing the dynamic `new RegExp(...)` altogether in favor of `startsWith` + a static regex literal for the platform-tag tail.

src/client/core.ts
```diff
@@ -96,16 +105,18 @@ export async function aptosRequest<Req extends {}, Res extends {}>(
...
       throw new AptosApiError({
         apiType,
         aptosRequest: aptosRequestOpts,
         aptosResponse,
       });
     }
-    aptosResponse.data = indexerResponse.data as Res;
+    if ("data" in indexerResponse) {
+      aptosResponse.data = indexerResponse.data as Res;
+    }
```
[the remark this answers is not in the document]
gregnazario: Fixed in 65e7a0c1c. Added `typeof indexerResponse === "object" && indexerResponse !== null` around the `errors`/`data` key probes so a primitive/null payload no longer throws a `TypeError` and masks the real API error.

src/internal/transactionSubmission.ts
```diff
@@ -361,17 +367,19 @@ export async function submitTransaction(
...
+      if (variant === AnyPublicKeyVariant.Keyless || variant === AnyPublicKeyVariant.FederatedKeyless) {
+        // Dynamic import to avoid pulling poseidon-lite into the main bundle
+        const { AbstractKeylessAccount } = await import("../account/AbstractKeylessAccount.js");
+        type KP = import("../core/crypto/keyless.js").KeylessPublicKey;
+        type KS = import("../core/crypto/keyless.js").KeylessSignature;
+        await AbstractKeylessAccount.fetchJWK({
+          aptosConfig,
+          publicKey: signedTxn.authenticator.sender.public_key.publicKey as KP,
+          kid: (signedTxn.authenticator.sender.signature.signature as KS).getJwkKid(),
+        });
```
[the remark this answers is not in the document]
gregnazario: Fixed in 65e7a0c1c. Widened the local `KP` type alias to `KeylessPublicKey | FederatedKeylessPublicKey` to match `AbstractKeylessAccount.fetchJWK`'s signature, so the `AnyPublicKeyVariant.FederatedKeyless` branch is no longer miscast.

src/account/keylessSigner.ts
```diff
@@ -0,0 +1,29 @@
...
+   * `AbstractKeylessAccount` subclasses; exposed here so modules that must
+   * avoid a compile-time dependency on the keyless account classes (to keep
+   * `poseidon-lite` out of their bundle) can still interact structurally.
+   */
+  waitForProofFetch(): Promise<void>;
+}
+
+export function isKeylessSigner(obj: any): obj is KeylessSigner {
+  return obj !== null && obj !== undefined && typeof obj.checkKeylessAccountValidity === "function";
+}
```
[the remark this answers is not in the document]
gregnazario: Fixed in 65e7a0c1c. `isKeylessSigner` now also verifies `typeof obj.waitForProofFetch === "function"`, keeping the runtime check aligned with the narrowed `KeylessSigner` interface. All existing call sites pass `AbstractKeylessAccount` instances, which implement both methods, so this is strictly tightening.

src/internal/ans.ts
```diff
@@ -1006,17 +1008,17 @@ function sanitizeANSName({
...
     expiration_timestamp,
     expiration_status,
-    domain_expiration_timestamp: domain_expiration_timestamp!,
+    domain_expiration_timestamp: domain_expiration_timestamp,
     expiration: new Date(expiration),
-    token_standard: name.token_standard! as AnsTokenStandard,
-    is_primary: name.is_primary!,
+    token_standard: name.token_standard ?? "v2",
+    is_primary: name.is_primary ?? false,
     subdomain_expiration_policy: name.subdomain_expiration_policy ?? SubdomainExpirationPolicy.FollowsDomain,
```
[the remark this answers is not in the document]
gregnazario: Good point — fixed in 8c729f0b3. Reverted the fabrication: `AnsName.token_standard` and `is_primary` are now optional (`AnsTokenStandard | undefined`, `boolean | undefined`), and `sanitizeANSName` passes them through as `undefined` instead of defaulting to `"v2"`/`false`. `subdomain_expiration_policy` keeps its `FollowsDomain` fallback because that's a documented semantic default for subdomains, not a fabricated value. Documented under Breaking in CHANGELOG.

src/utils/helpers.ts
```diff
@@ -106,23 +117,41 @@ export function floorToWholeHour(timestampInSeconds: number): number {
...
+ * @group Implementation
+ * @category Utils
+ */
+export function base64UrlEncode(input: string | Uint8Array): string {
+  const bytes = typeof input === "string" ? TEXT_ENCODER.encode(input) : input;
+  // btoa requires a Latin-1 string — map each byte to a char code.
+  let binary = "";
+  for (let i = 0; i < bytes.length; i++) {
+    binary += String.fromCharCode(bytes[i]);
+  }
```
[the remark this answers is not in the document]
gregnazario: Fixed in 8c729f0b3. `base64UrlEncode` now builds the Latin-1 string in 8 KiB chunks via `String.fromCharCode.apply` + `Array.join` instead of per-byte `+=` concatenation. That keeps it linear-time and also avoids the call-stack overflow you'd hit from `String.fromCharCode(...bytes)` on large inputs.

src/internal/transactionSubmission.ts
```diff
@@ -361,17 +367,23 @@ export async function submitTransaction(
...
-        signedTxn.authenticator.sender.public_key.publicKey instanceof FederatedKeylessPublicKey)
-    ) {
-      await AbstractKeylessAccount.fetchJWK({
-        aptosConfig,
-        publicKey: signedTxn.authenticator.sender.public_key.publicKey,
-        kid: (signedTxn.authenticator.sender.signature.signature as KeylessSignature).getJwkKid(),
-      });
+    if (signedTxn.authenticator.isSingleSender() && signedTxn.authenticator.sender.isSingleKey()) {
+      const { variant } = signedTxn.authenticator.sender.public_key;
+      if (variant === AnyPublicKeyVariant.Keyless || variant === AnyPublicKeyVariant.FederatedKeyless) {
```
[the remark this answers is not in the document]
gregnazario: Fixed in 8c729f0b3. Wrapped the entire best-effort keyless diagnostic (deserialize → variant check → dynamic import → `fetchJWK`) in its own `try { … } catch { /* swallow */ }` so any failure there cannot mask the original submission error `e`. `e` is always the thing that gets rethrown.

src/internal/ans.ts
```diff
@@ -1006,17 +1008,17 @@ function sanitizeANSName({
   }
 
   return {
-    domain: name.domain!,
+    domain: name.domain ?? "N/A",
     subdomain: name.subdomain || undefined,
     expiration_timestamp,
```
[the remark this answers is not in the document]
gregnazario: Fixed in 8c729f0b3. Agreed — fabricating `"N/A"` for a nullable field is worse than being honest. `AnsName.domain` is now `string | undefined`, and `sanitizeANSName` passes `name.domain` through untouched (no placeholder). Same treatment for `owner_address` and `registered_address`. Documented under Breaking in CHANGELOG; callers that previously relied on a non-null `domain` should now handle `undefined` (or filter the row as bad indexer data).

kaw2k (approved):

PR #926, opened 2026-07-06
feat(keyless): add aptos.keyless.getPepperBase

maykonmichel:
### Description

Adds `aptos.keyless.getPepperBase(...)` — fetches a keyless account's 48-byte `pepper_base` (the VUF signature / compressed BLS12-381 G1 point from which the final pepper is derived) from the pepper service's existing `signature` endpoint. It complements `getPepper`, which returns the 31-byte derived pepper.

Unlike the final pepper, `pepper_base` is independent of the ephemeral key and the derivation path, so it's a stable per-identity seed. It is the seed for the confidential-asset keyless decryption-key (DK) derivation (`@aptos-labs/confidential-asset`'s `TwistedEd25519PrivateKey.fromPepperBase`). Deriving the DK from `pepper_base` rather than the final pepper (a one-way hash of it) ensures a leaked pepper cannot recover the DK.

No server/aptos-core change is required — the `/v0/signature` endpoint already exists; this only adds the SDK client for it:
- `SignatureFetchResponse` type (`src/types/keyless.ts`)
- internal `getPepperBase` posting to `path: "signature"` (`src/internal/keyless.ts`)
- public `Keyless.getPepperBase` (`src/api/keyless.ts`)

### Test Plan

- Unit tests in `tests/unit/internal/keyless-pepper.test.ts` (mirroring `getPepper`): assert `getPepperBase` POSTs to the `signature` endpoint, returns a 48-byte value, and forwards `uidKey`/`derivationPath`.
- Manually validated against the live **devnet and mainnet** pepper services with the `test.oidc.provider` fixture: `getPepperBase` returns a real 48-byte `pepper_base`, and the same request's `getPepper` returns the known test-issuer pepper (`0x7727…56c`).
- `pnpm build` and `pnpm check` pass.

### Related Links

- Consumer: `@aptos-labs/confidential-asset` `TwistedEd25519PrivateKey.fromPepperBase` (keyless DK derivation).
- aptos-core confidential-assets keyless integration (PR #19458).

### Checklist

### DIFF
--- a/src/internal/keyless.ts
+++ b/src/internal/keyless.ts
@@ -27,7 +27,13 @@ import { Account } from "../account/index.js";
 import { EphemeralKeyPair } from "../account/EphemeralKeyPair.js";
 import { KeylessAccount } from "../account/KeylessAccount.js";
 import { ProofFetchCallback } from "../account/AbstractKeylessAccount.js";
-import { PepperFetchRequest, PepperFetchResponse, ProverRequest, ProverResponse } from "../types/keyless.js";
+import {
+  PepperFetchRequest,
+  PepperFetchResponse,
+  ProverRequest,
+  ProverResponse,
+  SignatureFetchResponse,
+} from "../types/keyless.js";
 import { lookupOriginalAccountAddress } from "./account.js";
 import { FederatedKeylessPublicKey } from "../core/crypto/federatedKeyless.js";
 import { FederatedKeylessAccount } from "../account/FederatedKeylessAccount.js";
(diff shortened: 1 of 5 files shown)


src/internal/keyless.ts
```diff
@@ -76,6 +82,52 @@ export async function getPepper(args: {
...
+ * @group Implementation
+ */
+export async function getPepperBase(args: {
+  aptosConfig: AptosConfig;
+  jwt: string;
+  ephemeralKeyPair: EphemeralKeyPair;
+  uidKey?: string;
+  derivationPath?: string;
+}): Promise<Uint8Array> {
+  const { aptosConfig, jwt, ephemeralKeyPair, uidKey = "sub", derivationPath } = args;
```
gregnazario, line 111: interesting I didn't know you could inline default like this.  Probably better to just put it in the input arg though (oh, but then you need a giant blob, nevermind)

gregnazario (approved):

